PR #32, opened 2019-05-16
to cleanup all created resources, not only kube-ovn namespace.

fanfanbj:
There are some resources that are not being deleted, such as role, sa, rolebinding and so on.
So delete kube-ovn namespace not enough.


dist/images/cleanup.sh
```diff
@@ -6,7 +6,8 @@ kubectl patch svc -n kube-ovn ovn-nb --type='json' -p '[{"op": "replace", "path"
 kubectl patch svc -n kube-ovn ovn-sb --type='json' -p '[{"op": "replace", "path": "/metadata/finalizers", "value": []}]' || true
 
 # Delete Kube-OVN components
-kubectl delete ns kube-ovn
+kubectl delete -f ovn.yaml
```
oilbeater: The yaml file may not exist in the current folder. Use url points to the yaml might better

fanfanbj:
Done. I add url in yaml files.

oilbeater:
Thanks! @fanfanbj

PR #131, opened 2019-11-07
fix: add keepalive to ovn-controller

oilbeater:


halfcrazy:
from https://www.slideshare.net/hanzhou1978/large-scale-overlay-networks-with-ovn-problems-and-solutions page 16, he recommends an extra configuration for server-side.

oilbeater:
updated

PR #140, opened 2019-11-13
fix: network unreachable when add egress qos for pod

qsyqian:
when add egress qos for pod, the pod network is unreachable.
through research, the container's virtual nic's qlen is 0 is
the root reason. so set the nic's qlen to 1000(default for linux)
when create pod.


oilbeater:
Thanks! @qsyqian

PR #149, opened 2019-11-19
use short kubernetes.default service

sbskas:
Use the short name kubernetes.default which is always resolvable instead of the fqdn which may be different from cluster to cluster.


oilbeater:
Thanks! @sbskas

PR #155, opened 2019-11-25
fix: index out of range err when create lsp

qsyqian:
when subnet address space has been exhausted with one ls,
the output when get dynamic address for one lsp will only
return mac for lsp, so cann't assume the output has both
ip and mac.


oilbeater:
Thanks! @qsyqian

PR #187, opened 2019-12-04
fix: ovn-cluster mode

oilbeater:
After update ovn to 2.12, the previous ovn cluster can not be established. The reasons are nbctl/sbctl can not write data to local database and return error if it's not the leader and when restart it should point to the new leader to rejoin the cluster.


oilbeater:
nb and sb leader may in different pods, we should handle them separately.

PR #226, opened 2020-01-14
fix: add kube-ovn types to default scheme

mars1024:
If we want to add kube-ovn types to default scheme for eventing, we should use `AddToScheme` function from kube-ovn but not a function from kubernetes.


oilbeater:
Good catch, thanks! @mars1024

PR #231, opened 2020-02-10
fix: wait for container network ready before cni return

oilbeater:
When cluster with thousands pods, it take time for ovs network take effect. To avoid pods run before network ready, wait network ready in cniserver

### DIFF
--- a/pkg/daemon/gateway.go
+++ b/pkg/daemon/gateway.go
@@ -52,7 +52,6 @@ var (
 )
 
 func (c *Controller) runGateway() {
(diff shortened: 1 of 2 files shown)


halfcrazy:
A little suggestion, if the node network not ready, we can set as NetworkUnavailable.

oilbeater:
> A little suggestion, if the node network not ready, we can set as NetworkUnavailable.

This patch add waiting in cni/add operation to wait the pod related flows to be installed on the node. We need other method to check if the node network is ready

PR #244, opened 2020-02-20
webhook: remove leader-election configs

mars1024:
Webhooks are several web servers which is serving independently, leader election is not needed in this scenario.

### DIFF
--- a/cmd/webhook/server.go
+++ b/cmd/webhook/server.go
@@ -57,9 +57,6 @@ func main() {
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme:                  scheme,
-		LeaderElection:          true,
-		LeaderElectionNamespace: os.Getenv("KUBE_NAMESPACE"),
-		LeaderElectionID:        os.Getenv("POD_NAME"),
 		// disable metrics to avoid port conflict
 		MetricsBindAddress: "0",
 	})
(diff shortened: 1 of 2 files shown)


mars1024:
friendly ping @oilbeater

oilbeater:
@mars1024 the reason for leader election is that we should check the ip conflict one by one to avoid potential conflict caused by concurrent requests.

We know that there are some flaws in current webhook implementation and the conflict check part will be move to the new IPAM later.

mars1024:
I'm sorry but the leader election won't work with `webhook.Server` because it is an implementation of  `LeaderElectionRunnable` interface and the result of method `NeedLeaderElection` is false, so it will be ignored in [Add](https://github.com/alauda/kube-ovn/blob/232251bbb1f09229e8ea01ba9a554e8a3f4e6809/vendor/sigs.k8s.io/controller-runtime/pkg/manager/internal.go#L141).

I think if you want to avoid ip check conflict, we can fix the replica of webhook to 1, what do you think?

oilbeater:
@mars1024 I think you are right, I misunderstood this part of code. 
@halfcrazy What do you think about the change

mars1024:
> @mars1024 I think you are right, I misunderstood this part of code.
> @halfcrazy What do you think about the change

BTW, should we fix the replica of webhook to 1?

oilbeater:
@mars1024 I remember the initial thought about the webhook. At that time, we want to prevent users set duplicated static addresses. However it's hard to handle all the edge cases, like concurrent requests, dynamic allocated address might steal the static allocated addresses etc. So the webhook just do its best to handle the common cases, there are still chances for conflicts happen.

To totally avoid conflicts needs further consideration. By now I think it's ok to merge this PR

PR #246, opened 2020-02-21
controller: squash redundant list actions in loop

mars1024:
As we know, list action (especially for pods) is expensive, so we can avoid listing again and again in a single loop.


oilbeater:
@mars1024 you remind me that I had a pod controller refactor that forget to commit.
In the refactor here:
https://github.com/alauda/kube-ovn/pull/248/files#diff-8d4650ed46e8156bd1ea16a8212e7f76R531
we use the cached pods to avoid list to apiserver

mars1024:
Of course! Obviously infomer cache is a better way :)

mars1024:
closing this

PR #247, opened 2020-02-21
webhook: use global variables to avoid repeated map constructing

mars1024:
use global variables to avoid repeated map constructing


oilbeater:
Thanks! @mars1024

PR #249, opened 2020-02-22
fix:enable bash alias option in Dockerfile CMD scripts

ohmyadd:
set `shopt -s expand_aliases` in bash scripts to make alias command available when using `./ovn-healthcheck.sh`


oilbeater:
Thanks! @ohmyadd

PR #255, opened 2020-03-04
refactor ipam

oilbeater:
Use kube-ovn ipam to replace ovn dynamic address allocation. We will design more flexible ip allocation strategy and provide ipam capacity to other cni plugin later.


halfcrazy:
Looks like all IPAM data stores in memory, what if controller pod restart?

oilbeater:
> Looks like all IPAM data stores in memory, what if controller pod restart?

The initIPAM will fetch the pod annotations and init ipam data

PR #295, opened 2020-04-03
init logger for controller-runtime

mars1024:
controller-runtime is throwing error with its own log tool, just like [GetConfigOrDie](https://github.com/kubernetes-sigs/controller-runtime/blob/df180accb901c816ced95a7b1a6eaf2c801f3560/pkg/client/config/config.go#L162), but this log tool should be initialized explicitly just as the [annotations](https://github.com/kubernetes-sigs/controller-runtime/blob/df180accb901c816ced95a7b1a6eaf2c801f3560/pkg/log/log.go#L45), otherwise the error message will be swallow and never show up through [a null logger](https://github.com/kubernetes-sigs/controller-runtime/blob/df180accb901c816ced95a7b1a6eaf2c801f3560/pkg/log/null.go#L28).


mars1024:
friendly ping @oilbeater

oilbeater:
Thanks! @mars1024

PR #301, opened 2020-04-08
OVS local interface table mac_in_use col is in lower case, but pod

halfcrazy:
annotation store mac in Upper case.
image


oilbeater:
mac 上这个替换好像不能用

oilbeater:
install.sh 那个脚本对应位置也得改

halfcrazy:
update

PR #316, opened 2020-04-10
chore: update kind node to 1.18 and ginkgo

oilbeater:


halfcrazy:
We've switched to  gomodule, so no need to bundle vendor anymore.

PR #350, opened 2020-05-18
Remotes/origin/release/1.1

qianxiaocao:
feat: 1. use annotation.ovn.kubernetes.io/north_gateway=[ip] set a ip as pod gw
          2. use annotation.ovn.kubernetes.io/port_security set a pod port no security can work as a gw


oilbeater:
reimplement in #421

PR #352, opened 2020-05-19
Add OVS-DPDK support, for issue 104

garyloug:
This commit adds OVS-DPDK support to Kube-OVN. User instructions are
included in a new file docs/dpdk.md

A new Dockerfile has been added to include OVS-DPDK along with OVN.
Where DPDK is required, this image is used for the ovs-ovn pod, in
place of the existing kernel-OVS “kube-ovn” image. This Dockerfile is
currently based on Fedora 32 for reasons noted as comments within the
file. It should later be possible to change this to CentOS when full
DPDK 19 support is available.

I recommend the above Dockerfile is built and tagged as
kube-ovn-dpdk:<version>, where the version corresponds to the DPDK
version used within the image (in this case 19:11) rather than the
Kube-OVN version. I recommend this as DPDK applications have a strong
dependency on DPDK version. If we force an end user to always use the
latest version, then we will likely break their DPDK app. I propose
over time we provide images for multiple DPDK versions and let the user
pick to suit their needs. I don’t see these images or Dockerfiles
requiring maintenance or support. They should be completely independent
of Kube-OVN versions and releases.

The install.sh script has been modified. It now takes a flag
--with-dpdk=<version> so the user can indicate they want to install
OVS-DPDK based on which version of DPDK. Version of DPDK required will
determine version of OVS and this will be already built into the Docker
image provided. The Kube-OVN version installed is still set at the top
of the script as the VERSION variable. This should still be the case
going forward, Kube-OVN and DPDK versions should operate independently
of each other. However, it’s something to watch. If future versions of
Kube-OVN have a strong dependency on newer versions of OVS, then the
older version of OVS used for DPDK may become an issue. We may have to
update the install script so a user wanting an older version of DPDK
has no choice but to use an older version of Kube-OVN that’s known to
be compatible. I don’t foresee this being an issue, but one to watch as
I said.

New startup and healthcheck scripts added for OVS-DPDK.

### DIFF
--- a/dist/images/Dockerfile.dpdk1911
+++ b/dist/images/Dockerfile.dpdk1911
@@ -18,22 +18,17 @@ ENV DPDK_DIR=/usr/src/dpdk-stable-${DPDK_VERSION}
 ENV PATH=${PATH}:/usr/share/openvswitch/scripts
 ENV PATH=${PATH}:/usr/share/ovn/scripts/
 
-RUN dnf install -y wget xz git make numactl-devel diffutils dnf-plugins-core dpdk-devel libcap-ng-devel libpcap-devel nc iptables ipset hostname && dnf clean all
-
+RUN dnf install -y make numactl-devel diffutils dnf-plugins-core dpdk-devel libcap-ng-devel libpcap-devel nc iptables ipset hostname \
 # NOTE: Fedora 32 comes with gcc version 10. This caused LD linker issues during DPDK build.
 # Install gcc version 9 and its dependencies.
-RUN mkdir /rpms && cd /rpms
-RUN wget https://kojipkgs.fedoraproject.org//packages/gcc/9.2.1/1.fc32.3/x86_64/cpp-9.2.1-1.fc32.3.x86_64.rpm \
-         https://kojipkgs.fedoraproject.org//packages/gcc/9.2.1/1.fc32.3/x86_64/libgomp-9.2.1-1.fc32.3.x86_64.rpm \
-         https://kojipkgs.fedoraproject.org//packages/gcc/9.2.1/1.fc32.3/x86_64/gcc-9.2.1-1.fc32.3.x86_64.rpm
-RUN dnf install -y cpp-9.2.1-1.fc32.3.x86_64.rpm libgomp-9.2.1-1.fc32.3.x86_64.rpm gcc-9.2.1-1.fc32.3.x86_64.rpm
-RUN rm -f cpp-9.2.1-1.fc32.3.x86_64.rpm libgomp-9.2.1-1.fc32.3.x86_64.rpm gcc-9.2.1-1.fc32.3.x86_64.rpm
+  https://kojipkgs.fedoraproject.org//packages/gcc/9.2.1/1.fc32.3/x86_64/cpp-9.2.1-1.fc32.3.x86_64.rpm \
+  https://kojipkgs.fedoraproject.org//packages/gcc/9.2.1/1.fc32.3/x86_64/libgomp-9.2.1-1.fc32.3.x86_64.rpm \
+  https://kojipkgs.fedoraproject.org//packages/gcc/9.2.1/1.fc32.3/x86_64/gcc-9.2.1-1.fc32.3.x86_64.rpm && \
+  dnf clean all
 
 # Install DPDK
 RUN cd /usr/src/ && \
-  wget http://fast.dpdk.org/rel/dpdk-${DPDK_VERSION}.tar.xz && \
-  tar xf dpdk-${DPDK_VERSION}.tar.xz && \
-  rm -f dpdk-${DPDK_VERSION}.tar.xz && \
+  curl http://fast.dpdk.org/rel/dpdk-${DPDK_VERSION}.tar.gz | tar xz && \
   cd ${DPDK_DIR} && \
   sed -i s/CONFIG_RTE_EAL_IGB_UIO=y/CONFIG_RTE_EAL_IGB_UIO=n/ config/common_linux && \
   sed -i s/CONFIG_RTE_LIBRTE_KNI=y/CONFIG_RTE_LIBRTE_KNI=n/ config/common_linux && \
@@ -42,11 +37,9 @@ RUN cd /usr/src/ && \
 
 # Install OVS-DPDK
 # NOTE: rpm-build is installed here, as installing it earlier conflicts with the workaround to roll back the gcc version.
-RUN dnf install -y rpm-build
-RUN cd /usr/src/ && \
-  wget https://www.openvswitch.org/releases/openvswitch-${OVS_VERSION}.tar.gz && \
-  tar xf openvswitch-${OVS_VERSION}.tar.gz && \
-  rm -f openvswitch-${OVS_VERSION}.tar.gz && \
+RUN dnf install -y rpm-build && \
+  cd /usr/src/ && \
+  curl https://www.openvswitch.org/releases/openvswitch-${OVS_VERSION}.tar.gz | tar xz && \
   cd ${OVS_DIR} && \
   sed -e 's/@VERSION@/0.0.1/' rhel/openvswitch-fedora.spec.in > /tmp/ovs.spec && \
   dnf builddep -y /tmp/ovs.spec && \
@@ -56,17 +49,20 @@ RUN cd /usr/src/ && \
 
 # Install OVN
 RUN cd /usr/src/ && \
-  git clone -b branch-${OVN_VERSION} --depth=1 https://github.com/ovn-org/ovn.git && \
+  cd /usr/src/ && \
+  curl -L https://github.com/ovn-org/ovn/tarball/branch-${OVN_VERSION} > ovn.tar.gz && \
+  mkdir ovn && tar -xf ovn.tar.gz -C ovn --strip-components 1 && \
+  rm -f ovn.tar.gz && \
   cd ovn && \
   ./boot.sh && \
   ./configure --with-ovs-source=${OVS_DIR} && \
   make rpm-fedora
 
-RUN cp ${OVS_DIR}/rpm/rpmbuild/RPMS/${RPM_ARCH}/* ${OVN_DIR}/rpm/rpmbuild/RPMS/${RPM_ARCH}/* /rpms
-
 RUN mkdir -p /var/run/openvswitch && \
-    mkdir -p /var/run/ovn
-
-RUN rpm -ivh --nodeps /rpms/*.rpm
+  mkdir -p /var/run/ovn && \
+  mkdir -p /rpms && \
+  cp ${OVS_DIR}/rpm/rpmbuild/RPMS/${RPM_ARCH}/* ${OVN_DIR}/rpm/rpmbuild/RPMS/${RPM_ARCH}/* /rpms && \
+  rpm -ivh --nodeps /rpms/*.rpm && \
+  rm -rf ${OVN_DIR} ${OVS_DIR} ${DPDK_DIR}
 
 COPY *.sh /kube-ovn/
--- a/dist/images/Dockerfile.dpdk1911
+++ b/dist/images/Dockerfile.dpdk1911
@@ -1,0 +1,72 @@
+# NOTE: At time of writing, CentOS 8 does not have dpdk-devel v19.11 in the Yum repo.
+# This package is required to build the OVS-DPDK RPMs and is available in Fedora 32.
+FROM fedora:32
+
+ENV RPM_ARCH=x86_64
+ENV ARCH=amd64
+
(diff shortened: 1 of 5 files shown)


garyloug:
I think this PR is ready for review. I'll leave it WIP/draft while we do some final testing on our end.

oilbeater:
@garyloug I didn't find how to manage the address/route of the dpdk interface. Can the dpdk interface be used by application as a normal interface? Or it can only be used by the dpdk application as a raw device?

dist/images/Dockerfile.dpdk1911
```diff
@@ -0,0 +1,72 @@
...
+  cd ovn && \
+  ./boot.sh && \
+  ./configure --with-ovs-source=${OVS_DIR} && \
+  make rpm-fedora
+
+RUN cp ${OVS_DIR}/rpm/rpmbuild/RPMS/${RPM_ARCH}/* ${OVN_DIR}/rpm/rpmbuild/RPMS/${RPM_ARCH}/* /rpms
+
+RUN mkdir -p /var/run/openvswitch && \
+    mkdir -p /var/run/ovn
+
```
oilbeater, line 69: I find the final image is 3.12GB can we use multi-stage build that mv all the rpm packages in a new Fedora image to install and reduce the image size?

docs/dpdk.md
```diff
@@ -0,0 +1,194 @@
...
+This document describes how to run Kube-OVN with OVS-DPDK.
+
+
+## Prerequest
+- Kubernetes >= 1.11
+- Docker >= 1.12.6
+- OS: CentOS 7.5/7.6/7.7, Ubuntu 16.04/18.04
+
+
+## To Install
```
oilbeater, line 12: we should add a section about how to setup dpdk on the host

dist/images/Dockerfile.dpdk1911
```diff
@@ -0,0 +1,72 @@
...
+RUN dnf install -y wget xz git make numactl-devel diffutils dnf-plugins-core dpdk-devel libcap-ng-devel libpcap-devel nc iptables ipset hostname && dnf clean all
+
+# NOTE: Fedora 32 comes with gcc version 10. This caused LD linker issues during DPDK build.
+# Install gcc version 9 and its dependencies.
+RUN mkdir /rpms && cd /rpms
+RUN wget https://kojipkgs.fedoraproject.org//packages/gcc/9.2.1/1.fc32.3/x86_64/cpp-9.2.1-1.fc32.3.x86_64.rpm \
+         https://kojipkgs.fedoraproject.org//packages/gcc/9.2.1/1.fc32.3/x86_64/libgomp-9.2.1-1.fc32.3.x86_64.rpm \
+         https://kojipkgs.fedoraproject.org//packages/gcc/9.2.1/1.fc32.3/x86_64/gcc-9.2.1-1.fc32.3.x86_64.rpm
+RUN dnf install -y cpp-9.2.1-1.fc32.3.x86_64.rpm libgomp-9.2.1-1.fc32.3.x86_64.rpm gcc-9.2.1-1.fc32.3.x86_64.rpm
+RUN rm -f cpp-9.2.1-1.fc32.3.x86_64.rpm libgomp-9.2.1-1.fc32.3.x86_64.rpm gcc-9.2.1-1.fc32.3.x86_64.rpm
```
oilbeater, line 30: The wget/install/rm steps should in one RUN step to reduce the image size

dist/images/Dockerfile.dpdk1911
garyloug: I’ve tried this. To install OVS-DPDK, you must first install DPDK in the environment, then build and install OVS against this DPDK install. Unfortunately, the resulting RPMs don’t seem to be portable, even if another environment already has same DPDK install.
I guess this is the reason you must build and install OVS-DPDK locally instead of it being available as RPM/YUM/DNF.

docs/dpdk.md
garyloug: OVS must be on the same environment as DPDK. If I move OVS-DPDK to the host then OVN also needs to move to the host because OVN RPMs depend on OVS RPMs. We agreed this does not fit well with Kube-OVN and so I focused my efforts on containerizing OVS-DPDK. The Docker image now contains DPDK, OVS and OVN. Nothing needs to be preinstalled on the host.

dist/images/Dockerfile.dpdk1911
garyloug: Cool. I'll try this and I'll also see if I can do anything else to reduce image size.

garyloug:
> @garyloug I didn't find how to manage the address/route of the dpdk interface. Can the dpdk interface be used by application as a normal interface? Or it can only be used by the dpdk application as a raw device?

It’s not a normal interface. OVS-DPDK gives us vhost-user socket files. This is the reason we cannot move them to the container network namespace, we must mount it as a regular file (functionality provided by Userspace-CNI). It’s also the reason that we cannot use OVS-DPDK as the default network. K8s can’t reach the pod like this for healthcheck. Therefore, we depend on Multus to attach multiple networks. The kernel-OVS interface is still provided by Kube-OVN as normal, this is still the default network. In addition, Multus allows us to also provide the pod with OVS-DPDK networking. This is the same instance of OVS, running within the container, but the DPDK ports are on a different DPDK-enabled bridge.

docs/dpdk.md
oilbeater: Do we need setup hugepage, cpu-mask and other system config on host?

oilbeater:
@garyloug I find some examples of Userspace-CNI, they are all about how dpdk applications use the DPDK ports. I wonder if a normal application, like a nginx webserver can take use of the dpdk port as a normal network interface.

docs/dpdk.md
```diff
@@ -0,0 +1,194 @@
...
+          "bridgeName": "br0"
+        }
+      },
+      "container": {
+        "engine": "ovs-dpdk",
+        "iftype": "vhostuser",
+        "netType": "interface",
+        "vhost": {
+          "mode": "client"
+        }
```
oilbeater, line 92: Do we need any ipam here?

docs/dpdk.md
garyloug: Good catch on the hugepages, these need to be setup on the host. I’ll document how to do this.

CPU mask can be used but not necessary. I purposely didn’t mention it here because it gets quite complex. If using CPU mask, you should also use something like [CMK](https://github.com/intel/CPU-Manager-for-Kubernetes). Configuration here is complex as I said, and it’s very difficult to write instructions that are generic for every user. I can mention that ovs-vsctl has cpu-mask option and I can link to the CMK project. But I think instructions are out of the scope of this document.

garyloug:
> @garyloug I find some examples of Userspace-CNI, they are all about how dpdk applications use the DPDK ports. I wonder if a normal application, like a nginx webserver can take use of the dpdk port as a normal network interface.

To use DPDK networking you must have a purpose built DPDK application. Regular apps like Nginx will not work. There are some sample apps like Testpmd and L2fwd. I do plan to include some instructions for these, but since using DPDK 19.11 (required for OVS 2.13) these apps seem to have some slightly different flags and args. I just need to figure this out.

docs/dpdk.md
garyloug: No, DPDK is primarily L2.

oilbeater:
@garyloug Thanks for the explanation. From my understand, DPDK applications can run in Pod with this pr. I think after some document fix mentioned above, this pr can be merged. Thanks again for the great work.

I still have another question about OVS-DPDK. I read some examples about how to set ovs-dpdk with qemu, inside the vm boot by qemu, the network interface is a normal interface from application point of view. Is there any driver for container to do stuff like that? I think it's more user friendly for application that they don't need any dpdk specific knowledge and can take advantage of the high performance network.

garyloug:
@oilbeater Cool, I’m currently working on image size reduction. I’ve saved about 300Mb so far by using preinstalled curl instead of wget and git. I’ve avoided the need for xz install by downloading tar.gz instead of tar.xz files. Combining a few RUNs to reduce layers, etc. I’m looking at reducing DPDK install now, I’m sure there are things installed that are not needed by OVS, but this might take some time as it’s not something I’m familiar with. I’ll also uninstall all packages needed just for the builds, like Make and other development tools.

On your OVS-DPDK question, in the case of Qemu, or I think any VM, the OVS-DPDK socket is mounted into the VM as a VirtIO device. Within the VM it appears as a regular PCI NIC. The kernel detects it and binds it to the kernel driver. At this point it will appear as a regular Netdev interface, but I’m not sure there will be much performance gain as it’s using the kernel driver and hence the kernel network stack.
The purpose of DPDK is to bypass the kernel network stack, this is where the performance comes from, unfortunately it’s also why regular apps can’t use DPDK. So, within the VM we must do as you would with any physical host and physical NIC - you must unbind the NIC from the kernel driver and bind to one of the DPDK drivers like uio_pci_generic, igb_uio or vfio-pci. Now you have a full DPDK interface within your VM, but it’s restricted to DPDK apps only.
I don’t think there is a driver like this for containers. It may be possible to use something like KubeVirt, where you run a lightweight VM within a container. However, if the purpose is to run regular apps on the DPDK interface, then I’m not sure there’s much point as I don’t think there’s much performance to be gained until you bind the DPDK driver.

przemeklal:
> I still have another question about OVS-DPDK. I read some examples about how to set ovs-dpdk with qemu, inside the vm boot by qemu, the network interface is a normal interface from application point of view. Is there any driver for container to do stuff like that? I think it's more user friendly for application that they don't need any dpdk specific knowledge and can take advantage of the high performance network.

Gary's answer is correct, also to expand a bit - a DPDK application running in a container will take use of https://doc.dpdk.org/guides/howto/virtio_user_for_container_networking.html to communicate with OVS dpdkvhostuser sockets.

And yes, while being able to use the regular kernel interfaces in a VM seems user-friendly, almost all performance benefits are going to be neglected in that case because you're coming back to using Linux kernel networking stack which is exactly what you're trying to avoid by using OVS-DPDK, vhost-user, virtio-user and all this good stuff ;) Actually mixing the datapaths like that can cause potential problems, e.g. for many years TCP segmentation offloading wasn't possible with OVS-DPDK, so you'd actually end up with worse performance by combining DPDK on the host and kernel stack in a VM. This is obviously not the case with containers and pure e2e DPDK dataplane such as in this PR.

oilbeater:
@garyloug @przemeklal Thanks for the explanation. I think I got the scenario of using OVS-DPDK.

ftiannew:
so，in OVS-DPDK，what is OVS for ？if you need a dpdk port，why not just assign a DPDK vhost port to it ？
> > I still have another question about OVS-DPDK. I read some examples about how to set ovs-dpdk with qemu, inside the vm boot by qemu, the network interface is a normal interface from application point of view. Is there any driver for container to do stuff like that? I think it's more user friendly for application that they don't need any dpdk specific knowledge and can take advantage of the high performance network.
> 
> Gary's answer is correct, also to expand a bit - a DPDK application running in a container will take use of https://doc.dpdk.org/guides/howto/virtio_user_for_container_networking.html to communicate with OVS dpdkvhostuser sockets.
> 
> And yes, while being able to use the regular kernel interfaces in a VM seems user-friendly, almost all performance benefits are going to be neglected in that case because you're coming back to using Linux kernel networking stack which is exactly what you're trying to avoid by using OVS-DPDK, vhost-user, virtio-user and all this good stuff ;) Actually mixing the datapaths like that can cause potential problems, e.g. for many years TCP segmentation offloading wasn't possible with OVS-DPDK, so you'd actually end up with worse performance by combining DPDK on the host and kernel stack in a VM. This is obviously not the case with containers and pure e2e DPDK dataplane such as in this PR.

garyloug:
Hi @ftiannew you only get the vhost socket with OVS-DPDK, you do not get it without OVS. Without OVS you are dealing with a NIC, so you are using a PCI device (remember it’s not a kernel Netdev). I guess one big advantage of having OVS is for E-W traffic (traffic between 2 pods on the same node). Without OVS, DPDK traffic between two pods must go down to the NIC and then back up. Also, I think if you want OVN integration then you must have OVS? I’m not very familiar with OVN, but I don’t think it can work directly with a NIC?

garyloug:
Hi @oilbeater I have added some documentation on Hugepages, a little on CPU mask and TestPMD and I have done some testing. I have also been working to try reduce the image size, but only pushed a small reduction for now. It’s just some small optimizations, I have not included changes to OVS-DPDK for now. If it’s agreeable to you I think PR is ready now. I will push the image reduction as a separate PR. I need more time to fully understand and test it. I’d also like to revisit why multistage build did not work previously, if I could make this work I think it would make a huge difference.

oilbeater:
@garyloug Thanks! Looking forward for future improvement.

PR #369, opened 2020-06-03
Config support for OVS-DPDK

ipatrykx:
This PR introduces the following changes to OVS-DPDK deployment:

1) Config support

Open vSwitch can be highly configured using `other_configs` options as described in [OpenvSwitch manual](http://www.openvswitch.org/support/dist-docs/ovs-vswitchd.conf.db.5.txt)).

Current deployment uses `dpdk-socket-mem="1024"` in start-ovs-dpdk.sh start script. However this approach means that OVS-DPDK can be only used on single NUMA node machines as OVS requires to allocate specific amount of hugepages for each NUMA node. So for machine with e.g. 2 NUMA nodes it should be  `dpdk-socket-mem="1024,1024"`. Therefore I've decided to create simple OVS-DPDK configuration mechanism to enable configuring OVS for multiple NUMA nodes machines (and other specific configurations) without rebuilding Docker image.

The configuration file called config.cfg is located in /opt/ovs-config on host. This directory is mounted to ovs-ovn pod. It contains list of `other_config` options (same as declared in [OpenvSwitch manual](http://www.openvswitch.org/support/dist-docs/ovs-vswitchd.conf.db.5.txt)). Each option should be placed in new line.

For example:

dpdk-socket-mem="1024,1024"
dpdk-init=true
pmd-cpu-mask=0x4
dpdk-lcore-mask=0x2
dpdk-hugepage-dir=/dev/hugepages

Will enable DPDK with 1024MB of hugepages for NUMA node 0 and NUMA node 1, PMD CPU mask 0x4, lcore mask 0x2 and hugepages in /dev/hugepages.

If file will not exist upon OVS initialization, the default configuration file will be created with values:

dpdk-socket-mem="1024"
dpdk-init=true
dpdk-hugepage-dir=/dev/hugepages

Which is the same as current configuration in start-ovs.dpdk.sh

2) Reworked start-ovs-dpdk.sh

I was testing current OVS-DPDK approach and I stumbled upon problem regarding communication between pods running on master and pods running on worker. For some reason the pods cant connect to each other.

It seems that if 'ovs-ctl restart' is used (as in start-ovs.sh) instead of 'ovs-ctl start' (as in current start-ovs-dpdk.sh) the problem does not occur. Therefore I've decided to rework 'start-ovs-dpdk.sh' to unify it with 'start-ovs.sh'.

3) Separate ovn-dpdk.yaml

As I've mentioned earlier - to enable multiple NUMA node support it's needed to request specific amount of hugepages for ovs-ovn pod. Therefore decided to introduce separate yaml that could be easily customized using kubernetes kustomize. This would ease the process of deployment on multi NUMA node machines.

### DIFF
--- a/dist/images/install.sh
+++ b/dist/images/install.sh
@@ -491,6 +491,8 @@ spec:
               name: host-log-ovs
             - mountPath: /var/log/ovn
               name: host-log-ovn
+            - mountPath: /opt/ovs-config
+              name: host-config-ovs
             - mountPath: /dev/hugepages
               name: hugepage
           readinessProbe:
@@ -542,6 +544,10 @@ spec:
         - name: host-log-ovn
           hostPath:
             path: /var/log/ovn
+        - name: host-config-ovs
+          hostPath:
+            path: /opt/ovs-config
+            type: DirectoryOrCreate
         - name: hugepage
           emptyDir:
             medium: HugePages
--- a/dist/images/start-ovs-dpdk.sh
+++ b/dist/images/start-ovs-dpdk.sh
@@ -8,12 +8,36 @@ function quit {
 }
 trap quit EXIT
 
-# Start ovsdb and vswitchd
-ovs-ctl --no-ovs-vswitchd start
-ovs-vsctl --no-wait set Open_vSwitch . other_config:dpdk-socket-mem="1024"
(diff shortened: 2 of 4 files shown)


oilbeater:
Thanks! @ipatrykx

PR #394, opened 2020-06-30
Solve the problem of non-standard statefulset creation mode

laik:


pkg/controller/pod.go
```diff
@@ -485,7 +485,7 @@ func (c *Controller) handleUpdatePod(key string) error {
 
 func isStatefulSetPod(pod *v1.Pod) (bool, string) {
 	for _, owner := range pod.OwnerReferences {
-		if owner.Kind == "StatefulSet" {
+		if owner.Kind == "StatefulSet" && owner.APIVersion == "apps" {
```
oilbeater, line 488: it should be apps/v1 ?

laik: No, just use apiVersion to distinguish the difference between native resources and CRD resources, the version can be compatible with later statefulsets upgrade v2 or v2alpha1....

oilbeater: I create a StatefulSet and the ownerReferences is below:
 
```ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: StatefulSet
    name: web
    uid: dea1a3a1-0b33-4928-9fb1-52e338ff92da
```
It seems that the apiVersion field contains both the api and the version

oilbeater:
@laik I modified the apiversion match logical. Thanks for the contribution!

PR #518, opened 2020-10-22
feature: support custom vpc

fanriming:
Support custom VPC with annotations, remain the cluster VPC as the default VPC.
Custom VPC not support lb yet.


oilbeater:
Thanks @fanriming, we will review this pr asap

PR #556, opened 2020-11-24
Fix the problem of confusion between old and new versions of crd

xieyanker:
Fix the problem of confusion between old and new versions of crd, mainly as follows:

1. The `.spec.provider` column should be the subnet's, but it is on the ip now.

2. The `vpc` crd is a new feature, but both of the `crd-pre-1.16.yaml` and the `install-pre-1.16.sh` are not be updated.

3. The validation about crd should be the same, but the `crd` with `pre-1.16-crd` are different.


oilbeater:
Thanks! @xieyanker

PR #565, opened 2020-11-29
<feat>(dual stack) dual stack support

feixiang43:
This patch is developed and tested base one kube-ovn 1.2.3 and have not been fully test on the current master branch. I notice that lots of code changed and some new feature had been added, so this pr just sharing some of my past work and should not be merged until fully rebase and test work been done.

Subnet/IP CRD dual stack support in this patch. User can create v4/v6/dual type of subnets, and each type of subnet will allocate corresponding ip family ipAddress. User can get pod ips in pod status IP and IPs filed, which followed the standard k8s defined. 

Following are the pre requirment for using dual stack.
1. Node enable Ipv6 
```
sysctl -w net.ipv6.conf.all.disable_ipv6=0
```
2. Set up join subnet  with dual cidrs in install.sh, only a dual  join subnet support create dual subnet, otherwise datapath will have some problem.
```
# install.sh 
JOIN_CIDR="100.64.0.0/16,,fd00:100:64::/64"
```

By now we can create dual subnet and create pod with it, here is a example yaml file to create Dual  subnet:
```
apiVersion: kubeovn.io/v1
kind: Subnet
metadata:
  name: service-subnet-dual
spec:
  protocol: Dual
  default: false
  namespaces:
  - default
  cidrBlock:
    IPv6: FC00:100:64::/80
    IPv4: 12.15.0.0/16
  private: false
  gatewayType: centralized
  gatewayNode: kube-ovn-worker
  natOutgoing: true
```


oilbeater:
@feixiang43 Thanks! We will take it as a reference

oilbeater:
The related changes have been merged into:
https://github.com/alauda/kube-ovn/pull/577 
https://github.com/alauda/kube-ovn/pull/573

Thanks! @feixiang43

PR #576, opened 2020-12-06
add pod static ip validate

Icedroid:
the pod static ip need to verify is the network number ip or the broadcast ip.


oilbeater (approved):

oilbeater:
Thanks! @Icedroid

PR #586, opened 2020-12-10
Debian: Add debian docker image support

junka:
Add debian docker image support, this patch works with patch in ovn repo
"debian: fix ovn debian packaging and split from openvswitch"

Also, as debian default shell is not bash, set install script cmd to
bash.


oilbeater (approved):

oilbeater:
Thanks! @junka

PR #591, opened 2020-12-11
fix: cni-server default encap ip use right interface ip

Icedroid:
When a k8s node use multiple card, the cni server started for an incorrect interface . If the config.iface is empty, will use the node ip to be tunnel interface encap ip. If the config.iface set, use the config.iface for tunnel interface.

### DIFF
--- a/pkg/daemon/config.go
+++ b/pkg/daemon/config.go
@@ -122,20 +122,39 @@ func ParseFlags() (*Configuration, error) {
 }
 
 func (config *Configuration) initNicConfig() error {
+	var (
+		iface   *net.Interface
+		err     error
+		encapIP string
+	)
+
 	if config.Iface == "" {
-		i, err := getDefaultGatewayIface()
-		if err != nil {
-			return err
-		} else {
-			config.Iface = i
+		podIP, ok := os.LookupEnv("POD_IP")
+		if !ok || podIP == "" {
+			return errors.New("failed to lookup env POD_IP")
 		}
+		iface, err = getIfaceOwnPodIP(podIP)
+		if err != nil {
+			klog.Errorf("failed to find POD_IP iface %v", err)
+			return err
+		}
+		encapIP = podIP
+	} else {
+		iface, err = findInterface(config.Iface)
+		if err != nil {
+			klog.Errorf("failed to find iface %s, %v", config.Iface, err)
+			return err
+		}
+		addrs, err := iface.Addrs()
+		if err != nil {
+			return fmt.Errorf("failed to get iface addr. %v", err)
+		}
+		if len(addrs) == 0 {
+			return fmt.Errorf("iface %s has no ip address", config.Iface)
+		}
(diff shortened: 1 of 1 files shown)


pkg/daemon/config.go
```diff
@@ -122,20 +122,44 @@ func ParseFlags() (*Configuration, error) {
...
+		encapIP string
+	)
+
 	if config.Iface == "" {
 		i, err := getDefaultGatewayIface()
 		if err != nil {
 			return err
 		} else {
-			config.Iface = i
+			iface, err = findInterface(i)
```
oilbeater, line 136: the `config.Iface` here should be the interface that owns the `POD_IP`, not the default gateway

oilbeater (changes requested):

pkg/daemon/config.go
Icedroid: The default gateway only for mtu, must use the interface that owns the POD_IP?

oilbeater: yes, the packet will be tunneled by the interface that owns the encapIP, so the mtu should also coordinate with the same interface

pkg/daemon/config.go
```diff
@@ -238,6 +255,31 @@ func getDefaultGatewayIface() (string, error) {
 	return "", errors.New("unable to find default route")
 }
 
+func getIfaceOwnPodIP(podIP string) (string, error) {
+	routes, err := netlink.RouteList(nil, syscall.AF_INET)
```
oilbeater, line 259: you can use `netlink.LinkList` and `netlink.AddrList` to retrieve the ip address of each interface. The route way might be incorrect if no `src` field is set

oilbeater (changes requested):

Icedroid (commented):
get interface owns POD_IP use netlink.LinkList

pkg/daemon/config.go
```diff
@@ -238,6 +255,27 @@ func getDefaultGatewayIface() (string, error) {
...
+			return "", fmt.Errorf("failed to get a list of IP addresses %v", err)
+		}
+		for _, addr := range addrs {
+			if addr.IPNet.Contains(net.ParseIP(podIP)) && addr.IP.String() == podIP {
+				return link.Attrs().Name, nil
+			}
+		}
+	}
+
+	return "", errors.New("unable to find podIP interface")
```
oilbeater, line 276: it can return netlink directly here, no need to use `findInterface` later to find by name

oilbeater (changes requested):

pkg/daemon/config.go
Icedroid: Good idea.

oilbeater:
@Icedroid  need to squash all commit into one commint, you can check https://www.internalpointers.com/post/squash-commits-into-one-git and use `git push -f` to force update the pr branch

oilbeater (approved):

oilbeater:
Thanks! @Icedroid

PR #603, opened 2020-12-17
modify subnet and ip crds

hongzhen-ma:


pkg/controller/init.go
```diff
@@ -308,3 +310,49 @@ func (c *Controller) initDefaultVlan() error {
...
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+
+	for _, ipCr := range ips.Items {
+		ip := ipCr
+		v4IP, v6IP := util.SplitStringIP(ip.Spec.IPAddress)
```
oilbeater, line 325: If `V4IPAddress` or `V6IPAddress` exists, no need to resync

oilbeater (changes requested):

pkg/controller/init.go
hongzhen-ma: 已经补充检查

oilbeater (approved):

Issue #610, opened 2020-12-22
np unexpected break 


feixiang43:
Hi,  why there is a break here, is it a bug?
https://github.com/alauda/kube-ovn/blob/f299aa7afb225408f4c60e55deba0dc045d28bcf/pkg/controller/network_policy.go#L222

PR #611, opened 2020-12-22
<fix>(np) fix for empty "from" filed networkpolicy

feixiang43:
while np contains multi rules, and some rules only has ports filed, this rule will be skipped.


oilbeater:
#610 same bug exists in egress rules "to" field, can you fix it together

feixiang43:
see new pr https://github.com/alauda/kube-ovn/pull/613

PR #618, opened 2020-12-25
Update the container nic name use the CNI_IFNAME parameter which passed by kubelet

xieyanker:
Update the container nic name use the CNI_IFNAME parameter which passed by kubelet

The default CNI_IFNAME parameter passed by kubelet is "eth0", so we can use this.

Reference link: https://github.com/kubernetes/kubernetes/blob/88a05df5ff1b311e8e92f64b4f3a2c7d4329d14e/pkg/kubelet/dockershim/network/cni/cni.go#L398

### DIFF
--- a/cmd/cni/cni.go
+++ b/cmd/cni/cni.go
@@ -52,6 +52,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 		PodNamespace: podNamespace,
 		ContainerID:  args.ContainerID,
 		NetNs:        args.Netns,
+		IfName:       args.IfName,
 		Provider:     netConf.Provider,
 		DeviceID:     netConf.DeviceID,
 	})
--- a/pkg/daemon/handler.go
+++ b/pkg/daemon/handler.go
@@ -101,8 +101,8 @@ func (csh cniServerHandler) handleAdd(req *restful.Request, resp *restful.Respon
 	}
 
 	if podRequest.Provider == util.OvnProvider {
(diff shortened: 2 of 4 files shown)


xieyanker:
@oilbeater
The e2e test failed: `pods "normal-ip" is forbidden: error looking up service account static-ip/default: serviceaccount "default" not found`. But it's not caused by me, how to retest it?

oilbeater:
> @oilbeater
> The e2e test failed: `pods "normal-ip" is forbidden: error looking up service account static-ip/default: serviceaccount "default" not found`. But it's not caused by me, how to retest it?

@xieyanker Thanks I have triggered the e2e again

oilbeater:
Thanks! @xieyanker

oilbeater:
related to https://github.com/alauda/kube-ovn/issues/601

feixiang43:
Hi, have you test multi cni with kube-ovn in your environment? I think controller pod handler logical should also do some adaption.

xieyanker:
> Hi, have you test multi cni with kube-ovn in your environment? I think controller pod handler logical should also do some adaption.

@feixiang43  Yeah, what I want to do is to use kube-ovn as the second network card, and the calico as the main cni. But now there are many conflicts, this pr only fix the nic name's issue.

 In addition, the default route in the pod is also in conflict. 

As for the controller, I think it may also need to be modified. According to the current logic, as long as a new pod is created, a bunch of annotation parameters will be added to the pod. If I don’t use kube-ovn, these won’t be useful at all.

There may be other issues, but I haven’t considered. But I think my usage may be special, and I may not submit some modifications to the open source community. If you are interested in these or have related needs, we can communicate.

PR #631, opened 2021-01-05
modify test problem for dual-stack upgrade

hongzhen-ma:


pkg/controller/subnet.go
```diff
@@ -300,6 +296,31 @@ func formatSubnet(subnet *kubeovnv1.Subnet, c *Controller) error {
 	return nil
 }
 
+func checkSubnetChanged(subnet *kubeovnv1.Subnet) (bool, error) {
+	var err error
+	changed := false
```
oilbeater, line 301: `ret := false`
后续判断
`if changed {
  ret = ture
}`
不需要来回判断 changed 和 forced 是否一致

oilbeater (changes requested):

pkg/controller/subnet.go
hongzhen-ma: 已经修改

oilbeater (approved):

PR #644, opened 2021-01-15
fix: make sure northd leader change

oilbeater:
When node power down, the exist connection will not be closed. If no inactivity_probe set,
the stale northd will still be the leader and all process from nb to sb will stop. Change the
inactivity_probe to 3min to fix this.


halfcrazy:
There's some `inactivity_probe=0` left in this file. Don't they need to be modified?

oilbeater:
> There's some `inactivity_probe=0` left in this file. Don't they need to be modified?

All of them should be changed

halfcrazy (approved):

PR #664, opened 2021-01-30
Fix: check kube-proxy's 10256 port healthz

xieyanker:
Fix https://github.com/alauda/kube-ovn/issues/654

### DIFF
--- a/dist/images/install-pre-1.16.sh
+++ b/dist/images/install-pre-1.16.sh
@@ -2084,7 +2084,7 @@ checkDaemonSet(){
 }
 
 checkKubeProxy(){
-  healthResult=`kubectl get node -o wide | grep -v "INTERNAL-IP" | awk '{printf "curl -sL -w %{http_code} http://%s:10256/healthz -o /dev/null | grep -v 200 \n", $6}' | bash` || true
+  healthResult=`kubectl get node -o wide | grep -v "INTERNAL-IP" | awk '{printf "curl -g -6 -sL -w %{http_code} http://%s:10256/healthz -o /dev/null | grep -v 200 \n", $6}' | bash` || true
   if [ -n "$healthResult" ]; then
     echo "kube-proxy's health check failed"
     exit 1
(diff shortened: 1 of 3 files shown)


dist/images/kubectl-ko
```diff
@@ -263,6 +263,14 @@ checkDeployment(){
   fi
 }
 
+checkKubeProxy(){
+  healthResult=`kubectl get node -o wide | grep -v "INTERNAL-IP" | awk '{printf "curl -sL -w %{http_code} http://%s:10256/healthz -o /dev/null | grep -v 200 \n", $6}' | bash` || true
```
oilbeater, line 267: ipv6 e2e test failed. Should use `curl -g -6` for ipv6 cluster

oilbeater (changes requested):

dist/images/kubectl-ko
xieyanker: @oilbeater Thanks for your reply! However, I don’t have an ipv6 cluster. What I want to know is whether `curl -g -6` is also valid for ipv4 cluster. Should I need different commands for these two clusters?

oilbeater: After some tests, these options also work for ipv4 cluster. We can go on

dist/images/install-pre-1.16.sh
```diff
@@ -2083,6 +2083,14 @@ checkDaemonSet(){
   fi
 }
 
+checkKubeProxy(){
+  healthResult=`kubectl get node -o wide | grep -v "INTERNAL-IP" | awk '{printf "curl -g -6 -sL -w %{http_code} http://%s:10256/healthz -o /dev/null | grep -v 200 \n", $6}' | bash` || true
```
oilbeater, line 2087: `http://[%s]:10256/healthz` square brackets is needded

oilbeater (changes requested):

oilbeater (approved):

oilbeater:
Thanks! @xieyanker

PR #668, opened 2021-02-02
Fix: remove IsNotFound when get configmap external gateway

xieyanker:
Fix: remove IsNotFound when get configmap external gateway

In my view, if `enable` is `true`, the configmap `ovn-external-gw-config` must exist. 

Even if it doesn't exist, we should return the error, otherwise, a panic about nil pointer will be reported  when use `cm.Data`


oilbeater (approved):

oilbeater:
Thanks! @xieyanker

PR #679, opened 2021-02-08
ovn-central: set default db addr same with leader node 

junka:
fix nb and sb ovsdb server error 'bind: Address already in use'
which cause too much logs


oilbeater:
Thanks! @junka

PR #688, opened 2021-02-16
Change ipv6 disabled in install.md

christiangonre:
Change enabled ipv6 with disabled ipv6 in install.md


oilbeater:
@christiangonre the kernel fix is in 5.0, most linux distribution still with the tunnel bug, so the ipv6 should NOT be disabled

christiangonre:
@oilbeater Thanks for the info, I think that the Prerequest part of the install.md maybe can confuse the users, that´s why I did this pull request because I thought it was a mistake.
If you take a look,in the Prerequest says that the Kernel must boot with ipv6.disable and in the Notes says that the ipv6 must be enabled.

PR #697, opened 2021-02-21
fix checkSBBindings error when hostname is not nodeName

junka:
in some case the k8s spec.nodeName is not same with a `hostname`,
then the sbctl could not get the chassis uuid, the check process
will fail here.
A workaroud solution is that try to get chassis name first from encap
table and then do the left lookup.
It is not a good solution. When we use a second nic and change the encap
ip to the second nic, it could not work either.

Signed-off-by: Wan Junjie <wanjunjie@bytedance.com>


oilbeater:
The `start-ovs.sh` will use nodeName to set the hostname in ovn-sb. They should be same in theory
https://github.com/kubeovn/kube-ovn/blob/8c5ae3131711a350852aadc5aaefd6522123bccf/dist/images/start-ovs.sh#L100

junka:
@oilbeater you are right，will do that. Close.

PR #698, opened 2021-02-21
Update install.sh

yanweininenine:
Optimize Command


oilbeater:
Thanks! @yanweininenine

PR #712, opened 2021-03-06
Add 'kubectl ko trace' command's default namespace

xieyanker:


oilbeater (changes requested):
can you add it to `install.sh`

oilbeater:
Thanks @xieyanker !

PR #716, opened 2021-03-13
refactor: remove unnecessary config logic

xieyanker:
As far as I know, both of the `config.ExternalAddress` and `config.ExternalDNS` can get the value from the  cmd arguments. I don't know why need to take the value from the daemonset's template again.

whether the `config.Mode` is `job` or `server`, the logic should be the same.

### DIFF
--- a/pkg/pinger/config.go
+++ b/pkg/pinger/config.go
@@ -1,14 +1,11 @@
 package pinger
 
 import (
-	"context"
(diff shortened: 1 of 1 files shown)


oilbeater (approved):

oilbeater:
Thanks! @xieyanker

PR #718, opened 2021-03-15
feat: multus ovn nic

fanriming:
# Kubeovn for Multus-CNI
Full support for multiple nic, not just IPAM.
### 1.Create a Kube-OVN subnet
It is no different from regular subnet.
```
apiVersion: kubeovn.io/v1
kind: Subnet
metadata:
  name: my-subnet
spec:
  protocol: IPv4
  cidrBlock: 172.17.0.0/16
  gateway: 172.17.0.1
  excludeIps:
  - 172.17.0.0..172.17.0.10
```
### 2.Create network attachment definition
```
apiVersion: "k8s.cni.cncf.io/v1"
kind: NetworkAttachmentDefinition
metadata:
  name: attachnet
  namespace: default
spec:
  config: '{
      "cniVersion": "0.3.0",
      "type": "kube-ovn",
      "server_socket": "/run/openvswitch/kube-ovn-daemon.sock",
      "provider": "attachnet.default.ovn"
    }'
```
**type**: Should be `kube-ovn` to invoke Kube-OVN plugin to fetch the ovn subnet.
**server_socket**: Is the socket file that Kube-OVN plugin communicate with. Default location is `/run/openvswitch/kube-ovn-daemon.sock`.
**provider**: The `<name>.<namespace>.ovn` of this NetworkAttachmentDefinition, The kube-OVN plug-in will use it later to determine whether a native OVN subnet should be used. Be sure to add the OVN suffix.

### 3.Create Pod with multi network
For random allocation from ovn-default:
```
...
  annotations:
    k8s.v1.cni.cncf.io/networks: default/attachnet
...
```
For allocation from the specified subnet:
```
...
  annotations:
    k8s.v1.cni.cncf.io/networks: attachnet
    attachnet.default.ovn.kubernetes.io/logical_switch: my-subnet
...
```
For allocation from the static ip:
```
...
  annotations:
    k8s.v1.cni.cncf.io/networks: attachnet
    attachnet.default.ovn.kubernetes.io/logical_switch: my-subnet
    attachnet.default.ovn.kubernetes.io/ip_address: 172.17.0.20
...
```


pkg/controller/pod.go
```diff
@@ -469,22 +494,62 @@ func (c *Controller) handleDeletePod(key string) error {
...
+			}
+		}
 	}
-
 	c.ipam.ReleaseAddressByPod(key)
 	return nil
 }
 
+func (c *Controller) getPodPorts(pod, namespace string) (ports []string, err error) {
+	defaultPort := ovs.PodNameToPortName(pod, namespace, util.OvnProvider)
```
oilbeater, line 520: we can set lsp external_ids with the pod namespace/name, so we can more clearly filter the lsp here

pkg/daemon/handler.go
```diff
@@ -201,8 +203,8 @@ func (csh cniServerHandler) handleDel(req *restful.Request, resp *restful.Respon
 	}
 
 	klog.Infof("delete port request %v", podRequest)
-	if podRequest.Provider == util.OvnProvider {
-		err = csh.deleteNic(podRequest.PodName, podRequest.PodNamespace, podRequest.ContainerID, podRequest.DeviceID)
+	if podRequest.Provider == util.OvnProvider || podRequest.CniType == util.CniTypeName {
```
oilbeater, line 206: check `CniType` should be enough here

pkg/controller/pod.go
```diff
@@ -730,48 +815,89 @@ func (c *Controller) getPodAttachmentSubnet(pod *v1.Pod) ([]*kubeovnv1.Subnet, e
 		return nil, err
 	}
 
-	result := make([]*kubeovnv1.Subnet, 0, len(attachments))
+	result := make([]*kubeovnNet, 0, len(attachments))
 	for _, attach := range attachments {
-		provider := fmt.Sprintf("%s.%s", attach.Name, attach.Namespace)
+		networkClient := c.config.AttachNetClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(attach.Namespace)
```
oilbeater, line 820: We prefer to use informer lister cache than client directly calls to enhance performance. However if use informer we require that the attachment crd exists otherwise it will fail to start the informer, users who haven't install multus will panic.

oilbeater: We can install multus crds by default later

oilbeater:
rebase is needed to resovle the confilcts @fanriming

pkg/daemon/handler.go
fanriming: By default, there is no Multus-CNI and CniType is empty, so I think the OVNProvider still needs to be checked here

pkg/controller/pod.go
fanriming: Rright, I will improve according to the suggestion

pkg/controller/pod.go
fanriming: the best case scenario is to install multus by default, then we can use informerlister cache herer, but I'm afraid it won't be very friendly to users who don't need multus.

pkg/daemon/ovs.go
```diff
@@ -159,6 +162,11 @@ func configureContainerNic(nicName, ifName string, ipAddr, gateway string, macAd
 			return err
 		}
 
+		if ifName != "eth0" {
```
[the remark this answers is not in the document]
fanriming: yes, flexible router configuration for the network card has not been considered yet, I think configuring the default router for the default network card is probably a more common configuration, and multus routes can be considered later.

oilbeater (approved):

oilbeater:
Thanks ! @fanriming Can you add the user guide to docs later

fanriming:
> Thanks ! @fanriming Can you add the user guide to docs later

ok, i will add the user guide to docs in another PR later

PR #719, opened 2021-03-16
Update vendor

xieyanker:
Some files are missing in the current vendor package, compiling will fail if I don't use go mod.

I updated the vendor the vendor by command `go mod tidy` and `go mod vendor`.


oilbeater (approved):

oilbeater:
Thanks! @xieyanker

PR #725, opened 2021-03-24
livenessprobe fail if ovn nb/ovn sb not running

halfcrazy:
ovn-ctl status_ovnnb
not running

### DIFF
--- a/dist/images/ovn-healthcheck.sh
+++ b/dist/images/ovn-healthcheck.sh
@@ -5,5 +5,5 @@ shopt -s expand_aliases
 alias ovn-ctl='/usr/share/ovn/scripts/ovn-ctl'
 
 ovn-ctl status_northd
-ovn-ctl status_ovnnb
-ovn-ctl status_ovnsb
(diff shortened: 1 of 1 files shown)


dist/images/ovn-healthcheck.sh
```diff
@@ -5,5 +5,5 @@ shopt -s expand_aliases
 alias ovn-ctl='/usr/share/ovn/scripts/ovn-ctl'
 
 ovn-ctl status_northd
-ovn-ctl status_ovnnb
-ovn-ctl status_ovnsb
+ovn-ctl status_ovnnb | grep -q -v not-running
```
oilbeater, line 8: If the ovn-nb/sb is not running, why we treat it as healty?

halfcrazy: -v here indicates we exit with a none zero code if we found status not-running.

ovn-ctl status_ovnxb will always exit with 0.
https://github.com/ovn-org/ovn/blob/master/utilities/ovn-ctl#L366

oilbeater (approved):

PR #738, opened 2021-04-08
Support ovn as the default attache network

fanriming:
Support ovn as the default attache network, sriov was also fixed.


oilbeater:
Thanks! @fanriming

PR #745, opened 2021-04-15
check chassis when node joining cluster

lut777:
check chassis-id at local config file and annotate chassis-id in node.yaml
controller check whether the annotation and local sbdb table are the same, if not, reset local table chassis.
To ensure chassis annotation, addNode/updateNode will always keep a chassis annotation from table

### DIFF
--- a/cmd/daemon/cniserver.go
+++ b/cmd/daemon/cniserver.go
@@ -125,7 +125,7 @@ func initChassisAnno(cfg *daemon.Configuration) error {
 	}
 
 	chassistr := string(chassisID)
-	node.Annotations[util.ChassisAnnotation] = strings.Replace(chassistr, "\n", "", -1)
+	node.Annotations[util.ChassisAnnotation] = strings.TrimSpace(chassistr)
 	patchPayloadTemplate :=
 		`[{
         "op": "%s",
--- a/cmd/daemon/cniserver.go
+++ b/cmd/daemon/cniserver.go
@@ -1,10 +1,16 @@
 package daemon
 
 import (
+	"context"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"k8s.io/apimachinery/pkg/types"
 	"net/http"
 	_ "net/http/pprof" // #nosec
+	"os"
+	"strings"
+	"time"
 
 	"github.com/kubeovn/kube-ovn/pkg/util"
 	"github.com/kubeovn/kube-ovn/versions"
@@ -24,6 +30,11 @@ func CmdMain() {
 	klog.Infof(versions.String())
 	daemon.InitMetrics()
 	config, err := daemon.ParseFlags()
+
+	if err := Retry(util.ChasRetryTime, util.ChasRetryIntev, initChassisAnno, config); err != nil {
+		klog.Fatalf("failed to annotate chassis id, %v", err)
+	}
+
 	if err != nil {
 		klog.Fatalf("parse config failed %v", err)
 	}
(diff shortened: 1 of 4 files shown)


cmd/daemon/cniserver.go
```diff
@@ -79,3 +90,56 @@ func mvCNIConf() error {
...
+		time.Sleep(time.Duration(sleep) * time.Second)
+	}
+	return err
+}
+
+func initChassisAnno(cfg *daemon.Configuration) error {
+
+	chassisID, err := ioutil.ReadFile(util.ChassisLoc)
+	if err != nil {
+		klog.V(3).Infof("read chassis file failed, %v", err)
```
oilbeater, line 112: klog.Errorf

cmd/daemon/cniserver.go
```diff
@@ -79,3 +90,56 @@ func mvCNIConf() error {
...
+		return err
+	}
+
+	hostname := os.Getenv(util.HostnameEnv)
+	if hostname == "" {
+		klog.V(3).Infof("env %s does not exist", util.HostnameEnv)
```
oilbeater, line 118: Errorf

pkg/controller/node.go
```diff
@@ -203,6 +204,10 @@ func (c *Controller) handleAddNode(key string) error {
 		return err
 	}
 
+	if c.checkChassisDupl(key, node) != nil {
```
oilbeater, line 207: err := c.checkChassisDupl(key, node); err != nil

pkg/controller/node.go
```diff
@@ -485,3 +512,24 @@ func (c *Controller) checkNodeEcmpRouteExist(nodeIp, cidrBlock string) (bool, er
...
+	// notice that multiple chassises may arise and we are not prepared
+	chassisAdd, err := c.ovnClient.GetChassis(key)
+	if err != nil {
+		klog.Errorf("failed to get node %s chassisID, %v", key, err)
+		return err
+	}
+
+	chassisAnn := node.Annotations[util.ChassisAnnotation]
+	if chassisAnn != "" && chassisAnn != chassisAdd {
+		klog.V(3).Infof("duplicate chassis for node %s and new chassis %s", key, chassisAdd)
```
oilbeater, line 526: klog.Infof

pkg/controller/node.go
oilbeater: the key here is equal to node.Name, no need for the extra arg

pkg/controller/node.go
```diff
@@ -254,6 +259,13 @@ func (c *Controller) handleAddNode(key string) error {
 		op = "add"
 	}
 
+	chassisAdd, err := c.ovnClient.GetChassis(key)
```
oilbeater, line 262: no need to add annotation here

pkg/controller/node.go
```diff
@@ -317,6 +329,21 @@ func (c *Controller) handleUpdateNode(key string) error {
 		return err
 	}
 
+	if c.checkChassisDupl(key, node) != nil {
+		return err
+	}
+	chassisAdd, err := c.ovnClient.GetChassis(key)
```
oilbeater, line 335: same as above

oilbeater (changes requested):

lut777:
done

cmd/daemon/cniserver.go
```diff
@@ -79,3 +90,56 @@ func mvCNIConf() error {
...
+		return err
+	}
+	node, err := cfg.KubeClient.CoreV1().Nodes().Get(context.Background(), hostname, v1.GetOptions{})
+	if err != nil {
+		klog.Errorf("failed to get node %s %v", hostname, err)
+		return err
+	}
+
+	chassistr := string(chassisID)
+	node.Annotations[util.ChassisAnnotation] = strings.Replace(chassistr, "\n", "", -1)
```
oilbeater, line 128: strings.TrimSpace(chassistr)

oilbeater (changes requested):

lut777:
add String.TrimSpace after removing extra \n

oilbeater (approved):

PR #749, opened 2021-04-17
fix: configure nic failed when ifname empty

danieldin95:
we get config nic failed with kube-ovn:v1.6.1 in kubernetes v1.20.5. and this patch will set defualt value:eth0 if podRequest.IfName is empty.


oilbeater (approved):

oilbeater:
Thanks! @danieldin95

PR #755, opened 2021-04-22
fix SNAT on pod startup

### LINKED ISSUE #732: External network past SNAT unavailable on pod start

zhangzujian:
Fix #732 by updating ipset *local-pod-ip-nat* when adding/deleting nic.

### DIFF
--- a/pkg/daemon/gateway.go
+++ b/pkg/daemon/gateway.go
@@ -106,6 +106,60 @@ func (c *Controller) setIPSet() error {
 	return nil
 }
 
+func (c *Controller) addIPSetMembers(setID, subnet, ip string) error {
+	podSubnet, err := c.subnetsLister.Get(subnet)
+	if err != nil {
+		klog.Errorf("get subnet %s failed, %+v", subnet, err)
+		return err
+	}
+
+	if !podSubnet.Spec.NatOutgoing ||
+		podSubnet.Spec.Vpc != util.DefaultVpc ||
+		podSubnet.Spec.GatewayType != kubeovnv1.GWDistributedType {
+		return nil
+	}
+
+	podIPs := strings.Split(ip, ",")
+	if protocol := util.CheckProtocol(ip); protocol == kubeovnv1.ProtocolDual {
+		c.ipset[kubeovnv1.ProtocolIPv4].AddMembers(setID, []string{podIPs[0]})
+		c.ipset[kubeovnv1.ProtocolIPv6].AddMembers(setID, []string{podIPs[1]})
+		c.ipset[kubeovnv1.ProtocolIPv4].ApplyUpdates()
+		c.ipset[kubeovnv1.ProtocolIPv6].ApplyUpdates()
+	} else {
+		c.ipset[protocol].AddMembers(setID, []string{podIPs[0]})
+		c.ipset[protocol].ApplyUpdates()
+	}
+
+	return nil
+}
+
+func (c *Controller) removeIPSetMembers(setID, subnet, ip string) error {
+	podSubnet, err := c.subnetsLister.Get(subnet)
+	if err != nil {
+		klog.Errorf("get subnet %s failed, %+v", subnet, err)
+		return err
+	}
+
+	if !podSubnet.Spec.NatOutgoing ||
+		podSubnet.Spec.Vpc != util.DefaultVpc ||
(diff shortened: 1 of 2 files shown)


pkg/daemon/gateway.go
```diff
@@ -106,6 +106,88 @@ func (c *Controller) setIPSet() error {
 	return nil
 }
 
+func (c *Controller) addIPSetMembers(setID, subnet, ip string) {
```
oilbeater, line 109: should return err if something goes wrong

pkg/daemon/gateway.go
```diff
@@ -106,6 +106,88 @@ func (c *Controller) setIPSet() error {
...
+
+	protocols := make([]string, 2)
+	if c.protocol == kubeovnv1.ProtocolDual {
+		protocols[0] = kubeovnv1.ProtocolIPv4
+		protocols[1] = kubeovnv1.ProtocolIPv6
+	} else {
+		protocols[0] = c.protocol
+	}
+
+	podIPs := strings.Split(ip, ",")
```
oilbeater, line 130: Check ip protocol directly and add it to related ipset, no need to create `protocols` array

pkg/daemon/gateway.go
```diff
@@ -106,6 +106,88 @@ func (c *Controller) setIPSet() error {
...
```
oilbeater, line 171: same as above

pkg/daemon/handler.go
```diff
@@ -116,6 +116,8 @@ func (csh cniServerHandler) handleAdd(req *restful.Request, resp *restful.Respon
 			}
 			return
 		}
+
+		csh.Controller.addIPSetMembers(LocalPodSet, subnet, ip)
```
oilbeater, line 120: should check error here

pkg/daemon/handler.go
```diff
@@ -208,6 +210,10 @@ func (csh cniServerHandler) handleDel(req *restful.Request, resp *restful.Respon
 
 	klog.Infof("delete port request %v", podRequest)
 	if podRequest.Provider == util.OvnProvider || podRequest.CniType == util.CniTypeName {
+		subnet := pod.Annotations[fmt.Sprintf(util.LogicalSwitchAnnotationTemplate, podRequest.Provider)]
+		ip := pod.Annotations[fmt.Sprintf(util.IpAddressAnnotationTemplate, podRequest.Provider)]
+		csh.Controller.removeIPSetMembers(LocalPodSet, subnet, ip)
```
oilbeater, line 215: check error here

oilbeater (changes requested):

pkg/daemon/gateway.go
```diff
@@ -106,6 +106,68 @@ func (c *Controller) setIPSet() error {
...
+		return err
+	}
+
+	if !podSubnet.Spec.NatOutgoing ||
+		podSubnet.Spec.Vpc != util.DefaultVpc ||
+		podSubnet.Spec.GatewayType != kubeovnv1.GWDistributedType {
+		return nil
+	}
+
+	protocols := make([]string, 2)
```
oilbeater, line 122: no need here

oilbeater (changes requested):

oilbeater:
please squash to one commit

oilbeater (approved):

PR #761, opened 2021-04-26
feature: forward egress traffic to external gateway

zhangzujian:
Gateway node(s) forwards egress traffic to external gateway based on policy routing. Supports both distributed gateway and centralized gateway modes. New properties are available in subnet spec: externalGateway, policyRoutingPriority & policyRoutingTableID.

Example:

```yaml
apiVersion: kubeovn.io/v1
kind: Subnet
metadata:
  name: subnet1
spec:
  cidrBlock: 192.168.1.0/24
  default: false
  gateway: 192.168.1.1
  gatewayType: centralized
  gatewayNode: kube-ovn-worker
  natOutgoing: false
  private: false
  externalGateway: 172.18.0.1
  policyRoutingPriority: 2000
  policyRoutingTableID: 2000
  namespaces:
    - ns1
```

### DIFF
--- a/pkg/daemon/controller.go
+++ b/pkg/daemon/controller.go
@@ -2,37 +2,38 @@ package daemon
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
+	"os"
 	"os/exec"
+	"reflect"
 	"strings"
+	"syscall"
 	"time"
-
-	"k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/client-go/tools/record"
-
-	"k8s.io/apimachinery/pkg/api/errors"
-
-	"github.com/kubeovn/kube-ovn/pkg/ovs"
 
 	"github.com/alauda/felix/ipsets"
 	"github.com/coreos/go-iptables/iptables"
 	"github.com/vishvananda/netlink"
 	v1 "k8s.io/api/core/v1"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes/scheme"
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	listerv1 "k8s.io/client-go/listers/core/v1"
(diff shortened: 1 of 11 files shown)


zhangzujian:
@oilbeater please review

pkg/daemon/controller.go
```diff
@@ -181,12 +177,69 @@ func (c *Controller) processNextSubnetWorkItem() bool {
...
+
+	// handle policy routing
+	rulesToAdd, rulesToDel, routesToAdd, routesToDel, err := c.diffPolicyRouting(oldSubnet, newSubnet)
+	if err != nil {
+		klog.Errorf("failed to get policy routing difference: %v", err)
+		return err
+	}
+	// add new routes first
+	for _, r := range routesToAdd {
+		if err = netlink.RouteAdd(&r); err != nil && !errors.Is(err, syscall.EEXIST) {
```
oilbeater, line 210: what will happen, if the route already exists in host

pkg/daemon/controller.go
```diff
@@ -181,12 +177,69 @@ func (c *Controller) processNextSubnetWorkItem() bool {
...
+	}
+	if event.new != nil {
+		if newSubnet, ok = event.new.(*kubeovnv1.Subnet); !ok {
+			klog.Errorf("expected new subnet in subnetEvent but got %#v", event.new)
+			return nil
+		}
+	}
+
+	// handle policy routing
+	rulesToAdd, rulesToDel, routesToAdd, routesToDel, err := c.diffPolicyRouting(oldSubnet, newSubnet)
```
oilbeater, line 203: why not just compare to exist routes and rules in the host?

pkg/daemon/controller.go
zhangzujian: `!errors.Is(err, syscall.EEXIST)` ensures the existing route will not be replaced.

pkg/daemon/controller.go
zhangzujian: In case the controller deletes rules or routes which are created by users or by other processes.

pkg/daemon/controller.go
zhangzujian: > what will happen, if the route already exists in host

@oilbeater  `!errors.Is(err, syscall.EEXIST)` ensures the existing route will not be replaced.

pkg/daemon/controller.go
zhangzujian: > why not just compare to exist routes and rules in the host?

@oilbeater In case the controller deletes rules or routes which are created by users or by other processes.

oilbeater (approved):

PR #766, opened 2021-04-28
handle update deployment vpc-nat-gw

xieyanker:
1. The function `handleAddOrUpdateVpcNatGw` only create the deployment when the deployment doesn't exist, I think the deployment should also be updated when it already exists.
2. The container name `vpc-gat-gw` is a typo about `vpc-nat-gw`?
3. The labelSelector should include the `util.VpcNatGatewayLabel: "true"`, to fully match the pod.


oilbeater:
@fanriming can you take a look at it

fanriming:
@oilbeater I think it works

oilbeater (approved):

PR #785, opened 2021-05-07
feat: add holdtime function

withlin:
### Hold Time 
The Hold Time attribute sets the Hold Timer in seconds for each BGP neighbor. Upon receipt of an UPDATE or KEEPALIVE, the Hold Timer resets to the initial value. If the Hold Timer reaches zero, the BGP session is torn down, routes from that neighbor are removed, and an appropriate update route withdraw message is sent to other BGP neighbors for the impacted prefixes. The Hold Time is a heartbeat mechanism for BGP neighbors to ensure that the neighbor is healthy and alive.

### Why add holdtime function
* If you want the BGP neighbor route to disappear as soon as possible after the ovn-speaker goes down, which the holdtime should be setting smaller.
* In some scenarios, after the ovn-speaker goes down, the pod/subnet network is normal and can still communicate with the outside world. You want BGP neighbor routing to last longer, so in this case you can set the holdtime to be larger.

### DIFF
--- a/pkg/speaker/config.go
+++ b/pkg/speaker/config.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"time"
 
 	clientset "github.com/kubeovn/kube-ovn/pkg/client/clientset/versioned"
 	api "github.com/osrg/gobgp/api"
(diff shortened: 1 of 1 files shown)


oilbeater (approved):

oilbeater:
Thanks! @withlin

PR #799, opened 2021-05-08
Fix the different time from container and host

caoyingjunz:
The kube-ovn containers will use CST time by default, this PS to use
host's localtime instead

more time info:
host time:  2021年 05月 09日 星期日 00:24:17 CST
container time: Sat May  8 16:24:15 UTC 2021


oilbeater:
@caoyingjunz Can you also update `install.sh` and `install-pre-1.16.sh`

caoyingjunz:
> @caoyingjunz Can you also update `install.sh` and `install-pre-1.16.sh`

Done

oilbeater:
Thanks! @caoyingjunz

PR #800, opened 2021-05-09
Remove the unnecessary rm command

caoyingjunz:
00-kube-ovn.conflist not exists in /etc/cni/net.d, no needed to remove it.


oilbeater:
Thanks! @caoyingjunz

PR #812, opened 2021-05-11
add judge before use index

pengbinbin1:
- API changes

add judge before use slices's index avoid crash

/kind <bug>


oilbeater:
Thanks! @pengbinbin1

PR #815, opened 2021-05-11
Check and Fetch all ValidatePodNetwork errors

caoyingjunz:
ValidatePodNetwork return directly when one error occur, and the
next Validates will not happen.

This PS validates all the annotations, then aggregate errors together.


oilbeater (approved):

oilbeater:
Thanks! @caoyingjunz

PR #848, opened 2021-06-02
Dev

pengbinbin1:
API changes
add judgw before use slices's index,if the index is out of the length then is will be crashed
/kind
Supported kinds are: bug, cleanup, design, documentation, failing-test, feature, flake, misc, question, tep
-->


oilbeater:
@pengbinbin1 please squash the commits into one commit

PR #852, opened 2021-06-03
add judge before use slices's index

pengbinbin1:
API changes
add judgw before use slices's index,if the index is out of the length then is will be crashed
/kind
Supported kinds are: bug, cleanup, design, documentation, failing-test, feature, flake, misc, question, tep
-->


oilbeater (approved):

oilbeater:
Thanks! @pengbinbin1

PR #863, opened 2021-06-08
unify logical entity list funcs

lut777:
Unify list func for list logical_switch and logical_router.
Add external_id when adding logical_router.

### DIFF
--- a/pkg/ovs/ovn-nbctl.go
+++ b/pkg/ovs/ovn-nbctl.go
@@ -314,10 +314,16 @@ func (c Client) DeleteGatewaySwitch(name string) error {
 }
 
 // ListLogicalSwitch list logical switch names
-func (c Client) ListLogicalSwitch() ([]string, error) {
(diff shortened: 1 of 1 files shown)


pkg/ovs/ovn-nbctl.go
```diff
@@ -427,6 +418,10 @@ func (c Client) DeleteLogicalSwitch(ls string) error {
 // CreateLogicalRouter delete logical router in ovn
 func (c Client) CreateLogicalRouter(lr string) error {
 	_, err := c.ovnNbCommand(MayExist, "lr-add", lr)
+	if err == nil {
```
oilbeater, line 421: combine `lr-add` and `set` in one command

oilbeater (changes requested):

pkg/ovs/ovn-nbctl.go
lut777: done.
add agrs to list entities.

lut777: I can't find all git logs in my record...
may need help with squash and merge.

oilbeater (approved):

PR #885, opened 2021-06-20
Fix multicast when nic type is OVS internal port

### LINKED ISSUE #881: Multicast not working when nic type is internal-port
# Expected Behavior

# Actual Behavior

1. Create two pods with `internal-port`, or one pod with `internal-port` and the other with `veth-pair`;
2. Run `omping -m 233.8.5.1 172.18.0.17 172.18.0.16` in both pods to test multicast connectivity (Pod CIDR 172.18.0.0/16); 

`omping` in one pod exits with the following output:

```txt
172.18.0.17 : waiting for response msg
172.18.0.17 : joined (S,G) = (*, 233.8.5.1), pinging
172.18.0.17 :   unicast, seq=1, size=69 bytes, dist=0, time=0.179ms
omping: Cannot send message: Network is unreachable
```

Verbose output is as follow:

```txt
msgsend.c:67 Sending unicast answer msg to 172.18.0.17
msgsend.c:82 Sending multicast answer msg to 233.8.5.1
rsfunc.c:251 sendto error - errno = 101
omping: Cannot send message: Network is unreachable
``` 

# Additional Info

- Kubernetes version:

- kube-ovn version:   1.7.0+

When pod nic type is `internal-port`, an extra dummy interface with the same IP address will be created. The dummy interface does not affect normal TCP/UDP functionality since it's down, but multicast can not work.

I created a `bridge` interface and moved IP address from the dummy interface to the bridge, multicast did not work as well. However, after deleting IP address on the OVS internal port and readding it, multicast works fine.

I cannot figure out the reason for now. But we can fix the misbehavior temporarily by removing the dummy interface or by assigning IP address(es) to the OVS internal port after configuring the dummy interface.

@oilbeater @hongzhen-ma

zhangzujian:
Fix #881

### DIFF
--- a/pkg/daemon/ovs.go
+++ b/pkg/daemon/ovs.go
@@ -175,13 +175,13 @@ func configureContainerNic(nicName, ifName string, ipAddr, gateway string, macAd
 		}
 
 		if nicType == util.InternalType {
-			if err = configureNic(nicName, ipAddr, macAddr, mtu); err != nil {
-				return err
-			}
 			if err = addAdditonalNic(ifName); err != nil {
 				return err
 			}
 			if err = configureAdditonalNic(ifName, ipAddr); err != nil {
+				return err
+			}
+			if err = configureNic(nicName, ipAddr, macAddr, mtu); err != nil {
 				return err
 			}
 		} else {


oilbeater (approved):

oilbeater:
Nice try!

PR #903, opened 2021-07-03
Underlay/VLAN refactoring

zhangzujian:
Changes and new features:

1. Introduce new CRD `ProviderNetwork` for underlay/vlan networking management;
2. Specify provider inetrface on nodes in provider network;
4. Exclude nodes in provider network;
5. Link local IPv6 address and route will NOT be transferred to OVS bridge;
6. Builtin hybrid network support - create/update/delete underlay/vlan networks dynamically;
7. Set MTU of Pod interface to provider interface's MTU on each node;
8. Add new fields in CRD `Vlan`.


pkg/daemon/controller.go
```diff
@@ -127,6 +147,298 @@ func NewController(config *Configuration, podInformerFactory informers.SharedInf
...
+
+	if err != nil {
+		utilruntime.HandleError(err)
+		c.deleteProviderNetworkQueue.AddRateLimited(obj)
+		return true
+	}
+	return true
+}
+
+func (c *Controller) handleAddOrUpdateProviderNetwork(key string) error {
```
oilbeater, line 247: separate create and clean logical into different functions

pkg/daemon/init.go
```diff
@@ -52,71 +54,67 @@ func InitNodeGateway(config *Configuration) error {
...
+		return nil
+	}
+
+	// get host nic
+	if output, err = ovs.Exec("list-ports", brName); err != nil {
+		return fmt.Errorf("failed to list ports of OVS birdge %s, %v: %q", brName, err, output)
+	}
+	var nic string
+	if output != "" {
+		nic = strings.Split(output, "\n")[0]
```
oilbeater, line 100: why the first nic

pkg/daemon/ovs.go
```diff
@@ -471,14 +473,26 @@ func configureNic(link, ip string, macAddr net.HardwareAddr, mtu int) error {
...
 	}
-	output, err = ovs.Exec(ovs.IfExists, "get", "open", ".", "external-ids:ovn-bridge-mappings")
-	if err != nil {
-		return fmt.Errorf("failed to get external-ids, %v", err)
+	if output, err = ovs.Exec("list-ports", bridge); err != nil {
+		return fmt.Errorf("failed to list ports of OVS birdge %s, %v: %q", bridge, err, output)
+	}
+	if output != "" {
+		for _, port := range strings.Split(output, "\n") {
+			if port != nic {
```
oilbeater, line 486: why remove other ports here

oilbeater (changes requested):

pkg/daemon/ovs.go
zhangzujian: 1. If an OVS bridge needs multiple ports, maybe the ports should be bonded into a bond interface or use OVS bonding;
2. Currently add events and update events of provider networks are handled in the same function(s), so other ports are considered as old bridged ones and should be removed from the bridge.

oilbeater: If I remember correctly, ovn will create patch port on provider bridge

zhangzujian: You are right. I ignored patch ports. What do you think about removing all other ports excluding the patch ones?

oilbeater (approved):

PR #907, opened 2021-07-05
label lr, ls and lsp, and add label filter when gc

lut777:
Add label when "lr-add", "ls-add" and "lsp-add", and list labeled network  entities when gc to avoid deleting entities which does not belong to Kube-ovn.


lut777:
Loadbalancer, portgroup, StaticRoute and VpcNatGateway will soon be added in processing.

PR #910, opened 2021-07-06
update networkpolicy process

hongzhen-ma:
#### What type of this PR
- Bug fixes
update networkpolicy process, add port-group base node, and update acl for policy limit exactly

### DIFF
--- a/pkg/controller/node.go
+++ b/pkg/controller/node.go
@@ -273,6 +273,13 @@ func (c *Controller) handleAddNode(key string) error {
 		return err
 	}
 
+	// ovn acl doesn't support address_set name with '-', so replace '-' by '.'
+	pgName := strings.Replace(node.Annotations[util.PortNameAnnotation], "-", ".", -1)
+	if err := c.ovnClient.CreatePortGroup(pgName, "node", key); err != nil {
+		klog.Errorf("failed to create port group %v for node %s, %v", portName, key, err)
+		return err
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


pkg/controller/node.go
```diff
@@ -509,3 +523,130 @@ func (c *Controller) checkChassisDupl(node *v1.Node) error {
...
+	pgPorts, err := c.ovnClient.ListPgPorts(pgName)
+	if err != nil {
+		klog.Errorf("failed to fetch ports for pg %v, %v", pgName, err)
+		return changed, err
+	}
+
+	pordIds := make([]string, 0, len(ports))
+	for _, port := range ports {
+		portId, err := c.ovnClient.ConvertLspNameToUuid(port)
+		if err != nil {
```
zhangzujian, line 558: We should log error.

pkg/controller/node.go
```diff
@@ -509,3 +523,130 @@ func (c *Controller) checkChassisDupl(node *v1.Node) error {
...
+			continue
+		}
+		pordIds = append(pordIds, portId)
+	}
+
+	for _, portId := range pordIds {
+		if !util.IsStringIn(portId, pgPorts) {
+			klog.Infof("new added pod %v should add to node port group %v", portId, pgName)
+			changed = true
```
zhangzujian, line 567: We can return with `true, nil`.

pkg/controller/node.go
```diff
@@ -509,3 +523,130 @@ func (c *Controller) checkChassisDupl(node *v1.Node) error {
...
+			break
+		}
+	}
+
+	for _, pgPort := range pgPorts {
+		if !util.IsStringIn(pgPort, pordIds) {
+			klog.Infof("can not find match pod for pgport %v", pgPort)
+			changed = true
```
zhangzujian, line 575: Same with above.

zhangzujian (changes requested):

zhangzujian (approved):

PR #920, opened 2021-07-13
add node internal ip into ovn-ic advertise blacklist

lut777:
Add nodes' internal ips into blacklist in ovn-ic. 
Therefore data streams from pod to another node will not be routed to ovn-ic.


pkg/controller/ovn-ic.go
```diff
@@ -146,6 +147,13 @@ func (c *Controller) establishInterConnection(config map[string]string) error {
 			blackList = append(blackList, subnet.Spec.CIDRBlock)
 		}
 	}
+	nodes, err := c.nodesLister.List(labels.Everything())
```
oilbeater, line 150: node and subnet may changed, blacklist set should be outside the `establishInterConnection`

oilbeater (changes requested):

oilbeater (approved):

PR #927, opened 2021-07-15
<fix> update qos ingress_policing_burst

feixiang43:
Ovs document suggest set ingress_policing_burst  at least as large as 80% of  ingress_policing_rate, this helps TCP come closer to achieving the full rate.


oilbeater:
Thanks! @feixiang43

PR #932, opened 2021-07-18
[WIP] Security group

fanriming:
#### What type of this PR
feature
- API changes

TODO list:
- [x] Define the CRD
- [x] ACL command for security group rules
- [x] Security group lifecycle management and rule synchronization
- [x] Associate pod's logical port with a security group
- [x] Update install scripts and yaml

# Refer to the link
Implementing Security Groups using OVN Port Groups
http://dani.foroselectronica.es/implementing-security-groups-in-openstack-using-ovn-port-groups-478/
https://www.ovirt.org/develop/release-management/features/network/networking-api-security-groups.html

# How does kube-ovn implement security groups
- By default, a port_group for deny all is created and the following ACL rules are configured.
```
ovn-nbctl acl-add ovn.sg.kubeovn_deny_all to-lport 2003 'outport==@ovn.sg.kubeovn_deny_all ip' drop
ovn-nbctl acl-add ovn.sg.kubeovn_deny_all from-lport 2003 'inport==@ovn.sg.kubeovn_deny_all ip' drop
```
- All ports that use security groups will be added to the deny_all port_group first.
- Each user-defined security group corresponds to one port_group instance.
```
ovn-nbctl pg-add ovn.sg.sg1 -- set port_group sg1 external_ids:type=security_group external_ids:sg=sg1

# If AllowSameGroupTraffic is enabled, add the following ACL rules
ovn-nbctl create address_set ovn.sg.sg1.associated.v4 external_ids:sg=sg1 external_ids:sg_associated=true
ovn-nbctl create address_set ovn.sg.sg1.associated.v6 external_ids:sg=sg1 external_ids:sg_associated=true
ovn-nbctl acl-add ovn.sg.sg1 to-lport 2004 'outport==@ovn.sg.sg1 && ip4 && ip4.src==$ovn.sg.sg1.associated.v4' allow-related
ovn-nbctl acl-add ovn.sg.sg1 to-lport 2004 'outport==@ovn.sg.sg1 && ip6 && ip6.src==$ovn.sg.sg1.associated.v6' allow-related
ovn-nbctl acl-add ovn.sg.sg1 from-lport 2004 'inport==@ovn.sg.sg1 && ip4 && ip4.dst==$ovn.sg.sg1.associated.v4' allow-related
ovn-nbctl acl-add ovn.sg.sg1 from-lport 2004 'inport==@ovn.sg.sg1 && ip6 && ip6.dst==$ovn.sg.sg1.associated.v6' allow-related
```
- According to the definition of security group rules on Openstack and commercial public cloud, design CRD and implement the transformation of rules to OVN ACL in controller.
- Use the POD's annotation to correlate to the security group.

# Usage
## 1. Creating a Security Group
```
apiVersion: kubeovn.io/v1
kind: SecurityGroup
metadata:
  name: sg1
spec:
  allowSameGroupTraffic: true
  egressRules:
    - ipVersion: ipv4 
      policy: allow                           
      priority: 1
      protocol: all
      remoteAddress: 0.0.0.0/0
      remoteType: address
  ingressRules:
    - ipVersion: ipv4
      policy: allow
      priority: 10
      protocol: icmp
      remoteAddress: 0.0.0.0/0
      remoteType: address
    - ipVersion: ipv4
      policy: allow
      priority: 10
      protocol: tcp
      portRangeMin: 3306
      portRangeMax: 3306
      remoteAddress: 192.168.2.100
      remoteType: address
    - ipVersion: ipv4
      policy: allow
      priority: 10
      protocol: all
      remoteSecurityGroup: sg2
      remoteType: securityGroup
```
### Parameters
`allowSameGroupTraffic`: Whether to allow communication with the security group,  'true' for allows, 'false' for not.

`ipVersion`: Should be `ipv4` or `ipv6`.

`policy`:  Should be `allow` or `drop`.

`priority`: The value ranges from 1 to 200. The smaller the value, the higher the priority.

`protocol`: `tcp/udp/icmp/all` is supported. If `tcp` or `udp` is specified, the port range must be specified

`portRangeMin` & `portRangeMax`: The port range is 1 to 65535, PortRangeMin and portRangeMax can be set to the same value when you want to configure policies for a single port.

`remoteType`: Shoule be `address` or `securityGroup`.

`remoteSecurityGroup`: When `remoteType` is `securityGroup`, specify the securityGroup name associated with the policy.

## 2. Associate pod nic with  security groups
Only ports with security ports enabled can use security groups, so we need to add the annotation `ovn.kubernetes.io/port_security: 'true' `, then use the `ovn.kubernetes.io/security_groups` statement to correlate the security group.
```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    ovn.kubernetes.io/port_security: 'true'
    ovn.kubernetes.io/security_groups: 'sg1,sg3,sgx'
  namespace: default
  name: scrutiry-pod1
---
apiVersion: v1
kind: Pod
metadata:
  annotations:
    ovn.kubernetes.io/port_security: 'true'
    ovn.kubernetes.io/security_groups: 'sg1,sg3,sgx'
  namespace: default
  name: scrutiry-pod2
```
When you want to modify the pod associated security group strategy, we can dynamically update `ovn.Kubernetes.io/security_groups `, without having to rebuild pod, a new correlation policy will take effect immediately.

# Known limitations and usage recommendations
## Gateway address checking
If the egress direction is allowed but the Ingress direction is not allowed, the POD cannot ping the gateway address. This does not affect the routing function of OVN, but by default, CNI needs to ping the gateway to determine whether the network is correctly allocated, which leads to the failure of pod creation. There are two temporary solutions to this problem:
1. On the ingress direction rule, allow gateway access.
2. See the PR https://github.com/kubeovn/kube-ovn/pull/941, set subnet configures disableGatewayCheck parameters to false.
## Scope
Although the security group scope is not limited, it is not recommended to use the same security group instance on different VPC, especially if allowSameGroupTraffic is enabled. Because different VPC may have overlapping IP addresses, unpredictable errors may occur. You are advised to create independent security group instances in different VPC.


oilbeater (approved):

pkg/controller/security_group.go
```diff
@@ -0,0 +1,443 @@
...
+
+	sg, err := c.sgsLister.Get(key)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return nil
+		}
+		return err
+	}
+
+	// check sg rules
```
hongzhen-ma, line 223: the validation check operation maybe can be abstracted to a single function

hongzhen-ma (approved):

pkg/controller/security_group.go
fanriming: The optimization has been completed as recommended. Thanks.

fanriming: > the validation check operation maybe can be abstracted to a single function

The optimization has been completed as recommended. Thanks.

PR #963, opened 2021-08-12
if the string of ip  is empty,program will die

wangyd1988:
if the string of ip  is empty,program will die,so we need check the length of ip.


pkg/ovs/ovn-nbctl.go
```diff
@@ -583,7 +583,10 @@ func (c Client) createRouterPort(ls, lr, ip, mac string) error {
 		klog.Errorf("failed to create switch router port %s %v", lsTolr, err)
 		return err
 	}
-
+    if len(ip) == 0{
```
oilbeater, line 586: need space before brace
`== 0{` -> `== 0 {`

oilbeater (changes requested):

oilbeater:
Thanks! @wangyd1988

PR #966, opened 2021-08-13
fix uninstall.sh

zhangzujian:
Delete iptables rules and ipsets introduced in recent commits.


dist/images/cleanup.sh
```diff
@@ -29,7 +29,9 @@ do
   status=`echo ${ovsstatus#*+}`
   if [  "$status" = "Running"  ]; then
     ovs=`echo ${ovsstatus%+*}`
-    kubectl exec -n kube-system "$ovs" -- bash /kube-ovn/uninstall.sh
+    nodeName=`kubectl get pod -n kube-system $ovs -o jsonpath="{..spec.nodeName}"`
```
oilbeater, line 32: podIP or hostIP should be equal to nodeIP

oilbeater (changes requested):

oilbeater (approved):

PR #989, opened 2021-08-24
update eip arp cmd

hongzhen-ma:
#### What type of this PR
- Bug fixes
####
when arp cmd receives more response msg than request msg, the ret val of cmd is 1. This situation is correct and need go on process.


dist/images/vpcnatgateway/nat-gateway.sh
```diff
@@ -12,6 +12,28 @@ function exec_cmd() {
...
+# Unicast reply from 10.140.28.1 [00:00:5E:00:01:01]  3.026ms
+# Sent 3 probes (1 broadcast(s))
+# Received 6 response(s)
+# cmd "arping -c 3 -s 10.140.28.210 10.140.28.1", ret "1"
+
+function exec_eip_cmd() {
+    cmd=${@:1:${#}}
+    $cmd
+    ret=$?
+    if [ $ret -lt 0 ]; then
```
oilbeater, line 31: The shell command can not return negative value

PR #1005, opened 2021-09-03
fix external-vpc

fanriming:
#### What type of this PR
- Bug fixes


hongzhen-ma:
The binary file in dist/images/ should not be commited and merged.

fanriming:
> The binary file in dist/images/ should not be commited and merged.

fanriming:
> The binary file in dist/images/ should not be commited and merged.

The binary file has been deleted.

PR #1021, opened 2021-09-10
feature: vpc support policy route

luoyunhe:
vpc support policy route.

Policy-based routing (PBR) provides a mechanism to configure permit/deny and reroute policies on the router.
for more: https://mail.openvswitch.org/pipermail/ovs-dev/2018-November/353816.html


yamls/crd.yaml
```diff
@@ -407,6 +407,19 @@ spec:
                         type: string
                     type: object
                   type: array
+                policyRoutes:
+                  items:
```
oilbeater, line 411: `install.sh` and `install-pre-1.16.sh` also need be updated

oilbeater (changes requested):

yamls/crd.yaml
luoyunhe: ok, I will fix latest, and would you give me some suggestions about this featue, thanks.

oilbeater: It looks good to me, nice work!

oilbeater (approved):

oilbeater:
Thanks! @luoyunhe

PR #1022, opened 2021-09-13
fix gc lsp statistic and ipam init for multiple subnet

lut777:
in ipam, add ipam address access for attachNetworks.
in gc, ignore lsp compare for subnet which is not provided by ovn.

### DIFF
--- a/pkg/controller/gc.go
+++ b/pkg/controller/gc.go
@@ -213,6 +213,13 @@ func (c *Controller) markAndCleanLSP() error {
 				continue
 			}
 			providerName := strings.ReplaceAll(k, util.AllocatedAnnotationSuffix, "")
+			isProviderovn, err := c.isOVNProvided(providerName, pod)
+			if err != nil {
+				klog.Errorf("determine if provider is ovn failed %v", err)
+			}
+			if !isProviderovn {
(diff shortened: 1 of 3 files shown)


pkg/controller/gc.go
```diff
@@ -525,3 +532,16 @@ func (c *Controller) gcStaticRoute() error {
 	}
 	return nil
 }
+
+func (c *Controller) isOVNProvided(providerName string, pod *corev1.Pod) (bool, error) {
+	ls := pod.Annotations[fmt.Sprintf(util.LogicalSwitchAnnotationTemplate, providerName)]
+	subnet, err := c.config.KubeOvnClient.KubeovnV1().Subnets().Get(context.Background(), ls, metav1.GetOptions{})
+	if err != nil {
+		klog.Errorf("parse annotation logical switch %s error", ls)
```
zhangzujian, line 540: We should log error here to indicate what happended.

zhangzujian (changes requested):

pkg/controller/gc.go
lut777: right. error info added

zhangzujian (approved):

PR #1031, opened 2021-09-15
feat: support VLAN subnet in VPC

fanriming:
#### What type of this PR
- API changes

1. In a custom VPC, both overlay and underlay subnets can be routed.
2. Set 'ovn-chassis-mac-mappings' for provider to avoid mac moves on Top of the Rack (TOR) switch. Reference link : https://www.openvswitch.org/support/ovscon2018/6/1440-sharma.pdf

### DIFF
--- a/pkg/controller/subnet.go
+++ b/pkg/controller/subnet.go
@@ -569,20 +569,21 @@ func (c *Controller) handleAddOrUpdateSubnet(key string) error {
 		return err
 	}
 
+	needRouter := subnet.Spec.Vlan == "" || subnet.Spec.Vpc != util.DefaultVpc
 	if !exist {
 		subnet.Status.EnsureStandardConditions()
 		// If multiple namespace use same ls name, only first one will success
-		if err := c.ovnClient.CreateLogicalSwitch(subnet.Name, vpc.Status.Router, subnet.Spec.Protocol, subnet.Spec.CIDRBlock, subnet.Spec.Gateway, subnet.Spec.Vlan != ""); err != nil {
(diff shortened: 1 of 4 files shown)


zhangzujian:
Nice work!

Only one suggestion - We can add a field in subnet spec to control the feature, so that users can use "pure" underlay network in custom VPCs and we can provide compability.

fanriming:
> Nice work!
> 
> Only one suggestion - We can add a field in subnet spec to control the feature, so that users can use "pure" underlay network in custom VPCs and we can provide compability.

@zhangzujian Thank you for your suggestion.
In my opinion, a pure underlay network is irrelevant to a VPC because there is no logical router for it. If we want to use a "pure" underly network, simply leave the "VPC" field blank.

zhangzujian:
How does the `ovn-chassis-mac-mappings` configuration take effect? I tried to figure it out by capturing packets between Pods and between Pod and OVN router, but no significant changes were observed.

fanriming:
> How does the `ovn-chassis-mac-mappings` configuration take effect? I tried to figure it out by capturing packets between Pods and between Pod and OVN router, but no significant changes were observed.

Pod does not sense this change, but the physical switch does.
If 'ovn-chassis-mac-mappings' is not configured, the MAC address table on the physical switch frequently changes:
image

Continuous mac move of logical router port macs.
- Logical router ports are distributed across hypervisors.
- Packets with logical router port MAC address as source MAC will come from multiple chassis.
- Will cause mac moves on Top of the Rack (TOR) switch.
   ■ As a security measure packets could be dropped.
   ■ As a security measure corresponding TOR switch port could be blocked.

https://www.openvswitch.org/support/ovscon2018/6/1440-sharma.pdf

zhangzujian (approved):
LGTM, Thanks!

PR #1038, opened 2021-09-18
fix IPAM for StatefulSet

zhangzujian:
#### What type of this PR

- Bug fixes

Consider the following scenario:

1. A StatefulSet Pod with random IP address allocation is deleted and soon kube-ovn-contrller is restarted;
2. kube-ovn-controller initailizes IPAM on startup and no IPAM record for the deleted StatefuleSet Pod is created;
3. A new StatefulSet Pod with the same name is created and kube-ovn-controller assigns a new IP address to it.


oilbeater:
If StatefulSet Pod IP not initialized in IPAM, this IP might be acquired by another Pod, before the StatefulSet Pod acquires this IP

oilbeater (approved):

PR #1092, opened 2021-10-28
support using logical gateway in underlay subnet

zhangzujian:
#### What type of this PR

- Feature


docs/subnet.md
```diff
@@ -66,6 +66,7 @@ Since kube-ovn v1.8.0, kube-ovn support using designative egress ip on node, the
 ## Advance Options
 
 - `vlan`: if enable vlan network, use this field to specific which vlan the subnet should bind to.
+- `logicalGateway`: Create a logical gateway for the default subnet instead of using underlay gateway. Take effect only when the default subnet is in underlay mode. Default: `false`.
```
oilbeater, line 69: what about a vlan subnet that is not the default one with this option

zhangzujian: It's expected to work for all underlay subnets. Fixed in the new commit.

oilbeater (approved):

PR #1093, opened 2021-10-28
feat: pod can use multiple nic with the same subnet

fanriming:
- API changes
 
Pod can be assigned multiple nics of the same subnet, for example:
```
...
annotations
  ovn.kubernetes.io/logical_switch: ovn-default
  attachnet.default.ovn.kubernetes.io/logical_switch: ovn-default
...
```

### DIFF
--- a/pkg/ipam/subnet.go
+++ b/pkg/ipam/subnet.go
@@ -19,16 +19,17 @@ type Subnet struct {
 	V4FreeIPList     IPRangeList
 	V4ReleasedIPList IPRangeList
 	V4ReservedIPList IPRangeList
-	V4PodToIP        map[string]IP
+	V4NicToIP        map[string]IP
 	V4IPToPod        map[IP]string
 	V6CIDR           *net.IPNet
 	V6FreeIPList     IPRangeList
 	V6ReleasedIPList IPRangeList
 	V6ReservedIPList IPRangeList
-	V6PodToIP        map[string]IP
+	V6NicToIP        map[string]IP
 	V6IPToPod        map[IP]string
-	PodToMac         map[string]string
+	NicToMac         map[string]string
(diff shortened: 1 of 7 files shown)


pkg/ipam/subnet.go
```diff
@@ -103,83 +106,98 @@ func NewSubnet(name, cidrStr string, excludeIps []string) (*Subnet, error) {
...
+
+func (subnet *Subnet) popPodNic(podName, nicName string) {
+	subnet.PodToNicList[podName] = util.RemoveString(subnet.PodToNicList[podName], nicName)
+}
+
+func (subnet *Subnet) GetRandomAddress(podName, nicName string, skippedAddrs []string) (IP, IP, string, error) {
 	subnet.mutex.Lock()
-	defer subnet.mutex.Unlock()
+	defer func() {
+		subnet.pushPodNic(podName, nicName)
```
oilbeater, line 164: If error is not nil, is it necessary to still push the nic?

fanriming:
```
func (subnet *Subnet) pushPodNic(podName, nicName string) {
	if subnet.V4NicToIP[nicName] != "" || subnet.V6NicToIP[nicName] != "" || subnet.NicToMac[nicName] != "" {
		subnet.PodToNicList[podName] = util.UniqString(append(subnet.PodToNicList[podName], nicName))
	}
}
```
This method will check whether the nic info needs to be recorded, so I think it's ok.

oilbeater:
@fanriming Can you add an example in the `multi-nic.md`

fanriming:
> @fanriming Can you add an example in the `multi-nic.md`
@oilbeater Examples have been added to 'multi-nic.md'

oilbeater (approved):

PR #1102, opened 2021-11-01
Use logical router policy for accessing node

zhangzujian:
Use logical router policy instead of static route for accessing node.

### DIFF
--- a/pkg/controller/subnet.go
+++ b/pkg/controller/subnet.go
@@ -592,6 +592,11 @@ func (c *Controller) handleAddOrUpdateSubnet(key string) error {
 		}
 	}
 
+	if err = c.updateNodeAddressSetsForSubnet(subnet, false); err != nil {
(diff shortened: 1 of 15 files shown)


pkg/controller/subnet.go
```diff
@@ -695,6 +722,34 @@ func (c *Controller) handleDeleteLogicalSwitch(key string) (err error) {
...
+			if util.CIDRContainIP(v6CIDR, nodeIPv6) {
+				asName := nodeUnderlayAddressSetName(node.Name, 6)
+				if err = c.ovnClient.RemoveAddressSetAddresses(asName, v6CIDR); err != nil {
+					klog.Errorf("failed to remove CIDR %s from address set %s: %v", v6CIDR, asName, err)
+					return err
+				}
+			}
+		}
+	}
+
```
oilbeater, line 752: can this part be moved to a separated function

oilbeater (approved):

PR #1104, opened 2021-11-02
fix LB in dual stack cluster

zhangzujian:
#### What type of this PR

- Bug fixes

### DIFF
--- a/pkg/controller/endpoint.go
+++ b/pkg/controller/endpoint.go
@@ -13,7 +13,6 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
 
-	kubeovnv1 "github.com/kubeovn/kube-ovn/pkg/apis/kubeovn/v1"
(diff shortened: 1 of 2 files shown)


pkg/controller/endpoint.go
```diff
@@ -235,14 +235,35 @@ func getServicePortBackends(endpoints *v1.Endpoints, servicePort v1.ServicePort,
...
+					for _, podIP := range podIPs {
+						if util.CheckProtocol(podIP.IP) == protocol {
+							ip = podIP.IP
+							break
+						}
+					}
+					break
 				}
 			}
+			if ip == "" {
```
oilbeater, line 259: why not using addresses.IP directly?

zhangzujian: In dual stack cluster, endpoints address contains IPv4 only.

oilbeater (approved):

PR #1118, opened 2021-11-06
other CNI can be used as the default network

fanriming:
- Optimize

When other CNI network is attached to POD via annotations `v1.multus-cni.io/default-network`, and use kubeovn's IPAM feature, is now allowed.

### DIFF
--- a/cmd/cni/cni.go
+++ b/cmd/cni/cni.go
@@ -42,7 +42,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


oilbeater (approved):

infinitydon:
Does this mean that the following is no longer a limitation?

_**Kube-OVN now uses ovn network as the pod default network, other network can only act as network attachments. We will fully separate the IPAM functions to provide a more general IPAM later.**_

https://github.com/kubeovn/kube-ovn/blob/master/docs/multi-nic.md#limitation

PR #1119, opened 2021-11-09
add ovn-ic e2e

lut777:
Add e2e for ovn-ic as following procedure:
1, add two kind cluster as "kube-ovn" and "kube-ovn1"
2, install kube-OVN on both clusters.
3, install ovn-ic configs and Ic container.
4, check if logical switch "ts" and the corresponding logical switch ports are right on ovn-nb-db.
5, clean enviroment.

Since two clusters in kind ask for more RAM than 1 cluster, it's preferable that the host has 8G RAM. At least in my tests, this is what is needed for stable operation.

### DIFF
--- a/.github/workflows/build-x86-image.yaml
+++ b/.github/workflows/build-x86-image.yaml
@@ -886,6 +886,56 @@ jobs:
           sh -c 'while :; do if [ $(kubectl get --no-headers subnet | wc -l) -eq 2 ]; then break; fi; sleep 5; done'
           sh dist/images/cleanup.sh
 
+  ovnic-e2e:
+    needs: build
+    name: 1-master-ovnic-e2e
+    runs-on: ubuntu-18.04
+    timeout-minutes: 30
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Install Kind
+        run: |
+          curl -Lo ./kind https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-$(uname)-amd64
+          chmod +x ./kind
+          sudo mv kind /usr/local/bin
+
+      - name: Init Kind
+        run: |
+          pip install j2cli --user
+          pip install "j2cli[yaml]" --user
+          sudo PATH=~/.local/bin:$PATH make kind-init-cluster
+
+      - name: Download image
+        uses: actions/download-artifact@v2
+        with:
+          name: kube-ovn
+
+      - name: Install Kube-OVN
+        run: |
+          docker load --input kube-ovn.tar
+          sudo PATH=~/.local/bin:$PATH make kind-install-cluster
+          sudo PATH=~/.local/bin:$PATH make kind-install-ic
+
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.16
+        id: go
+
+      - name: Run E2E
+        run: |
+          go install github.com/onsi/ginkgo/ginkgo@latest
+          sudo kubectl cluster-info
+          sudo chmod -R 777 /home/runner/.kube/
+          make e2e-ovn-ic
+
+      - name: Cleanup
+        run: |
+          sh -c 'while :; do if [ $(kubectl get --no-headers subnet | wc -l) -eq 2 ]; then break; fi; sleep 5; done'
+          sh dist/images/cleanup.sh
+
   push:
     needs:
       - single-e2e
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -215,6 +239,18 @@ kind-install-underlay-logical-gateway-dual:
...
+.PHONY: kind-install-ic
+kind-install-ic:
+	$(eval HOSTIP = $(shell ip -4 route get 8.8.8.8 | awk {'print $$7'} | tr -d '\n'))
+	zone=az0 host_node_ip=${HOSTIP} gateway_node_name=kube-ovn-control-plane j2 yamls/ovn-ic.yaml.j2 -o /ovn-ic-0.yaml
+	zone=az1 host_node_ip=${HOSTIP} gateway_node_name=kube-ovn1-control-plane j2 yamls/ovn-ic.yaml.j2 -o /ovn-ic-1.yaml
+	kubectl config use-context kind-kube-ovn
+	kubectl apply -f /ovn-ic-0.yaml
+	kubectl config use-context kind-kube-ovn1
+	kubectl apply -f /ovn-ic-1.yaml
+	docker run --name=ovn-ic-db -d --network=host -v /etc/ovn/:/etc/ovn -v /var/run/ovn:/var/run/ovn -v /var/log/ovn:/var/log/ovn $(REGISTRY)/kube-ovn:$(RELEASE_TAG) bash start-ic-db.sh
```
zhangzujian, line 251: 1. How about moving the commands to `kind-install-cluster` so that we can get an ovn-ic environment using only one command?
2. `ip` command is not available in macos so it won't work in macos. How about creating the ic-db container with network `kind` and pass the container's IP to ovn-ic template?
3. We should remove the ic-db container in `kind-clean`.

Makefile
```diff
@@ -85,6 +85,18 @@ kind-init:
 	kind create cluster --config yamls/kind.yaml --name kube-ovn
 	kubectl describe no
 
+.PHONY: kind-init-cluster
+kind-init-cluster:
+	kind delete cluster --name=kube-ovn
+	kind delete cluster --name=kube-ovn1
```
zhangzujian, line 91: We should remove the ic-db container here and remove cluster `kube-ovn1` in `kind-clean`.

Makefile
```diff
@@ -129,6 +141,18 @@ kind-install:
...
+.PHONY: kind-install-cluster
+kind-install-cluster:
+	kind load docker-image --name kube-ovn $(REGISTRY)/kube-ovn:$(RELEASE_TAG)
+	kind load docker-image --name kube-ovn1 $(REGISTRY)/kube-ovn:$(RELEASE_TAG)
+	kubectl config use-context kind-kube-ovn
+	ENABLE_SSL=true dist/images/install.sh
+	kubectl describe no
+	kubectl config use-context kind-kube-ovn1
+	sed -e 's/10.16.0/10.18.0/g' -e 's/10.96.0/10.98.0/g' -e 's/100.64.0/100.68.0/g'  dist/images/install.sh > /install-multi.sh
+	ENABLE_SSL=true bash /install-multi.sh
```
zhangzujian, line 153: 1. Place `install-multi.sh` in current directory to avoid permission issues.
2. Add `install-multi.sh` to `.gitignore`.

.github/workflows/build-x86-image.yaml
```diff
@@ -886,6 +886,58 @@ jobs:
...
+
+      - name: Download image
+        uses: actions/download-artifact@v2
+        with:
+          name: kube-ovn
+
+      - name: Install Kube-OVN
+        run: |
+          pip install j2cli --user
+          pip install "j2cli[yaml]" --user
```
zhangzujian, line 918: Duplicate installation. `j2cli` has been installed in `Init Kind`.

zhangzujian (changes requested):

.github/workflows/build-x86-image.yaml
lut777: removed

Makefile
lut777: 1, I still prefer to separate the commands based on functionality...
2, u a right. this problem will be solved in ovn-ic db HA test.
3, a kind-clean-cluster is added now.

Makefile
lut777: added.

Makefile
lut777: added.

zhangzujian (approved):

PR #1157, opened 2021-12-08
chore: update klog to v2 which embed log rotation

oilbeater:


go.mod
```diff
@@ -51,6 +51,7 @@ require (
 	k8s.io/apimachinery v0.20.4
 	k8s.io/client-go v0.20.4
 	k8s.io/klog v1.0.0
```
hongzhen-ma, line 53: Is klog v1.0.0 should be deleted?

hongzhen-ma (approved):

PR #1161, opened 2021-12-09
update delete operation for statefulset pod

hongzhen-ma:
#### What type of this PR
- Bug fixes
####
1、statefulset pod is running
2、change subnet for ns which statefulset pod belongs to
3、delete statefulset pod, and it can not rerun successfully

### DIFF
--- a/pkg/controller/pod.go
+++ b/pkg/controller/pod.go
@@ -146,6 +146,11 @@ func (c *Controller) enqueueDeletePod(obj interface{}) {
 			klog.V(3).Infof("enqueue delete pod %s", key)
 			c.deletePodQueue.Add(obj)
 		}
+
+		if delete, err := appendCheckStatefulSetPodToDel(c, p); delete && err == nil {
+			klog.V(3).Infof("enqueue delete pod %s", key)
+			c.deletePodQueue.Add(obj)
+		}
 	} else {
 		klog.V(3).Infof("enqueue delete pod %s", key)
 		c.deletePodQueue.Add(obj)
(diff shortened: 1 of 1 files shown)


pkg/controller/pod.go
```diff
@@ -1210,3 +1215,21 @@ func (c *Controller) deleteAttachmentNetWorkIP(pod *v1.Pod) error {
 	}
 	return nil
 }
+
+func appendCheckStatefulSetPodToDel(c *Controller, pod *v1.Pod) bool {
+	// subnet for ns has been changed, and statefulset pod's ip is not in the range of subnet's cidr anymore
+	podNs, _ := c.namespacesLister.Get(pod.Namespace)
```
oilbeater, line 1221: should check error herre

pkg/controller/pod.go
```diff
@@ -1210,3 +1215,21 @@ func (c *Controller) deleteAttachmentNetWorkIP(pod *v1.Pod) error {
...
+	if podNs.Annotations[util.LogicalSwitchAnnotation] != "" && pod.Annotations[util.LogicalSwitchAnnotation] != "" && strings.TrimSpace(podNs.Annotations[util.LogicalSwitchAnnotation]) != strings.TrimSpace(pod.Annotations[util.LogicalSwitchAnnotation]) {
+		klog.Infof("ns %s annotation subnet is %s, which is inconstant with subnet for pod %s, delete pod", pod.Namespace, podNs.Annotations[util.LogicalSwitchAnnotation], pod.Name)
+		return true
+	}
+
+	// subnet cidr has been changed, and statefulset pod's ip is not in the range of subnet's cidr anymore
+	podSubnet, _ := c.subnetsLister.Get(strings.TrimSpace(pod.Annotations[util.LogicalSwitchAnnotation]))
```
oilbeater, line 1228: same as above

oilbeater (changes requested):

pkg/controller/pod.go
hongzhen-ma: modified

pkg/controller/pod.go
hongzhen-ma: modified

oilbeater (approved):

PR #1165, opened 2021-12-13
fix: add back the leader check

oilbeater:
Previous patch doesn't change the readiness prob interval which may leader two probs run in the same period and always steal the lock from ovn-northd. Increase the interval and add back the leader check.

#### What type of this PR
- API changes
- Bug fixes

### DIFF
--- a/dist/images/ovn-is-leader.sh
+++ b/dist/images/ovn-is-leader.sh
@@ -42,6 +42,22 @@ fi
 if [[ $sb_leader =~ "true" ]]
 then
    kubectl label --overwrite pod "$POD_NAME" -n "$POD_NAMESPACE" ovn-sb-leader=true
+   set +e
+   northd_svc=$(kubectl get svc -n kube-system | grep ovn-northd)
+   if [ -z "$northd_svc" ]; then
+    echo "ovn-northd svc not exist"
+   else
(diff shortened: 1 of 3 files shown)


dist/images/ovn-is-leader.sh
```diff
@@ -35,6 +35,22 @@ fi
...
+  northd_svc=$(kubectl get svc -n kube-system | grep ovn-northd)
+  if [ -z "$northd_svc" ]; then
+   echo "ovn-northd svc not exist"
+  else
+   northd_leader=$(kubectl get ep -n kube-system ovn-northd -o jsonpath={.subsets\[0\].addresses\[0\].ip})
+   if [ "$northd_leader" == "" ]; then
+      # no available northd leader try to release the lock
+      if [[ "$ENABLE_SSL" == "false" ]]; then
+        ovsdb-client -v -t 1 steal tcp:127.0.0.1:6642  ovn_northd
+      else
```
hongzhen-ma, line 48: It seems that the location is not right. The `ENABLE_SSL` is false in line 36.

oilbeater: move the location

hongzhen-ma (approved):

PR #1180, opened 2021-12-16
Do not send multicast packets to conntrack when stateful (allow-related) ACLs are configured

zhangzujian:
#### What type of this PR

- Performance

As discussed in ovn-org/ovn#108, we should also skip conntrack for multicast packets when stateful (allow-related) ACLs are configured.


oilbeater:
We can hold the pr and wait the commit finally merge to upstream and just cherry pick from upstream

PR #1184, opened 2021-12-20
add cilium e2e

lut777:
As title.
In e2e, kube-proxy is removed and cilium is introduced in but the svc dnat is still available.

### DIFF
--- a/.github/workflows/build-x86-image.yaml
+++ b/.github/workflows/build-x86-image.yaml
@@ -936,6 +936,64 @@ jobs:
           sh -c 'while :; do if [ $(kubectl get --no-headers subnet | wc -l) -eq 2 ]; then break; fi; sleep 5; done'
           sh dist/images/cleanup.sh
 
+  cilium-e2e:
+    needs: build
+    name: 1-master-cilium-e2e
+    runs-on: ubuntu-20.04
+    timeout-minutes: 30
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Install Kind
+        run: |
+          curl -Lo ./kind https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-$(uname)-amd64
+          chmod +x ./kind
+          sudo mv kind /usr/local/bin
+
+      - name: Init Kind
+        run: |
+          sudo pip3 install j2cli
+          sudo pip3 install "j2cli[yaml]"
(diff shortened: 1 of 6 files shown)


.github/workflows/build-x86-image.yaml
```diff
@@ -936,6 +936,60 @@ jobs:
...
+        with:
+          go-version: ^1.17
+        id: go
+
+      - name: Run E2E
+        run: |
+          go install github.com/onsi/ginkgo/ginkgo@latest
+          sudo kubectl cluster-info
+          sudo chmod -R 777 /home/runner/.kube/
+          make e2e-ovn-ebpf
```
oilbeater, line 986: Should run the basic e2e too

oilbeater (changes requested):

.github/workflows/build-x86-image.yaml
lut777: can't. basic e2e would check the status of kube-proxy which is already deleted in this scenario.

.github/workflows/build-x86-image.yaml
```diff
@@ -936,6 +936,64 @@ jobs:
...
+        id: go
+
+      - name: Run E2E
+        run: |
+          go install github.com/onsi/ginkgo/ginkgo@latest
+          sudo kubectl cluster-info
+          sudo cp -r /root/.kube/ /home/runner/.kube/
+          sudo chmod -R 777 /home/runner/.kube/
+          docker exec kube-ovn-control-plane bash -c "ls -al /etc/cni/net.d/"
+          sleep 30
```
oilbeater, line 989: the sleep can be removed

oilbeater (changes requested):

.github/workflows/build-x86-image.yaml
lut777: kubectl  rollout status -n kube-system ds cilium
Above command is added and replaced the "sleep" to ensure cilium pods are properly started.

test/e2e-ebpf/e2e_ebpf_suite_test.go
```diff
@@ -0,0 +1,80 @@
+package e2e_ebpf
+
+import (
+	"context"
+	_ "embed"
```
zhangzujian, line 5: Package `embed` shoud be removed from the import list.

lut777: deleted

oilbeater (approved):

PR #1185, opened 2021-12-20
compatible with OVN 20.06

fanriming:
OVN 20.06 does not support --may-exist/--if-exists options for policy routing, we expect policy-base routing work on older versions of OVN(20.06), so submit this PR.

Refer to the OVN upstream code 
https://github.com/ovn-org/ovn/commit/da0a96008616b81682aabe32eeba29974459ea4d


pkg/ovs/ovn-nbctl.go.bak
```diff
@@ -0,0 +1,1918 @@
+package ovs
```
oilbeater, line 1: this file should be removed

oilbeater (changes requested):

pkg/ovs/ovn-nbctl.go.bak
fanriming: done

PR #1166, opened 2021-12-13
perf: do not send traffic to ct if not designate to svc

oilbeater:
#### What type of this PR
Examples of user facing changes:
- Performance

PR #1202, opened 2021-12-27
add svc cidr in ovs LB for optimization

lut777:
As an automated optimisation patch for #1166 
kubeovn >= 1.9


oilbeater:
The commits need to be squashed

PR #1213, opened 2021-12-30
Optimized decision logic

wangyd1988:
#### What type of this PR
- API changes
- Bug fixes

Fixes #(issue-number)

### DIFF
--- a/pkg/controller/pod.go
+++ b/pkg/controller/pod.go
@@ -570,10 +570,7 @@ func (c *Controller) handleDeletePod(pod *v1.Pod) error {
 	c.podKeyMutex.Lock(key)
 	defer c.podKeyMutex.Unlock(key)
 
-	p, err := c.podsLister.Pods(pod.Namespace).Get(pod.Name)
-	if err != nil {
-		return err
-	}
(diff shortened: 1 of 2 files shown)


zhangzujian (approved):

wangyd1988:
?

pkg/controller/pod.go
```diff
@@ -570,7 +570,10 @@ func (c *Controller) handleDeletePod(pod *v1.Pod) error {
 	c.podKeyMutex.Lock(key)
 	defer c.podKeyMutex.Unlock(key)
 
-	p, _ := c.podsLister.Pods(pod.Namespace).Get(pod.Name)
+	p, err := c.podsLister.Pods(pod.Namespace).Get(pod.Name)
+	if err != nil {
+		return err
+	}
```
zhangzujian, line 576: This causes the e2e testing failure. We should ignore error or check whether the error is `Pod not found` since the controller is handling Pod deletion event.

wangyd1988: ok, ignore error

zhangzujian:
Please squash commits into one.

Issue #1151, opened 2021-12-08
Use go to rerimplement ovn-is-leader.sh


oilbeater:
[ovn-is-leader.sh](https://github.com/kubeovn/kube-ovn/blob/master/dist/images/ovn-is-leader.sh) is not convenient to be called by readiness prob, difficult to maintain and relies on kubectl binary.

We should move it to a go binary to periodically maintain the ovn leader and pod labels relations and other ovn db maintain works can grow in it later.

PR #1243, opened 2022-01-10
1 fix issue of  https://github.com/kubeovn/kube-ovn/issues/1151

sxilinux:
What type of this PR

    Bug fixes
  
   I tried to reimplement the func of ovn-leader checking. But I do not know if it is as your expectation.
   would you give me  some comments?

Which issue(s) this PR fixes:
https://github.com/kubeovn/kube-ovn/issues/1151
Fixes #(1151)

### DIFF
--- a/cmd/ovn_leader_checker/ovn_leader_checker.go
+++ b/cmd/ovn_leader_checker/ovn_leader_checker.go
@@ -1,0 +1,18 @@
+package ovn_leader_checker
+
+import (
+	"github.com/kubeovn/kube-ovn/pkg/ovn_leader_checker"
+	"k8s.io/klog/v2"
+)
+
+func CmdMain() {
+	cfg, err := ovn_leader_checker.ParseFlags()
+	if err != nil {
+		klog.Errorf("ovn_leader_checker parseFlags error %v", err)
+	}
+	err = ovn_leader_checker.KubeClientInit(cfg)
+	if err != nil {
+		klog.Errorf("KubeClientInit err %v", err)
+	}
+	ovn_leader_checker.StartOvnLeaderCheck(cfg)
+}
--- a/dist/images/start-db.sh
+++ b/dist/images/start-db.sh
@@ -269,4 +269,5 @@ ovs-appctl -t /var/run/ovn/ovnnb_db.ctl ovsdb-server/memory-trim-on-compaction o
 ovs-appctl -t /var/run/ovn/ovnsb_db.ctl ovsdb-server/memory-trim-on-compaction on
 
 chmod 600 /etc/ovn/*
-tail -f /var/log/ovn/ovn-northd.log
+/kube-ovn/kube-ovn-leader-checker
+
--- a/cmd/cmdmain.go
+++ b/cmd/cmdmain.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/kubeovn/kube-ovn/cmd/controller"
 	"github.com/kubeovn/kube-ovn/cmd/daemon"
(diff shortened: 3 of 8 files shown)


oilbeater:
@sxilinux the reason that we want to move it to go binary is that the `ovn-is-leader.sh` now is doing lots more than just readinessProbe. The leader detection and label pods work should be run in the main cmd rather than a readinessProbe.

What we need is to move this part of work to a new cmd and simplify the readinessProbe.

sxilinux:
@oilbeater , thanks for your reply.
 accodring to your description, my understanding is like that:  
1 replace old new ovn_is _checker.sh by using go cmd( for convenience, we call the go binary as ovn_leader, ovn_leader does the same job as  ovn_is _checker.sh ). and at pod readiness process, insteadlly, we use  ovn_leader.    

am i right?

oilbeater:
@sxilinux `ovn-leader` 需要完成 leader 判断和切换以及 compact 相关的周期性任务，readiness 只需要完成检查 nothd，ovnnb 和 ovnsb 状态的任务即可，相当于把真正的健康检查工作和周期性的处理工作分开。周期性的工作放在 go binary 里，让然后容器启动后最后就执行这个二进制文件即可。健康检查可以还用之前的脚本，但只保留最基础的检查操作，或者将这部分工作也转移到 go 代码里，提供 http 接口进行检查

sxilinux:
@oilbeater ，需求我理解了。我重新把代码调整下。

sxilinux:
@oilbeater  i have remodified the code like what you said. would you review the codes?

sxilinux:
1 i constructed kind-env in my local machine , did the test and  fixed code miss which will cause ci failture.

cmd/ovn_leader_checker/ovn_leader_checker.go
```diff
@@ -0,0 +1,25 @@
+package ovn_leader_checker
+
+import (
+	"github.com/kubeovn/kube-ovn/pkg/ovn_leader_checker"
+	"os"
+)
+
+func CmdMain() {
+	cfg, err := ovn_leader_checker.ParseFlags()
+	if err != nil {
```
oilbeater, line 10: It should run periodically no, not once.

We don't need to exit on error, log the error message is enough.

dist/images/start-db.sh
```diff
@@ -74,6 +74,12 @@ function is_clustered {
   return 1
 }
 
+#### first  try to boot ovn-leader-checker
+#### we put kube-ovn-leader-checker at the same place of ovn-cental
+#### seperate pod of is also ok
+
+/kube-ovn/kube-ovn-leader-checker &
```
oilbeater, line 81: it should replace the last tail command and always run

oilbeater (changes requested):

cmd/ovn_leader_checker/ovn_leader_checker.go
sxilinux: i modified codes just like you said.

dist/images/start-db.sh
sxilinux: i modified codes just like you said.

oilbeater:
Please follow https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git and squash the 8 commits into one

sxilinux:
ok. i will squash these commits into one

sxilinux:
Hi, [16477a8](https://github.com/kubeovn/kube-ovn/pull/1243/commits/16477a804aeb4351f30d946e27b65ecd8531cb96) has squashed commits into one.

oilbeater:
Thanks! @sxilinux

PR #1245, opened 2022-01-10
support to add multiple subnets for a namespace

### LINKED ISSUE #1224: Namespace with multiple subnets
Previously, one namespace can only bind to one subnet which leads to scale difficulty.

We need to bind multiple subnets to one namespace and automatically to use the next subnet if the previous one is full

hongzhen-ma:
#### What type of this PR
- API changes
####
1、support multus subnet annotation for a namespace
2、the namespace should be added to subnet by modify subnet.Spec.Namespaces
3、the namespace annotation should not be modified manually

Fixes #1224

### DIFF
--- a/pkg/controller/namespace.go
+++ b/pkg/controller/namespace.go
@@ -71,6 +71,10 @@ func (c *Controller) enqueueUpdateNamespace(old, new interface{}) {
 		klog.Warningf("no logical switch annotation for ns %s", newNs.Name)
 		c.addNamespaceQueue.Add(newNs.Name)
 	}
+
+	if newNs.Annotations != nil && newNs.Annotations[util.LogicalSwitchAnnotation] != "" && !reflect.DeepEqual(oldNs.Annotations, newNs.Annotations) {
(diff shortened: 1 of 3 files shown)


pkg/controller/namespace.go
```diff
@@ -71,6 +72,20 @@ func (c *Controller) enqueueUpdateNamespace(old, new interface{}) {
 		klog.Warningf("no logical switch annotation for ns %s", newNs.Name)
 		c.addNamespaceQueue.Add(newNs.Name)
 	}
+
+	if newNs.Annotations != nil && newNs.Annotations[util.LogicalSwitchAnnotation] != "" && !reflect.DeepEqual(oldNs.Annotations, newNs.Annotations) {
+		var subnet *kubeovnv1.Subnet
```
oilbeater, line 77: use `c.addNamespaceQueue.Add(newNs.Name)` can handle the update

oilbeater (changes requested):

oilbeater (approved):

PR #1247, opened 2022-01-10
Add args to configure port ln-ovn-external

### LINKED ISSUE #1237: port ln-ovn-external could be configurable
### Feature request

properties of port ln-ovn-external could be configurable according to the physical network, such as:
- vlanid,  for the case that underlay network has vlanid configured
- network_name of port options, for the case that connects ln-ovn-external to existing ovn bridge instead of creating br-external

chestack:
Fixes #1237


oilbeater (approved):

oilbeater:
Nice work! Thanks @chestack

PR #1267, opened 2022-01-19
we should first see if a condition is not going to be met

wangyd1988:
#### What type of this PR
- API changes
- Bug fixes

Fixes #(issue-number)

### DIFF
--- a/pkg/controller/subnet.go
+++ b/pkg/controller/subnet.go
@@ -1367,6 +1367,9 @@ func (c *Controller) deleteEcmpRouteForNode(subnet *kubeovnv1.Subnet) error {
 			continue
 		}
 
+		if subnet.Status.ActivateGateway != "" && subnet.Status.ActivateGateway == gw {
+			continue
+		}
 		ipStr := node.Annotations[util.IpAddressAnnotation]
 		for _, ip := range strings.Split(ipStr, ",") {
 			var cidrBlock string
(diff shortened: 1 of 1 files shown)


pkg/controller/subnet.go
```diff
@@ -1356,6 +1356,10 @@ func (c *Controller) checkGwNodeExists(gatewayNode string) bool {
 
 func (c *Controller) deleteEcmpRouteForNode(subnet *kubeovnv1.Subnet) error {
 	for _, gw := range strings.Split(subnet.Spec.GatewayNode, ",") {
+		if subnet.Status.ActivateGateway != "" && subnet.Status.ActivateGateway == gw {
+			continue
+		}
+
```
hongzhen-ma, line 1362: https://github.com/kubeovn/kube-ovn/wiki/%E9%9B%86%E4%B8%AD%E5%BC%8F%E5%AD%90%E7%BD%91%E7%BD%91%E5%85%B3NAT%E6%A8%A1%E5%BC%8F%E6%8C%87%E5%AE%9A%E5%87%BA%E7%BD%91IP

In this case, gw is not equal to subnet.Status.ActivateGateway. 
Check condition not to be met is a good suggestion, and the check can be moved after gw is valued.

hongzhen-ma (changes requested):

pkg/controller/subnet.go
wangyd1988: ok

hongzhen-ma (approved):

PR #1280, opened 2022-01-27
add policy route for distributed subnet in default vpc

### LINKED ISSUE #1256: Use policy route to replace origin src route
Previously we need one src route entry for each pod that use distribute gateway. The route number will blow up when cluster grows.  We can use the policy route to reduce the route number, also we don't need to maintain a custom ovn with modification to src route priority

hongzhen-ma:
#### What type of this PR
- API changes
####
1. Add policy-route for distributed subnet in default vpc
2. Add port-group based on combination of subnet and node, add pod's ports to port-group on pod's node
3. Add overlay subnet address-set to avoid pod traffic when pod visit pod of overlay subnet

Fixes #1256

### DIFF
--- a/pkg/controller/gc.go
+++ b/pkg/controller/gc.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
 
+	kubeovnv1 "github.com/kubeovn/kube-ovn/pkg/apis/kubeovn/v1"
 	"github.com/kubeovn/kube-ovn/pkg/ovs"
 	"github.com/kubeovn/kube-ovn/pkg/util"
 )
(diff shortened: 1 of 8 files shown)


pkg/controller/gc.go
```diff
@@ -496,6 +497,21 @@ func (c *Controller) gcPortGroup() error {
...
 		}
+
+		// append overlay subnets port group to npNames to avoid gc distributed subnets port group
+		subnets, err := c.subnetsLister.List(labels.Everything())
+		if err != nil {
+			klog.Errorf("failed to list subnets %v", err)
+			return err
+		}
+		for _, subnet := range subnets {
+			if subnet.Spec.Vpc != util.DefaultVpc || subnet.Spec.Vlan != "" || subnet.Spec.LogicalGateway || subnet.Name == c.config.NodeSwitch || subnet.Spec.GatewayType != kubeovnv1.GWDistributedType {
```
zhangzujian, line 508: Field `LogicalGateway` is meaningless for overlay subnets now, so we can remove this check.

hongzhen-ma: modified

zhangzujian:
We need to handle the obsolete `src-ip` routes in update scenarios.

zhangzujian (approved):

oilbeater (approved):

PR #1284, opened 2022-02-07
feat: update provider network via node annotation

zhangzujian:
#### What type of this PR

- Feature

Control node behavior in provider networks via node annotations. `PROVIDER.provider-network.kubernetes.io/exclude` is used to exclude the node and `PROVIDER.provider-network.kubernetes.io/interface` is used to specify the interface.

The annotations have less priority than the `spec` field in a provider network and will be removed soon after they are recognized.

### DIFF
--- a/pkg/controller/init.go
+++ b/pkg/controller/init.go
@@ -392,6 +392,12 @@ func (c *Controller) initDefaultProviderNetwork() error {
 		return err
 	}
 
+	nodes, err := c.nodesLister.List(labels.Everything())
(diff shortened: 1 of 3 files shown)


oilbeater (approved):

oilbeater:
Should we also delete related provider network configure when node is deleted?

zhangzujian:
> Should we also delete related provider network configure when node is deleted?

Added, please review.

oilbeater (approved):

PR #1295, opened 2022-02-15
append add cidr and excludeIps annotation for namespace

### LINKED ISSUE #1224: Namespace with multiple subnets

hongzhen-ma:
#### What type of this PR
- Bug fixes

Fixes #1224


zhangzujian (approved):

hongzhen-ma:
已回合 release1.8 和  release1.9

PR #1298, opened 2022-02-15
fix: dnat can't take effect when dnat eIP is equal to floatingip eIP,

xujunjie-cover:
What type of this PR
- Bug fixes

whether dnatrules chain（SHARED_DNAT） should have priority.

### DIFF
--- a/dist/images/vpcnatgateway/nat-gateway.sh
+++ b/dist/images/vpcnatgateway/nat-gateway.sh
@@ -26,8 +26,8 @@ function init() {
     iptables -t nat -N SHARED_SNAT
 
     iptables -t nat -A PREROUTING -j DNAT_FILTER
-    iptables -t nat -A DNAT_FILTER -j EXCLUSIVE_DNAT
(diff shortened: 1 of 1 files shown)


oilbeater:
@fanriming can you help to review this pr?

fanriming:
@xujunjie-cover 
Configuring floatingIP means that the EIP is instance-exclusive. That is, the EIP is configured with both DNAT and SNAT full-port mapping rules to the back-end instance. 
The EIP used as floatingIP cannot be used in single-port DNAT rules. Therefore, the EXCLUSIVE_DNAT priority is higher than that of SHARED_DNAT.

fanriming (commented):
Configuring floatingIP means that the EIP is instance exclusive. That is, the EIP and the back-end instance are configured with DNAT and SNAT full-port mapping rules.
The EIP used as the floating IP address cannot be used for single-port DNAT rules. This is why the EXCLUSIVE_DNAT priority is higher than SHARED_DNAT. So I don't think we need to change it here.

PR #1316, opened 2022-02-23
when update gateway info, we should append old to new deploy

wangyd1988:
#### What type of this PR
   
meet user  defined annotation requirements

### DIFF
--- a/pkg/controller/vpc_nat_gateway.go
+++ b/pkg/controller/vpc_nat_gateway.go
@@ -246,7 +246,7 @@ func (c *Controller) handleAddOrUpdateVpcNatGw(key string) error {
 
 	// check or create deployment
 	needToCreate := false
-	_, err = c.config.KubeClient.AppsV1().Deployments(c.config.PodNamespace).
+	oldDp, err := c.config.KubeClient.AppsV1().Deployments(c.config.PodNamespace).
 		Get(context.Background(), genNatGwDpName(gw.Name), metav1.GetOptions{})
 
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/vpc_nat_gateway.go
```diff
@@ -246,7 +246,7 @@ func (c *Controller) handleAddOrUpdateVpcNatGw(key string) error {
 
 	// check or create deployment
 	needToCreate := false
-	_, err = c.config.KubeClient.AppsV1().Deployments(c.config.PodNamespace).
+	oldDp, err := c.config.KubeClient.AppsV1().Deployments(c.config.PodNamespace).
```
oilbeater, line 249: We should make a deepcopy of the oldDp, other wise the `newPodAnnotations[key] = value` operations below will affect the cached value.

oilbeater (changes requested):

pkg/controller/vpc_nat_gateway.go
wangyd1988: it's only get  old  cached value from oldDp, how  affect?

oilbeater: https://go.dev/play/p/63SXnSpqSNp

wangyd1988: > https://go.dev/play/p/63SXnSpqSNp

good ，i see

wangyd1988: But I find  that this client doesn't have caching
image
image

oilbeater: you're right, i didn't notice it

oilbeater (approved):

PR #1340, opened 2022-03-01
add check for pod update process

hongzhen-ma:
#### What type of this PR
- Bug fixes
####
1、The pod is added but not updated in some case, which will lead the failure of handleadd in daemon.

Fixes #(issue-number)


oilbeater (approved):

hongzhen-ma:
已回合 release1.8 和  release1.9

PR #1353, opened 2022-03-02
adding IP Protocol enumeration to CRD can reduce the kube-ovn Controller judgment logic  

wangyd1988:
#### What type of this PR
- API changes
- Bug fixes

Fixes #(issue-number)

### DIFF
--- a/pkg/controller/subnet.go
+++ b/pkg/controller/subnet.go
@@ -96,7 +96,8 @@ func (c *Controller) enqueueUpdateSubnet(old, new interface{}) {
 		oldSubnet.Spec.DHCPv4Options != newSubnet.Spec.DHCPv4Options ||
 		oldSubnet.Spec.DHCPv6Options != newSubnet.Spec.DHCPv6Options ||
 		oldSubnet.Spec.EnableIPv6RA != newSubnet.Spec.EnableIPv6RA ||
-		oldSubnet.Spec.IPv6RAConfigs != newSubnet.Spec.IPv6RAConfigs {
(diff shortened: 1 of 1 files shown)


pkg/controller/subnet.go
```diff
@@ -289,7 +289,7 @@ func formatSubnet(subnet *kubeovnv1.Subnet, c *Controller) error {
 		subnet.Spec.Provider = util.OvnProvider
 		changed = true
 	}
-	if subnet.Spec.Protocol == "" || subnet.Spec.Protocol != util.CheckProtocol(subnet.Spec.CIDRBlock) {
```
hongzhen-ma, line 292: The protocol field can be filled automatically if it is not set. The subnet crd should be created as easy as posible.

wangyd1988: if  i  create or update a error  ip  protocol(like  aaa )， subnet  crd help me changeing it to correct，i think it's not good idea. 
  If ip protocol filed is only be to demonstration  when someone  passes the  command  "kubectl get subnet "?  
I think it should be in the status  zone.

hongzhen-ma: Are there some situations in your environment to set the protocol field(correct or wrong)，and is used specially ?

wangyd1988: yes,  We support creating subnets using YAML. ApiServer  need to check whether the protocol fields are correct.

hongzhen-ma:
欢迎来参加kube-ovn的双周会，或者在微信群里讨论下，你那环境上对subnet使用的一些建议

oilbeater:
It LGTM, @wangyd1988 please rebase the master and fix the conflict

PR #1354, opened 2022-03-02
don't check conflict for migration pod with only static mac

chestack:
#### What type of this PR
- Bug fixes

It returned conflict error if kubevirt VM pod with only static mac(no static IP configured) during migration because that mac address is conflicted with the existing pod.

Add parameter to func GetRandomAddress() to skip check conflict for migration pod 

relates to #1253 #1001


oilbeater (approved):

oilbeater:
Thanks! @chestack

PR #1359, opened 2022-03-04
optimize kube-ovn-monitor and add db restore process

### LINKED ISSUE #1299: Automatically fix db in "inconsistent data" status
### Feature request

We meet errors that ovn-sb storage status is `inconsistent data` and ovn-northd stop translating flow from ovn-nb to ovn-sb, ovn-controller cannot claim portbingding to ovn-sb, then all flow updates are stopped.

It seems that some transactions violate the database constraints, however we are not aware about how it happens and can only manually rebuild the ovn-sb database. We need a new way to automatically fix the db status.

hongzhen-ma:
#### What type of this PR
- API changes

####
optimize the implementation of kube-ovn-monitor

Fixes #1299

### DIFF
--- a/dist/images/install.sh
+++ b/dist/images/install.sh
@@ -993,6 +993,7 @@ rules:
       - statefulsets
       - daemonsets
       - deployments
+      - deployments/scale
     verbs:
       - create
       - delete
@@ -1466,6 +1467,7 @@ rules:
       - statefulsets
       - daemonsets
       - deployments
+      - deployments/scale
     verbs:
       - create
       - delete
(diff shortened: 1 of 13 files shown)


dist/images/install.sh
```diff
@@ -2935,6 +2937,52 @@ dbtool(){
         dbstatus)
           kubectl exec "$OVN_NB_POD" -n $KUBE_OVN_NS -c ovn-central -- ovn-appctl -t /var/run/ovn/ovnnb_db.ctl ovsdb-server/get-db-storage-status OVN_Northbound
           ;;
+        restore)
+          # set ovn-central replicas to 0
+          replicas=$(kubectl get deployment -n $KUBE_OVN_NS ovn-central -o jsonpath={.spec.replicas})
+          kubectl scale deployment -n $KUBE_OVN_NS ovn-central --replicas=0
```
zhangzujian, line 2943: Should we wait for the operation to be completed?

dist/images/install.sh
```diff
@@ -2935,6 +2937,52 @@ dbtool(){
...
+          podNames=`kubectl get pod -n $KUBE_OVN_NS | grep ovs-ovn | awk '{print $1}'`
+          echo "first nodeIP is $firstIP"
+
+          i=0
+          for nodeIp in $nodeIps
+          do
+            for pod in $podNames
+            do
+              hostip=$(kubectl get pod -n $KUBE_OVN_NS $pod -o jsonpath={.status.hostIP})
+              if [ nodeIp==hostip ]; then
```
zhangzujian, line 2960: Missing `$` here.

dist/images/install.sh
hongzhen-ma: We need to move db file to /tmp in ovs-ovn pod. The ovs-ovn pod will be crashed if we wait enough time when I did test. That will result in the failure of the command.So don't wait here.

dist/images/install.sh
hongzhen-ma: modified

zhangzujian (approved):

zhangzujian:
Please change the title.

hongzhen-ma:
已回合1.8和1.9分支

PR #1371, opened 2022-03-10
add detail error when failed to create resource

wangchenglong01:
#### What type of this PR
- Code Style

/flake

### Which problem is this PR solving?
- add detail error when failed to create resource

### DIFF
--- a/pkg/controller/init.go
+++ b/pkg/controller/init.go
@@ -185,7 +185,11 @@ func (c *Controller) initNodeSwitch() error {
 	}
 
 	_, err = c.config.KubeOvnClient.KubeovnV1().Subnets().Create(context.Background(), &nodeSubnet, metav1.CreateOptions{})
-	return err
+	if err != nil {
+		klog.Errorf("failed to create subnet %s: %v", c.config.NodeSwitch, err)
+		return err
+	}
+	return nil
 }
 
 // InitClusterRouter init cluster router to connect different logical switches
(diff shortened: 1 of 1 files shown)


pkg/controller/init.go
```diff
@@ -185,6 +185,10 @@ func (c *Controller) initNodeSwitch() error {
 	}
 
 	_, err = c.config.KubeOvnClient.KubeovnV1().Subnets().Create(context.Background(), &nodeSubnet, metav1.CreateOptions{})
+	if err != nil {
+		klog.Errorf("failed to create subnet %s: %v", c.config.NodeSwitch, err)
+		return err
+	}
```
zhangzujian, line 191: No need to return here. Or `return nil` in the following line.

pkg/controller/init.go
```diff
@@ -491,6 +499,10 @@ func (c *Controller) initDefaultVlan() error {
 	}
 
 	_, err = c.config.KubeOvnClient.KubeovnV1().Vlans().Create(context.Background(), &defaultVlan, metav1.CreateOptions{})
+	if err != nil {
+		klog.Errorf("failed to create vlan %s: %v", defaultVlan, err)
+		return err
+	}
```
zhangzujian, line 505: Same.

pkg/controller/init.go
```diff
@@ -456,6 +460,10 @@ func (c *Controller) initDefaultProviderNetwork() error {
 	}()
 
 	_, err = c.config.KubeOvnClient.KubeovnV1().ProviderNetworks().Create(context.Background(), &pn, metav1.CreateOptions{})
+	if err != nil {
+		klog.Errorf("failed to create provider network %s: %v", c.config.DefaultProviderName, err)
+		return err
+	}
```
zhangzujian, line 466: Same.

pkg/controller/init.go
wangchenglong01: @zhangzujian You're right. I have changed them and the latest code has been commited.

zhangzujian (approved):

PR #1390, opened 2022-03-21
append vm deletion check

hongzhen-ma:
#### What type of this PR
- Bug fixes

####
append check for vm delete. The vm pod should be deleted when subnet for ns has been changed.

Fixes #(issue-number)

### DIFF
--- a/pkg/controller/pod.go
+++ b/pkg/controller/pod.go
@@ -152,12 +152,16 @@ func (c *Controller) enqueueDeletePod(obj interface{}) {
 			c.deletePodQueue.Add(obj)
 		}
 
-		if delete, err := appendCheckStatefulSetPodToDel(c, p); delete && err == nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/pod.go
```diff
@@ -1446,6 +1446,30 @@ func (c *Controller) isVmPodToDel(pod *v1.Pod, vmName string) bool {
 		return true
 	}
 
+	// subnet for ns has been changed, and vm pod's ip is not in the range of subnet's cidr anymore
```
oilbeater, line 1449: this should only happens when the pod is to be deleted

oilbeater: statefulset pod should also consider this

oilbeater (changes requested):

oilbeater (approved):

hongzhen-ma:
已回合release-1.8和release-1.9分支

PR #1391, opened 2022-03-21
adding IP Protocol enumeration to CRD can reduce the kube-ovn Controller judgment logic

wangyd1988:
…ler judgment logic

#### What type of this PR
- API changes
- Bug fixes

Fixes #(issue-number)

### DIFF
--- a/yamls/crd.yaml
+++ b/yamls/crd.yaml
@@ -168,6 +168,10 @@ spec:
                   type: boolean
                 protocol:
                   type: string
+                  enum:
+                    - IPv4
+                    - IPv6
+                    - Dual
                 cidrBlock:
                   type: string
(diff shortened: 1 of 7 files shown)


yamls/crd.yaml
```diff
@@ -168,6 +168,10 @@ spec:
                   type: boolean
                 protocol:
                   type: string
```
oilbeater, line 170: `install.sh` should also be changed

oilbeater (changes requested):

yamls/crd.yaml
wangyd1988: ok

oilbeater (approved):

PR #1395, opened 2022-03-22
add custom acls for subnet

### LINKED ISSUE #1383: ACL in subnet
### Feature request

Kube-OVN previously supports `private` and `allowSubnets` to control access. However, users need fine grained access control for different direction, protocols match rules and actions. We need to support full function set of ACL in Kube-OVN subnet.

hongzhen-ma:
#### What type of this PR
- API changes

####
1、add acls in subnet, which can be used to implement customized acl

Fixes #1383


pkg/ovs/ovn-nbctl.go
```diff
@@ -2483,3 +2483,64 @@ func (c *Client) UpdateRouterPortIPv6RA(ls, lr, cidrBlock, gateway, ipv6RAConfig
...
+
+func (c Client) UpdateSubnetACL(ls string, acls []kubeovnv1.Acl) error {
+	if len(acls) == 0 {
+		return nil
+	}
+
+	if err := c.DeleteSubnetACL(ls); err != nil {
+		klog.Errorf("failed to delete acls for subnet %s, %v", ls, err)
+		return err
+	}
```
zhangzujian, line 2517: Should we delete ACLs first so that the old ACLs can be removed when no ACL exists in the new spec?

hongzhen-ma: good suggestion. Modified.

dist/images/install.sh
```diff
@@ -643,6 +643,30 @@ spec:
...
+                        type: string
+                        enum:
+                          - from-lport
+                          - to-lport
+                      priority:
+                        type: integer
+                        minimum: 0
+                        maximum: 32767
+                      match:
+                        type: string
```
zhangzujian, line 661: If `match` must not be empty, we can add it to `required`. Example: https://github.com/kubeovn/kube-ovn/blob/master/yamls/crd.yaml#L278.

zhangzujian (approved):

dist/images/install.sh
hongzhen-ma: This is the simplest implementation for acls. Maybe we can build the match expression automatically in the future, in which case the match field can be nil.

Issue #1223, opened 2022-01-03
Windows support


oilbeater:
- [x] Investigate the system requirements
- [x] Build and run ovs/ovn windows binaries
- [x] Build and run CNI windows binaries
- [x] How to use docs
- [x] Investigate way to implement distributed gateway
- [x] Investigate way to support underlay network

PR #1397, opened 2022-03-24
update cni version to 1.0

### LINKED ISSUE #1229: Upgrade CNI to v1.0.1
CNI has a new spec, we should check if we can upgrade to it and if it is compatible with old version kubernetes

zhangzujian:
#### What type of this PR

- Enhancement

Fix #1229. This is also part of Windows support (#1223).


oilbeater:
What's the relation with the `cniVersion` in `01-kube-ovn.conflist`

zhangzujian:
> What's the relation with the `cniVersion` in `01-kube-ovn.conflist`

`cniVersion` in the config tells a CNI plugin which version is being requested. The CNI plugin then generates the corresponding output.

oilbeater (approved):

PR #1401, opened 2022-03-25
[release-1.8] avoid frequent ipset update

zhangzujian:
#### What type of this PR

- Bug Fix

Use `subnets-nat` instead of `local-pod-ip-nat` to avoid frequent ipset update.

CI result: https://github.com/zhangzujian/kube-ovn/actions/runs/2038105832


oilbeater:
need to consider how to upgrade

zhangzujian:
> need to consider how to upgrade

Updated. Now we clear the ipset instead of removing it.

oilbeater:
need to make sure that if set member not changed, the `AddOrReplaceIIPSet` and `ApplyUpdates` function will not actually call ipset to make change

oilbeater (approved):

PR #1423, opened 2022-04-03
feat：add webhook vaildate the vpc resource whether can be deleted.

wlooppy:
#### What type of this PR
* feature
webhook vaildate the vpc resource whether can be deleted:
vpc cannot be deleted when there is any subnet belong to the vpc；because delete the vpc means delete the Logical Router in ovn .


oilbeater (approved):

oilbeater:
Thanks! @wlooppy

PR #1441, opened 2022-04-13
feature: add sb/nb db check bash script

lut777:
Add a bash script for timed tasks to monitor the status of sb/nb db.
If the db status is abnormal, scrip will exit with code 1.
Code in the comment could outputs the full db information.


dist/images/dbcheck_tmp.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
```
oilbeater, line 1: this function should be merged into kubeclt-ko db status command

oilbeater (changes requested):

oilbeater (approved):

PR #1442, opened 2022-04-13
modify webhook img to independent image

### LINKED ISSUE #1424: workqueue and client-go metrics lost for kube-ovn-controller/kube-ovn-cni
# Expected Behavior
Run in kube-ovn-controller pod
```bash
curl 127.0.0.1:10660/metrics | grep workqueue
curl 127.0.0.1:10660/metrics | grep rest_client
```
should show metrics

# Actual Behavior
No metrics

# Additional Info

The reason is that the webhook introduce the `controller-runtime` framework which will init workqueue and rest_client metrics first and register them to a different registry rather than the default registry. As we build all the binaries into one, the metrics registered by kube-ovn-controller and kube-ovn-cni will be skipped. We need to find a way to separate the metrics or build the webhook alone

- kube-ovn version:
  
```
v1.10.0 and v1.9.1
```

hongzhen-ma:
#### What type of this PR
- Bug fixes
####
modify webhook img to independent image since webhook affects kube-ovn-controller/kube-ovn-cni metrics

Fixes #1424

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .DS_Store
 dist/images/kube-ovn-cmd
 dist/images/kube-ovn.exe
+dist/images/kube-ovn-webhook
 test/e2e/ovnnb_db.*
 test/e2e/ovnsb_db.*
 kube-ovn.yaml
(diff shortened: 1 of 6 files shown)


oilbeater (approved):

hongzhen-ma:
已回合release-1.9 分支，release-1.8 分支不需要回合

Issue #1452, opened 2022-04-18
Annotation network_type always is geneve


oilbeater:
<img width="1061" alt="wecom-temp-365bcb18a0b22116da0a2f604ac0772e" src="https://user-images.githubusercontent.com/1189736/163761578-3cf2e350-9b6c-4338-bef1-b47522e7eaf1.png">

When changes the tunnel type to vxlan, the annotation is still geneve.

The annotation should match then encapsulation method like geneve/vxlan/vlan/stt.

PR #1460, opened 2022-04-20
Annotation network_type always is geneve

wangyd1988:
#### What type of this P
annotation network_type is not used in kube-ovn-controller project,So it can be deleted.

Fixes #(1452)

### DIFF
--- a/pkg/controller/pod.go
+++ b/pkg/controller/pod.go
@@ -502,11 +502,6 @@ func (c *Controller) handleAddPod(key string) error {
 			return err
 		}
 		ipStr := util.GetStringIP(v4IP, v6IP)
-		if subnet.Spec.Vlan != "" {
-			pod.Annotations[fmt.Sprintf(util.NetworkTypeTemplate, podNet.ProviderName)] = util.NetworkTypeVlan
-		} else {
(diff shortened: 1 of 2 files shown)


wangyd1988:
"Some checks were not successful " is   not caused by this PR

zhangzujian:
As described in #1452, we should keep the annotation instead of removing it.

wangyd1988:
> As described in #1452, we should keep the annotation instead of removing it.
> 
> Currently, the only value is `geneve` when the Pod is in overlay networking. We should correct it when the encapsulation protocol is `vxlan` or `stt`.

不不，首先这个注解在项目中没有用到，其次双周例会时和刘工讨论的是，如果没有使用到，就删除这个注解

zhangzujian:
> > As described in #1452, we should keep the annotation instead of removing it.
> > Currently, the only value is `geneve` when the Pod is in overlay networking. We should correct it when the encapsulation protocol is `vxlan` or `stt`.
> 
> 不不，首先这个注解在项目中没有用到，其次双周例会时和刘工讨论的是，如果没有使用到，就删除这个注解

Then we should also remove the annotation definition in `pkg/util/const.go`.

zhangzujian (approved):
LGTM

zhangzujian:
Thanks! 👍

PR #1466, opened 2022-04-21
fix routes for packets from Pods to other nodes

zhangzujian:
#### What type of this PR

- Bug fixes

`join` subnet should also be treated as overlay subnets in logical router route policies.

This patch also changes the IPv6 CIDRs used in e2e testing to avoid issue #1467.

### DIFF
--- a/pkg/controller/subnet.go
+++ b/pkg/controller/subnet.go
@@ -1599,7 +1599,7 @@ func (c *Controller) createOverlaySubnetsAddressSet() error {
 	}
 
 	for _, subnet := range subnetList {
-		if subnet.Spec.Vlan != "" || subnet.Spec.Vpc != util.DefaultVpc || subnet.Name == c.config.NodeSwitch {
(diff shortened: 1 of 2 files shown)


pkg/controller/subnet.go
```diff
@@ -1599,7 +1599,7 @@ func (c *Controller) createOverlaySubnetsAddressSet() error {
 	}
 
 	for _, subnet := range subnetList {
-		if subnet.Spec.Vlan != "" || subnet.Spec.Vpc != util.DefaultVpc || subnet.Name == c.config.NodeSwitch {
+		if subnet.Spec.Vlan != "" || subnet.Spec.Vpc != util.DefaultVpc {
```
hongzhen-ma, line 1602: join 子网应该不用再单独建port-group了，使用原来underlay 添加的策略路由就可以

hongzhen-ma (approved):

PR #1473, opened 2022-04-24
delete monitor noexecute toleration

hongzhen-ma:
#### What type of this PR
- Bug fixes
####
The kube-ovn-monitor replicas is set to 1，so the pod can be scheduled between master nodes

Fixes #(issue-number)


zhangzujian (approved):

hongzhen-ma:
已回合release-1.9和release-1.8分支

PR #1490, opened 2022-04-27
delete ipam record and static route when gc lsp

hongzhen-ma:
#### What type of this PR
- API changes
- Bug fixes

Fixes #(issue-number)

### DIFF
--- a/pkg/controller/gc.go
+++ b/pkg/controller/gc.go
@@ -3,6 +3,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"net"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/controller/gc.go
```diff
@@ -258,6 +261,20 @@ func (c *Controller) markAndCleanLSP() error {
...
 				}
+
+				for podName, nsName := range nameNsMap {
+					key := fmt.Sprintf("%s/%s", nsName, podName)
+					c.ipam.ReleaseAddressByPod(key)
+				}
+
+				for _, podAddr := range podAddres {
+					if net.ParseIP(podAddr).To4() != nil || net.ParseIP(podAddr).To16() != nil {
+						if err := c.ovnClient.DeleteStaticRoute(podAddr, c.config.ClusterRouter); err != nil {
```
oilbeater, line 272: should make sure this ip is not used by other running pod

oilbeater (changes requested):

pkg/controller/gc.go
hongzhen-ma: 多个lsp不能同时配置相同的地址吧？
如果IP地址被当前lsp占用的情况下，其他的lsp不应该能配置上相同的IP地址才对

oilbeater: ok

oilbeater (approved):

PR #1511, opened 2022-05-09
optimized initialization and gc for the chassis

fanriming:
#### What type of this PR
- Bug fixes

image

In some environments, the ovs hostname parameter is changed to be different from the node name, which triggers an undesired chassis deletion. And the external chassis should not be removed by kube-ovn. So do the following optimization:

1. Add labels to the chassis managed by kube-ovn to avoid the impact on external chassis.
2. The gc code for redundant chassis is executed in the initialization phase.
3. Chassis 'name' is used instead of 'node name' as the unique index.
4. Optimize the delete command according to the OVN version, compatible with versions earlier than 20.09 .

### DIFF
--- a/pkg/controller/gc.go
+++ b/pkg/controller/gc.go
@@ -23,6 +23,7 @@ var lastNoPodLSP map[string]bool
 func (c *Controller) gc() error {
 	gcFunctions := []func() error{
 		c.gcNode,
+		c.gcChassis,
 		c.gcLogicalSwitch,
 		c.gcCustomLogicalRouter,
 		c.gcLogicalSwitchPort,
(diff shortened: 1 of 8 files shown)


pkg/controller/init.go
```diff
@@ -796,3 +796,30 @@ func (c *Controller) initDeleteOverlayPodsStaticRoutes() error {
 
 	return nil
 }
+
+func (c *Controller) initNodeChassis() error {
```
lut777, line 800: what if the chassis id of a node is deleted, and then this node is rejoined in the cluster?
In this scenario, the func `InitChassisNodeTag` will not be invoked.
Also, the chassisGC  will not re-build the chassis annotation of node.

lut777 (changes requested):

pkg/controller/init.go
fanriming: > what if the chassis id of a node is deleted, and then this node is rejoined in the cluster? In this scenario, the func `InitChassisNodeTag` will not be invoked. Also, the chassisGC will not re-build the chassis annotation of node.

It is only when the OVS DB is deleted and reconstructed that the inconsistency with the node annotation will occur. In this case, I think we should rebuild the OVN-CNI-SERVER pod on the node to refresh the node annotations.
And, the chassis external_id tag will be written the next time the AddNode event comes.

lut777 (approved):

oilbeater:
@fanriming can you resolve the conflict in master? We did some refactor work to replace ovn-nbctl with libovsdb and may lead conflicts

fanriming:
> @fanriming can you resolve the conflict in master? We did some refactor work to replace ovn-nbctl with libovsdb and may lead conflicts

resolved

pkg/controller/gc.go
```diff
@@ -609,6 +610,35 @@ func (c *Controller) gcStaticRoute() error {
...
+	chassises, err := c.ovnClient.GetAllChassis()
+	if err != nil {
+		klog.Errorf("failed to get all chassis, %v", err)
+	}
+	nodes, err := c.nodesLister.List(labels.Everything())
+	if err != nil {
+		klog.Errorf("failed to list nodes, %v", err)
+		return err
+	}
+	for _, chassis := range chassises {
```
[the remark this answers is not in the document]
fanriming: I don't think it's profitable and we still need to iterate over the results of the Chassis to determine if a cleanup is needed.

PR #1522, opened 2022-05-12
remove name for default drop acl in networkpolicy

### LINKED ISSUE #1445: networkpolicy does not work when policy‘s name length is more than 63 characters
# Expected Behavior
create a networkpolicy, with normal ingress or egress acl, the name is more than 63 characters, expect the networkpolicy create successfully and work normally.

# Actual Behavior
the networkpolicy can be created successfully but does not work

1.  create networkpolicy with name length more than 63 characters 
2. test if the networkpolicy works

# Additional Info

- Kubernetes version:

  ```
  (paste your output here)
  ```

- kube-ovn version:
  
  ```
  v1.7.3
  ```

  **Output of `uname -r`:**
  ```
  (paste your output here)
  ```
wecom-temp-470d1633beb67b5107cc39a0266f290f

hongzhen-ma:
#### What type of this PR
- Bug fixes

####
remove name for default drop acl in networkpolicy

Fixes #1445

### DIFF
--- a/pkg/ovs/ovn-nbctl.go
+++ b/pkg/ovs/ovn-nbctl.go
@@ -1567,14 +1567,14 @@ func (c Client) CreateNpAddressSet(asName, npNamespace, npName, direction string
 	return err
 }
 
-func (c Client) CreateIngressACL(npName, pgName, asIngressName, asExceptName, svcAsName, protocol string, npp []netv1.NetworkPolicyPort) error {
+func (c Client) CreateIngressACL(pgName, asIngressName, asExceptName, svcAsName, protocol string, npp []netv1.NetworkPolicyPort) error {
 	var allowArgs []string
 
 	ipSuffix := "ip4"
 	if protocol == kubeovnv1.ProtocolIPv6 {
 		ipSuffix = "ip6"
 	}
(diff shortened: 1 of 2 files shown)


pkg/ovs/ovn-nbctl.go
```diff
@@ -1569,6 +1569,9 @@ func (c Client) CreateNpAddressSet(asName, npNamespace, npName, direction string
 
 func (c Client) CreateIngressACL(npName, pgName, asIngressName, asExceptName, svcAsName, protocol string, npp []netv1.NetworkPolicyPort) error {
 	var allowArgs []string
+	if len(npName) > 63 {
```
oilbeater, line 1572: the  name may conflict with each other, we should consider using hash or uuid

oilbeater (changes requested):

pkg/ovs/ovn-nbctl.go
hongzhen-ma: modified

oilbeater: what about delete and update

hongzhen-ma: The name is only used for default drop acl with priority equals 2000 and the name is never used later.
All the processes about acls later is based on port-group name, including create acls with other priority、update and delete.
Maybe it's better to delete the acl name.

oilbeater (approved):

hongzhen-ma:
已回合release-1.8和release-1.9分支

Issue #1540, opened 2022-05-17
Missing ports in the port group


zhangzujian:
# Expected Behavior

# Actual Behavior

<img width="1360" alt="wecom-temp-2c19a27f54caa1c5cfb650344ee2d833" src="https://user-images.githubusercontent.com/13048299/168708400-2c15b9e6-11a3-47b8-b72c-e5e0d18bbce2.png">

Port `205dda1b-xxx-xxx-xxx` is missing in the port group `ovn.default.win.h9ek9jkm5hp`.

# Steps to Reproduce the Problem

1. Create a deployment YAML, set the replicas to 10, and then specify the node name.
2. Apply the deployment YAML.

# Additional Info

- Kubernetes version:

  **Output of `kubectl version`:**

  ```
  (paste your output here)
  ```

- kube-ovn version:
  
  ```
  (paste your output here)
  ```

- operation-system/kernel version:
  
  **Output of `awk -F '=' '/PRETTY_NAME/ { print $2 }' /etc/os-release`:**
  **Output of `uname -r`:**
  ```
  (paste your output here)
  ```

PR #1544, opened 2022-05-18
replace ovn-nbctl daemon with libovsdb in frequent operations

### LINKED ISSUE #1255: Use libovsdb to replace frequently read operations
Each call to ovn-nbctl daemon will invoke a process which  brings extra cost, and some complex transaction operations are hard to move to libovsdb. We need to move the frequently read options to libovsdb first to reduce the cost

zhangzujian:
#### What type of this PR

- Performance

This patch fixes #1255 and #1540 and is part of #1533.

Improve performance of Pod event processing by replacing ovn-nbctl daemon with libovsdb and by optimizing the Pod process logic.

Seconds took to process restart of a deployment with replicas set to 100 in a cluster with ~10k Pods:

| Scenario | Seconds |
| --- | ---: |
| Before | 180 |
| After | 42 |

#### Method to generate ovn nb package

Get OVN NB schema definition from the OVN source code, e.g. https://github.com/ovn-org/ovn/blob/branch-22.03/ovn-nb.ovsschema. The schema version must match with the OVN version used in `Dockerfile.base`.

Run the following commands:

```sh
 go install github.com/ovn-org/libovsdb/cmd/modelgen
 ~/go/bin/modelgen -p ovnnb -o pkg/ovsdb/ovnnb ovn-nb.ovsschema
```


pkg/controller/subnet.go
```diff
@@ -1100,26 +1106,48 @@ func (c *Controller) reconcileOvnRoute(subnet *kubeovnv1.Subnet) error {
...
+					changed = true
+					klog.Infof("new port %v should be added to port group %s", portsToAdd, pgName)
+					newPgPorts := make([]string, len(portsToAdd), len(portsToAdd)+len(pgPorts))
+					copy(newPgPorts, portsToAdd)
+					for port := range pgPorts {
+						newPgPorts = append(newPgPorts, port)
+					}
+					if err = c.ovnLegacyClient.SetPortsToPortGroup(pgName, newPgPorts); err != nil {
+						klog.Errorf("failed to set ports to port group %v, %v", pgName, err)
+						return err
```
hongzhen-ma, line 1146: return 前需要释放ovnPgKeyMutex

hongzhen-ma (changes requested):

pkg/controller/subnet.go
zhangzujian: fixed

pkg/controller/controller.go
```diff
@@ -327,6 +331,11 @@ func NewController(config *Configuration) *Controller {
 		kubeovnInformerFactory: kubeovnInformerFactory,
 	}
 
+	var err error
+	if controller.ovnClient, err = ovs.NewOvnClient(config.OvnNbAddr, config.OvnTimeout); err != nil {
```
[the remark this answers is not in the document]
zhangzujian:
```go
controller.ovnClient, err := ovs.NewOvnClient()
```
有语法错误：expected identifier on left side of :=

hongzhen-ma (approved):

lut777 (approved):

PR #1550, opened 2022-05-19
fix: opt kubectl-ko install solution

lut777:
#### What type of this PR
- Bug fixes

Now only one kubectl-ko file need to be maintained.


dist/images/install.sh
```diff
@@ -3903,7 +3149,7 @@ if ! sh -c "echo \":$PATH:\" | grep -q \":/usr/local/bin:\""; then
 fi
 
 echo "[Step 6/6] Run network diagnose"
-kubectl ko diagnose all
+docker exec kube-ovn-control-plane kubectl ko diagnose all
```
oilbeater, line 3152: kubectl ko should work in current node.

And kube-ovn-control-plane doesn't exist in a non-kind cluster

oilbeater (changes requested):

oilbeater (approved):

PR #1551, opened 2022-05-20
Always set mac address to sriov vf

### LINKED ISSUE #1519: Always Set MAC address to SRIOV VF
# Expected Behavior
Alway set MAC address to SRIOV VF.

# Actual Behavior
Kubeovn cni plugin set MAC address to SRIOV VF only if the vf goes with a VfioPciDriver, that will not work with virtualization based runtime such as Kata. If we do not set MAC address to VF directly in the host machine, the network driver(e.g. mlx5_core) in the guest os will set a random mac to the vf passed in if the vf mac address is 00:00:00:00:00:00 on the host, or it'll keep the mac address set before by other program. Then the final mac address of the vf on the guest os will not be the same one allocated by kube-ovn.
 

1. Setup K8s with Kata container as runtime.
3. Build from main branch  and Install kube-ovn with HW_OFFLOAD on.
4. Follow the instruction in https://github.com/openshift/ovn-kubernetes/blob/master/docs/ovs_offload.md
5. Create a Kata container pod with sriov dev.
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: ovs-offload-pod1
  annotations:
    v1.multus-cni.io/default-network: default
spec:
  runtimeClassName: kata
  containers:
  - name: appcntr1
    image: centos/tools
    resources:
      requests:
        mellanox.com/cx5_sriov_switchdev: '1'
      limits:
        mellanox.com/cx5_sriov_switchdev: '1'
```
6. the pod will not start, if we do a describe: `kubectl describe pod ovs-offload-pod1`, we can found something like
```
...
update interface: Link not found (Address: 00:00:00:f5:eb:63)
...
```
I can confirm the mac address in the guest os is not the one set by kube-ovn and the vf mac address kept unchanged on the host.

# Additional Info
The code related sits here:
https://github.com/kubeovn/kube-ovn/blob/master/pkg/daemon/ovs_linux.go#L814
```golang
	if isVfioPciDriver {
		// 7. set MAC address to VF
		if err := setVfMac(deviceID, vfIndex, mac); err != nil {
			return "", "", err
		}
	}
```
The vf driver is `mlx5_core`, not a VfioPciDriver. VF mac was set after we remove `if isVfioPciDriver` and rebuild it.

- Kubernetes version:

  ```
  (paste your output here)
  ```

- kube-ovn version:
  
  ```
  (paste your output here)
  ```

  **Output of `uname -r`:**
  ```
  (paste your output here)
  ```

xiaosuiba:
#### What type of this PR
 Bug fixes

Fixes #1519

### DIFF
--- a/pkg/daemon/ovs_linux.go
+++ b/pkg/daemon/ovs_linux.go
@@ -811,12 +811,12 @@ func setupSriovInterface(containerID, deviceID, vfDriver, ifName string, mtu int
 		return "", "", fmt.Errorf("failed to set MTU on %s: %v", hostNicName, err)
 	}
 
-	if isVfioPciDriver {
-		// 7. set MAC address to VF
-		if err := setVfMac(deviceID, vfIndex, mac); err != nil {
-			return "", "", err
-		}
+	
+	// 7. set MAC address to VF
+	if err = setVfMac(deviceID, vfIndex, mac); err != nil {
+		return "", "", err
 	}
+
 	return hostNicName, vfNetdevice, nil
 }
 
--- a/pkg/daemon/ovs_linux.go
+++ b/pkg/daemon/ovs_linux.go
@@ -811,7 +811,6 @@ func setupSriovInterface(containerID, deviceID, vfDriver, ifName string, mtu int
 		return "", "", fmt.Errorf("failed to set MTU on %s: %v", hostNicName, err)
 	}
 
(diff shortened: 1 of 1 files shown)


zhangzujian:
Thanks! Please format the code using `go fmt ./...`.

PR #1558, opened 2022-05-25
do not gc vm pod lsp when vm still exists

### LINKED ISSUE #1502: ipam init and gc issues
wecom-temp-53b8c1548e76614aaee78758e5e46c1d

hongzhen-ma:
#### What type of this PR
- Bug fixes
####
1、do not gc vm pod lsp when vm still exists (keep-vm-ip flag is set in kube-ovn-controller)
2、The same problem exists for sts when sts pods do not use ip_pool to assign static ips

Fixes #1502

### DIFF
--- a/pkg/controller/gc.go
+++ b/pkg/controller/gc.go
@@ -236,6 +236,10 @@ func (c *Controller) markAndCleanLSP() error {
 		}
 	}
 
+	// The lsp for vm pod should not be deleted if vm still exists
+	vmLsps := c.getVmLsps()
+	ipNames = append(ipNames, vmLsps...)
+
 	lsps, err := c.ovnClient.ListLogicalSwitchPort(c.config.EnableExternalVpc)
 	if err != nil {
 		klog.Errorf("failed to list logical switch port, %v", err)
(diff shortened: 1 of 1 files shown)


pkg/controller/gc.go
```diff
@@ -572,3 +576,34 @@ func (c *Controller) isOVNProvided(providerName string, pod *corev1.Pod) (bool,
...
+	}
+
+	nss, err := c.namespacesLister.List(labels.Everything())
+	if err != nil {
+		klog.Errorf("failed to list namespaces, %v", err)
+		return vmLsps
+	}
+
+	for _, ns := range nss {
+		vms, err := c.config.KubevirtClient.VirtualMachine(ns.Name).List(&metav1.ListOptions{})
```
zhangzujian, line 594: Can we replace `ns.Name` with `""` to list resources in all namespaces? 
https://github.com/kubernetes/client-go/issues/159#issuecomment-288624475

pkg/controller/gc.go
```diff
@@ -572,3 +576,34 @@ func (c *Controller) isOVNProvided(providerName string, pod *corev1.Pod) (bool,
...
+		if err != nil {
+			if !k8serrors.IsNotFound(err) {
+				klog.Errorf("failed to list vm in namespace %s, %v", ns, err)
+			}
+			continue
+		} else {
+			for _, vm := range vms.Items {
+				vmLsp := ovs.PodNameToPortName(vm.Name, ns.Name, util.OvnProvider)
+				vmLsps = append(vmLsps, vmLsp)
+			}
```
zhangzujian, line 604: What if the VM has multiple LSP? Is it possible to get all LSPs?

pkg/controller/gc.go
hongzhen-ma: The apis from KubevirtClient can not  set namespace to "", the parameter is requested.
image

pkg/controller/gc.go
hongzhen-ma: 1、Kubevirt can be configured multiple lsps with multi-cni.
2、The pod does not exist, only lsp for vm pod exists. So there is no way to get attachnetworks by pod.
3、It's not worth it to check every lsp and make sure it is used for vm pod.
So the default provider can satisfy common situations

zhangzujian (approved):

hongzhen-ma:
release-1.9 和 release-1.10 已经cherry-pick回合
master分支，通过 #1561 合入

PR #1565, opened 2022-05-29
fix: delete pod panic when delete vm or statefulset.

xujunjie-cover:
sorry, if vm or statefulset is deleted, vm or statefulset  != nil; so need use err == isNotFound.

What type of this PR
- Bug fixes


pkg/controller/pod.go
```diff
@@ -1354,14 +1354,24 @@ func appendCheckPodToDel(c *Controller, pod *v1.Pod, ownerRefName, ownerRefKind
 	var ownerRefSubnetExist bool
 	switch ownerRefKind {
 	case "StatefulSet":
-		ss, _ := c.config.KubeClient.AppsV1().StatefulSets(pod.Namespace).Get(context.Background(), ownerRefName, metav1.GetOptions{})
-		if ss != nil && ss.Spec.Template.ObjectMeta.Annotations[util.LogicalSwitchAnnotation] != "" {
+		ss, err := c.config.KubeClient.AppsV1().StatefulSets(pod.Namespace).Get(context.Background(), ownerRefName, metav1.GetOptions{})
+		if err != nil {
+			if k8serrors.IsNotFound(err) {
+				return true, nil
+			}
```
hongzhen-ma, line 1361: What about other errors?

hongzhen-ma: It's better to add a log here if don't need to process other errors.

hongzhen-ma (approved):

hongzhen-ma:
已回合release-1.9 和 release-1.10 分支

PR #1601, opened 2022-06-10
Add ebgp-multihop function for kube-ovn-speaker

KillMaster9:
#### What type of this PR
- Feature

When using kube-ovn-speaker to establish an EBGP neighbor with a BGP client outside the cluster, the neighbor establishment fails because the EBGP TTL defaults to 1. 
Therefore, increase the ebgp-mutihop parameter for kube-ovn-speaker to set the EBGP TTL value.

you can set the args --ebgp-multihop=64, set the ebgp multihopTtl，default is 1.

/kind feature

Add Feature


oilbeater (approved):

oilbeater:
Thanks! @KillMaster9

PR #1604, opened 2022-06-13
Modify the next hop calculation method for kube-ovn-speaker

KillMaster9:
#### What type of this PR
- Bug fixes

When kube-ovn-speaker monitors resource changes such as Pods, it will generate an update message and use BGP to advertise it. When generating the update message, the nexthop attribute needs to be calculated. The implementation code is as follows：
```
`func (c *Controller) getNlriAndAttrs(route string) (*anypb.Any, []*anypb.Any, error) {
	prefix, prefixLen, err := parseRoute(route)
	if err != nil {
		return nil, nil, err
	}
	nlri, _ := ptypes.MarshalAny(&bgpapi.IPAddressPrefix{
		Prefix:    prefix,
		PrefixLen: prefixLen,
	})
	a1, _ := ptypes.MarshalAny(&bgpapi.OriginAttribute{
		Origin: 0,
	})
	a2, _ := ptypes.MarshalAny(&bgpapi.NextHopAttribute{
		NextHop: c.config.RouteId,
	})
	attrs := []*anypb.Any{a1, a2}
	return nlri, attrs, err
}`
```
the NextHop = c.config.RouteId, This method is incorrect. if we set the RouteId=1.2.3.4(you can actually set it to whatever you want) , **when the BGP clinet received the update message，parsing the next hop, the routing information cannot be learned because the next hop is unreachable**.

In response to this problem, the solution we propose is to obtain the kernel route according to the longest matching principle based on the BGP NeighborAddress as the dstIP.  then nexthop = route.Src.

/kind bug

### DIFF
--- a/pkg/speaker/config.go
+++ b/pkg/speaker/config.go
@@ -29,6 +29,7 @@ const (
 	DefaultPprofPort                   = 10667
 	DefaultGracefulRestartDeferralTime = 360 * time.Second
 	DefaultGracefulRestartTime         = 90 * time.Second
+	DefaultEbgpMultiHop                = 1
 )
 
 type Configuration struct {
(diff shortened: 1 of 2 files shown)


oilbeater:
@KillMaster9 Thanks! It should work like this. Can you also optimize the bgp.md docs later with the description args you added before and anything that your feel incorrect

PR #1616, opened 2022-06-17
do not snat packets only for subnets with distributed gateway when external traffic policy is set to local

zhangzujian:
#### What type of this PR

- Bug fixes

This patch avoids node port access failure for subnets with a centralized gateway when the external traffic policy is set to Local.


oilbeater (approved):

oilbeater:
should have a test for this

PR #1630, opened 2022-06-20
change vpc gw pod workload from deployment to statefulset

bobz965:
### in order to avoid ip recreate, change vpc gw pod workload from deployment to statefulset

### DIFF
--- a/pkg/controller/gc.go
+++ b/pkg/controller/gc.go
@@ -79,7 +79,7 @@ func (c *Controller) gcVpcNatGateway() error {
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


fanriming:
api定义ippool
新增ip pool 控制器逻辑
These two commits should be independent of the other PR.

bobz965:
> api定义ippool 新增ip pool 控制器逻辑 These two commits should be independent of the other PR.

很抱歉， 不小心 提交了多余的特性，已回退。

fanriming (approved):
LGTM

PR #1640, opened 2022-06-23
fix: make sure pod annotation switch is the first choice to allocate …

bobz965:
#### Bug fixes

1. fix: make sure pod annotation switch is the first choice to allocate 
2. fix: vpc nat gw sts not delete

Fixes #(1631)


bobz965:
related issue: https://github.com/kubeovn/kube-ovn/issues/1631

hongzhen-ma (approved):

PR #1666, opened 2022-07-05
ignore pod not scheduled when reconcile subnet

hongzhen-ma:
#### What type of this PR
- Bug fixes
####
1、when reconcile subnet, all pods in this subnet will be checked and add to port-group. Should ignore pod which has not been scheduled to node.
2、 This is caused by https://github.com/kubeovn/kube-ovn/pull/1655

Fixes #(issue-number)


zhangzujian (approved):

hongzhen-ma:
已回合release-1.10 和 release-1.9 分支

PR #1671, opened 2022-07-07
refactor: extract external routes from eip func, make it the same as …

bobz965:
#### What type of this PR

refactor: extract external routes from eip func, make it the same as vpc subnet route


oilbeater (approved):

bobz965:
cherry pick this commit into 1.10

PR #1684, opened 2022-07-12
Create default vpc with 'Status.Default = true'

### LINKED ISSUE #1683: [discussion] InitDefaultVpc, it's better to set 'vpc.Status.Default = true' during create instead of patch
```
func (c *Controller) InitDefaultVpc() error {
+               orivpc.Status.Default = true
                orivpc, err = c.config.KubeOvnClient.KubeovnV1().Vpcs().Create(context.Background(), orivpc, metav1.CreateOptions{})
        ...
        vpc.Status.Standby = true
-       vpc.Status.Default = true
	_, err = c.config.KubeOvnClient.KubeovnV1().Vpcs().Patch(context.Background(), vpc.Name, types.MergePatchType, bytes, metav1.PatchOptions{}, "status")
```

chestack:
Fixes #1683


oilbeater:
status field cannot be created, should use `UpdateStatus`

chestack:
> status field cannot be created, should use `UpdateStatus`

oopsie, thank you for pointing that out.

Maybe we don't need this change or 
```
func (c *Controller) handleAddOrUpdateVpc(key string) error {
-	orivpc, err := c.vpcsLister.Get(key)
+	// get latest vpc info
+	orivpc, err := c.config.KubeOvnClient.KubeovnV1().Vpcs().Get(context.Background(), key, metav1.GetOptions{})

```

oilbeater (approved):

PR #1726, opened 2022-07-21
fix: wrong info when update subnet from dual to ipv4 or ipv6.

xujunjie-cover:
when update subnet from dual to ipv4 or ipv6, available and used ip not correct

3506c800b925ac4f6d81109028928f4

#### What type of this PR
- Bug fixes


oilbeater:
Looks good to me. But I am wondering what will happens to the ipam when change subnet protocol

xujunjie-cover:
> Looks good to me. But I am wondering what will happens to the ipam when change subnet protocol

When change from v4 or v6 to dual，will add v6 or v4 cidr，excludeips to ipam.
When change from dual to single stack, will do nothing to ipam.

oilbeater:
Got it. It should be changed later, thanks!

oilbeater:
backport down to release-1.8

PR #1762, opened 2022-08-02
support adding routes in underlay Pods for access to overlay Pods

zhangzujian:

#### What type of this PR

- Features

Fixes #(issue-number)

### DIFF
--- a/pkg/daemon/handler.go
+++ b/pkg/daemon/handler.go
@@ -247,7 +247,8 @@ func (csh cniServerHandler) handleAdd(req *restful.Request, resp *restful.Respon
 
 		// routes used for access from underlay to overlay
 		var u2oRoutes []request.Route
-		if podSubnet.Spec.U2oRouting && podSubnet.Spec.Vlan != "" && !podSubnet.Spec.LogicalGateway {
+		if podSubnet.Spec.U2oRouting && podSubnet.Spec.Vlan != "" &&
+			!podSubnet.Spec.LogicalGateway && podSubnet.Spec.Vpc == util.DefaultVpc {
 			subnets, err := csh.Controller.subnetsLister.List(labels.Everything())
 			if err != nil {
 				errMsg := fmt.Errorf("failed to list subnets: %v", err)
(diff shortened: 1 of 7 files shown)


oilbeater (changes requested):

pkg/daemon/handler.go
```diff
@@ -245,14 +245,61 @@ func (csh cniServerHandler) handleAdd(req *restful.Request, resp *restful.Respon
 			mtu = csh.Config.MTU
 		}
 
-		klog.Infof("create container interface %s mac %s, ip %s, cidr %s, gw %s, custom routes %v", ifName, macAddr, ipAddr, cidr, gw, podRequest.Routes)
+		// routes used for access from underlay to overlay
```
oilbeater, line 248: what about overlay to underlay, do we need a nat when across nodes

pkg/daemon/handler.go
```diff
@@ -245,14 +245,61 @@ func (csh cniServerHandler) handleAdd(req *restful.Request, resp *restful.Respon
...
+					return
+				}
+			}
+
+			podCidrV4, podCidrV6 := util.SplitStringIP(cidr)
+			nodeIPv4, nodeIPv6 := util.GetNodeInternalIP(*node)
+			v4Routing := util.CIDRContainIP(podCidrV4, nodeIPv4)
+			v6Routing := util.CIDRContainIP(podCidrV6, nodeIPv6)
+			for _, subnet := range subnets {
+				if subnet.Spec.Vlan == "" || subnet.Spec.LogicalGateway {
```
oilbeater, line 277: should limit subnets to default vpc

oilbeater (changes requested):

pkg/daemon/handler.go
zhangzujian: > what about overlay to underlay, do we need a nat when across nodes

Overlay to underlay has the same packet path with overlay to outside the cluster. Iptables rules introduced in PR #1578 allow this kind of traffic for ICMP and TCP.

oilbeater (approved):

PR #1770, opened 2022-08-03
replace some of acl table with libovsdb

gugulee:

#### What type of this PR
- Features

Fixes #1675 

1. replace some of acl table with libovsdb
2. optimise some code logic

### DIFF
--- a/pkg/ovs/util.go
+++ b/pkg/ovs/util.go
@@ -2,6 +2,7 @@ package ovs
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	kubeovnv1 "github.com/kubeovn/kube-ovn/pkg/apis/kubeovn/v1"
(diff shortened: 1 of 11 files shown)


pkg/ovs/util.go
```diff
@@ -75,3 +76,16 @@ func getIpv6Prefix(networks []string) []string {
 
 	return ipv6Prefix
 }
+
+func matchAddressSetName(asName string) (bool, error) {
+	matched, err := regexp.MatchString(`^[a-zA-Z_.][a-zA-Z_.0-9]*$`, asName)
```
zhangzujian, line 81: We can compile the regular expression in the init function to improve performance.

gugulee: good idea, how about global variables?

```
var addressSetNameRegex = regexp.MustCompile(`^[a-zA-Z_.][a-zA-Z_.0-9]*$`)

func matchAddressSetName(asName string) (bool, error) {
	matched := addressSetNameRegex.MatchString(asName)
	if !matched {
		return false, fmt.Errorf("address set %s must match `[a-zA-Z_.][a-zA-Z_.0-9]*`", asName)
	}

	return true, nil
}
```

zhangzujian: LGTM, but there is no need to introduce the variable `matched`.

zhangzujian (approved):

PR #1775, opened 2022-08-03
fix: chassis id is empty but cannt identify

lut777:

#### What type of this PR
- Bug fixes

A bug shows that the empty sure can not identify the scenario where chassis id is missed in ovn sbdb.
Now judgement is based on the length of chassis id.

### DIFF
--- a/pkg/controller/node.go
+++ b/pkg/controller/node.go
@@ -837,7 +837,7 @@ func (c *Controller) checkChassisDupl(node *v1.Node) error {
 		klog.Errorf("failed to get node %s chassisID, %v", node.Name, err)
 		return err
 	}
-	if chassisAdd == "" {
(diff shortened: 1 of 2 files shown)


pkg/controller/node.go
```diff
@@ -837,7 +837,7 @@ func (c *Controller) checkChassisDupl(node *v1.Node) error {
 		klog.Errorf("failed to get node %s chassisID, %v", node.Name, err)
 		return err
 	}
-	if chassisAdd == "" {
+	if len(chassisAdd) < util.ChassisIDLength {
```
oilbeater, line 840: chassis 不是空是什么？

lut777:
abnormal scenario shows up in 1.9.

PR #1817, opened 2022-08-16
release-1.8: do not check static route conflict

zhangzujian:

#### What type of this PR

- Bug fixes

Fixes #(issue-number)

Duplicate IP addresses will block LSP creation so there is no need to check static route conflict. This patch fixes the following error:

```txt
static route "policy=src-ip ip_prefix=10.3.0.132" with different nexthop already exists on logical router ovn-cluster, requeuing
```

### DIFF
--- a/pkg/controller/pod.go
+++ b/pkg/controller/pod.go
@@ -558,70 +558,35 @@ func (c *Controller) handleDeletePod(pod *v1.Pod) error {
 
 	p, _ := c.podsLister.Pods(pod.Namespace).Get(pod.Name)
 	if p != nil && p.UID != pod.UID {
-		// The existing OVN static route with a different nexthop will block creation of the new Pod,
-		// so we need to check the node names
-		if pod.Spec.NodeName == "" || pod.Spec.NodeName == p.Spec.NodeName {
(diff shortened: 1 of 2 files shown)


pkg/controller/pod.go
```diff
@@ -558,72 +558,37 @@ func (c *Controller) handleDeletePod(pod *v1.Pod) error {
...
+		if err := c.ovnClient.DeleteStaticRoute(address.Ip, vpc.Status.Router); err != nil {
+			return err
+		}
+		if err := c.ovnClient.DeleteNatRule(address.Ip, vpc.Status.Router); err != nil {
+			return err
 		}
 	}
 
-	ports, err := c.ovnClient.ListPodLogicalSwitchPorts(podName, pod.Namespace)
+	ports, err := c.ovnClient.ListPodLogicalSwitchPorts(pod.Name, pod.Namespace)
```
hongzhen-ma, line 586: 这里要用 podName，vm 保留IP的时候，podName 是 vmName ，部署 vm pod name

hongzhen-ma (changes requested):

PR #1824, opened 2022-08-17
fix: use full longest word to match full ip about dnat

bobz965:

#### What type of this PR
- Bug fixes

if not use full words (ip/32) to match will cause some ip not create fip


oilbeater:
Will be cherry-picked from master

PR #1848, opened 2022-08-24
fix ovs-ovn logging

zhangzujian:
tail command should follow file name instead of file descriptor

#### What type of this PR

- Bug fixes

After the log file has been rotated, tail stops printing the log, and still holds the file descriptor:

```txt
tail  8435  root 3r  REG  253,0  17385  67521247 /var/log/ovn/ovn-controller.log.1 (deleted)
```

This patch is part of fixes for #1849.


oilbeater:
for version before 1.10 ovn-central logs also need modification

PR #1865, opened 2022-08-29
cni should handler unmont volume, when delete pod

wlooppy:
- Bug fixes

cni should handler unmont volume, when delete pod.


oilbeater:
@wlooppy there are conflict with the master could you rebase your branch to the master?

PR #1873, opened 2022-08-31
cni should handler unmounted volume, when delete pod.

wlooppy:
- Bug fixes
cni should handler unmounted volume, when delete pod.


oilbeater:
need run `go mod tidy` and update the `go.sum` and `go.mod` files.

zhangzujian (approved):

PR #1951, opened 2022-10-03
Update install.sh

runzhliu:
Is there any particular reason for the default registry for ovn-central container?  If not, would it be better to have the same style as others :)

#### What type of this PR
- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)


zhangzujian (approved):
LGTM, thanks!

PR #2025, opened 2022-11-08
Mark node as NotReady when ovs-ovn failed, and use io.Copy() maked k8s cni config  at kube-ovn-daemon 

### LINKED ISSUE #1475: Mark node as NotReady when ovs-ovn failed
### Feature request

When ovs-ovn on when node fails, scheduler will still puts pods into this node and leads further problems.

We need to mark the node as NotReady to prevent this, one simple way is that kube-ovn-cni prob check if ovs is not ready and remove the cni conf.

sober-wang:
#### What type of this PR
Features 
- #1475 
- use io.Copy() replace os.ReadFile() next, os.WriteFile()  

Fixes #1475


oilbeater:
We need to reconsider if we should mark it as not ready and prevent schedule, close it at the the moment

PR #2070, opened 2022-11-22
fix libovsdb issues

zhangzujian:

#### What type of this PR

- Bug fixes

1. avoid tcp connection stuck by setting TCP_USER_TIMEOUT socket option;
2. fix connect/reconnect timeout;
3. ensure cache is consistent before reading data.

After a power/network failure, the libovsdb client cannot communicate with the ovn nb leader anymore. This causes the client-side TCP connection to be stuck for a long time.

By setting the TCP socket TCP_USER_TIMEOUT, the client will be able to close the connection and try to connect to other servers.

This PR also:

1. Use klog as the logger of libovsdb, so that libovsdb logs can be written to the log file;
2. Regenerate ovn nb packages.


zhangzujian:
Reference: https://github.com/golang/go/issues/31490#issuecomment-544140457

zhangzujian:
libovsdb patch: https://github.com/kubeovn/libovsdb/commit/3e352fe182635596163b74eac7f84b8ab88eff17

go.mod
```diff
@@ -16,17 +16,16 @@ require (
 	github.com/coreos/go-iptables v0.6.0
 	github.com/emicklei/go-restful/v3 v3.9.0
 	github.com/evanphx/json-patch v5.6.0+incompatible
-	github.com/go-logr/stdr v1.2.2
 	github.com/greenpau/ovsdb v1.0.2
 	github.com/k8snetworkplumbingwg/network-attachment-definition-client v1.3.0
+	github.com/kubeovn/libovsdb v0.0.0-20221123021657-3e352fe18263
```
oilbeater, line 21: we can use go mod replace to point to a new path and avoid all the import rename

oilbeater (changes requested):

oilbeater (approved):

oilbeater (approved):

PR #1480, opened 2022-04-25
fix pod annotation may be overridden by patch

halfcrazy:
#### What type of this PR
Examples of user facing changes:
- Bug fixes

#### Which issue(s) this PR fixes:
Fixes When other controllers add annotation before kube-ovn submit the patch, the annotation may be overridden by kube-ovn.

PR #2099, opened 2022-11-30
optimize provider network

fanriming:

#### What type of this PR
- Bug fixes

Currently, the same provider-network resource will be updated by all kube-ovn-cni pods. When the number of nodes is large and there are several inconsistent states, update events will trigger all kube-ovn-cni to patch provider-network again and again. In this case, the apiserver will receive a large number of patch requests and the cpu will be high load,  and the apiserver will become very slow to respond.

The state of provider-network does not need to be changed frequently. provider-network should be read-only to the kube-ovn-cni. Update operations of provider-network should be minimized to avoid conflicts and inconsistencies. The optimization is as follows:
1. Removes the code that performs updates to provider-network in kube-ovn-cni.
2. Write error messages to kube-ovn-cni pod annotations.
3. Optimize controller timing synchronization code to update provider-network status based on node and kube-ov-cni annotations.
4. Add the `notReadyNodes` field to the status for clearer display of configuration results.


pkg/controller/provider-network.go
```diff
@@ -4,126 +4,58 @@ import (
...
 
-	for _, pn := range pnList {
-		if providerNetworkIsReady(pn, nodeList) != pn.Status.Ready {
-			newPn := pn.DeepCopy()
-			newPn.Status.Ready = !newPn.Status.Ready
-			_, err = c.config.KubeOvnClient.KubeovnV1().ProviderNetworks().UpdateStatus(context.Background(), newPn, metav1.UpdateOptions{})
-			if err != nil {
-				klog.Errorf("failed to update status of provider network %s: %v", pn.Name, err)
+	for _, pn := range pns {
+		var readyNodes, notReadyNodes []string
```
zhangzujian, line 28: The listed results must be treated as read-only, so we should modify the copied object:
```go
for _, cachedPn := range pns {
    pn := cachedPn.DeepCopy();
    // ...
}
```

pkg/controller/provider-network.go
```diff
@@ -4,126 +4,58 @@ import (
...
 
-func (c *Controller) enqueueUpdateProviderNetwork(_, obj interface{}) {
-	if !c.isLeader() {
-		return
-	}
-
-	key, err := cache.MetaNamespaceKeyFunc(obj)
+func (c *Controller) resyncProviderNetworkStatus() {
+	klog.Infof("start to sync ProviderNetwork status")
+	pns, err := c.providerNetworksLister.List(labels.Everything())
```
zhangzujian, line 15: Can we refactor this function and make provider-network status updates triggered by node label updates?

pkg/controller/provider-network.go
```diff
@@ -4,126 +4,58 @@ import (
...
-
-func providerNetworkIsReady(pn *kubeovnv1.ProviderNetwork, nodes []*corev1.Node) bool {
-	for _, node := range nodes {
-		if !util.ContainsString(pn.Spec.ExcludeNodes, node.Name) &&
-			!util.ContainsString(pn.Status.ReadyNodes, node.Name) {
-			return false
+		pn.Status.ReadyNodes = readyNodes
+		pn.Status.NotReadyNodes = notReadyNodes
+		pn.Status.Ready = (len(notReadyNodes) == 0)
+		if _, err = c.config.KubeOvnClient.KubeovnV1().ProviderNetworks().UpdateStatus(context.Background(), pn, metav1.UpdateOptions{}); err != nil {
```
zhangzujian, line 57: Before applying update, we can check whether the status has updates. If no changes occurs, we can skip it.

pkg/daemon/controller.go
```diff
@@ -318,31 +292,51 @@ func (c *Controller) initProviderNetwork(pn *kubeovnv1.ProviderNetwork, node *v1
...
-		return nil
+	if currentPod.Annotations == nil {
+		currentPod.Annotations = make(map[string]string)
 	}
-
-	_, err := c.config.KubeOvnClient.KubeovnV1().ProviderNetworks().UpdateStatus(context.Background(), pn, metav1.UpdateOptions{})
-	if err != nil {
-		klog.Errorf("failed to update status of provider network %s: %v", pn.Name, err)
-		return err
+	if errMsg == "" {
```
zhangzujian, line 329: We should operate on `currentPod.DeepCopy()` instead of the cached object.

pkg/daemon/controller.go
```diff
@@ -318,31 +292,51 @@ func (c *Controller) initProviderNetwork(pn *kubeovnv1.ProviderNetwork, node *v1
...
+	} else {
+		currentPod.Annotations[fmt.Sprintf(util.ProviderNetworkErrMessageTemplate, providerNetwork)] = errMsg
 	}
 
-	return nil
+	patchPayload := []byte(fmt.Sprintf(`[{ "op": "replace", "path": "/metadata/annotations", "value": %s }]`, currentPod.Annotations))
+	if _, err = c.config.KubeClient.CoreV1().Pods(c.localNamespace).Patch(context.Background(), c.localPodName, types.JSONPatchType, patchPayload, metav1.PatchOptions{}); err != nil {
+		klog.Errorf("failed to patch pod %s: %v", c.localPodName, err)
+		return
+	}
```
zhangzujian, line 339: We should use the strategic merge patch method.

https://github.com/kubeovn/kube-ovn/pull/1480

zhangzujian: We should also check whether the annotations are changed before applying the patch.

pkg/controller/provider-network.go
fanriming: In my opinion, it is not necessary, because the state is only the presentation of the final result, and the delay in the presentation of the result will not cause functional impact. We should minimize the update operation to avoid the disturbance to the cluster, especially in the case of a large cluster size.

fanriming: > Can we refactor this function and make provider-network status updates triggered by node label updates?

@zhangzujian In my opinion, it is not necessary, because the state is only the presentation of the final result, and the delay in the presentation of the result will not cause functional impact. We should minimize the update operation to avoid the disturbance to the cluster, especially in the case of a large cluster size.

fanriming:
@zhangzujian  It has been modified as suggested, please review it again, thanks.

zhangzujian:
Please run `go fmt ./...` to fix gofmt issues.

pkg/daemon/controller.go
```diff
@@ -318,31 +292,57 @@ func (c *Controller) initProviderNetwork(pn *kubeovnv1.ProviderNetwork, node *v1
...
+			delete(newPod.Annotations, fmt.Sprintf(util.ProviderNetworkErrMessageTemplate, providerNetwork))
+		} else {
+			newPod.Annotations[fmt.Sprintf(util.ProviderNetworkErrMessageTemplate, providerNetwork)] = errMsg
+		}
+		patch, err := util.GenerateStrategicMergePatchPayload(currentPod, newPod)
+		if err != nil {
+			klog.Errorf("failed to gen patch payload pod %s: %v", c.localPodName, err)
+			return
+		}
+		if _, err = c.config.KubeClient.CoreV1().Pods(c.localNamespace).Patch(context.Background(), c.localPodName, types.JSONPatchType, patch, metav1.PatchOptions{}); err != nil {
```
zhangzujian, line 341: Should use `StrategicMergePatchType` instead of `JSONPatchType`.

pkg/controller/provider-network.go
```diff
@@ -4,126 +4,72 @@ import (
...
+				pn.Status.SetNodeReady(node.Name, "InitOVSBridgeSucceeded", "")
+				readyNodes = append(readyNodes, node.Name)
+			} else {
+				pods, err := c.config.KubeClient.CoreV1().Pods("").List(context.Background(), metav1.ListOptions{
+					LabelSelector: "app=kube-ovn-cni",
+					FieldSelector: fmt.Sprintf("spec.nodeName=%s", node.Name),
+				})
+				if err != nil {
+					klog.Errorf("failed to list pod: %v", err)
+				}
```
zhangzujian, line 45:
```suggestion
				if err != nil {
					klog.Errorf("failed to list pod: %v", err)
					continue
				}
```

pkg/controller/provider-network.go
```diff
@@ -4,126 +4,73 @@ import (
...
-		if providerNetworkIsReady(pn, nodeList) != pn.Status.Ready {
-			newPn := pn.DeepCopy()
-			newPn.Status.Ready = !newPn.Status.Ready
-			_, err = c.config.KubeOvnClient.KubeovnV1().ProviderNetworks().UpdateStatus(context.Background(), newPn, metav1.UpdateOptions{})
-			if err != nil {
-				klog.Errorf("failed to update status of provider network %s: %v", pn.Name, err)
+		expectNodes = append(readyNodes, notReadyNodes...)
+		conditionsChange := false
+		for _, c := range pn.Status.Conditions {
+			if util.ContainsString(expectNodes, c.Node) {
```
zhangzujian, line 60:
```suggestion
			if !util.ContainsString(expectNodes, c.Node) {
```

PR #2132, opened 2022-12-06
fix: sometimes allocate IPv6 address failed but there is available IP in V6FreeIPList

changluyi:
sometimes ipam.GetStaticAddress return NoAvailableAddress but there is available IP in V6FreeIPList

when use ipam.GetStaticAddress with param ipv6 address without clean Zero such as "fd00::00e9:0ed1", and exactly subnet.V6FreeIPList has a range with start fd00::e9:ed1 and end with any v6 address, or start with an any v6 address and end with fd00::e9:ed1, it will cause ipr.Start.Equal(ip) or ipr.End.Equal(ip) failed, Because equal only check the string match .

#### What type of this PR
- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)

### DIFF
--- a/pkg/ipam/subnet.go
+++ b/pkg/ipam/subnet.go
@@ -332,6 +332,11 @@ func (subnet *Subnet) GetStaticAddress(podName, nicName string, ip IP, mac strin
 	if v4 && !subnet.V4CIDR.Contains(net.ParseIP(string(ip))) {
 		return ip, mac, ErrOutOfRange
 	}
+
+	if v6 {
+		ip = IP(net.ParseIP(string(ip)).String())
+	}
+
 	if v6 && !subnet.V6CIDR.Contains(net.ParseIP(string(ip))) {
 		return ip, mac, ErrOutOfRange
 	}


pkg/ipam/ip.go
```diff
@@ -15,7 +16,7 @@ type IPRange struct {
 }
 
 func (a IP) Equal(b IP) bool {
-	return a == b
+	return net.ParseIP(string(a)).Equal(net.ParseIP(string(b)))
```
zhangzujian, line 19:
```suggestion
	return a == b || net.ParseIP(string(a)).Equal(net.ParseIP(string(b)))
```

zhangzujian: Where is `Equal()` used? Is there any chance that `net.ParseIP(string(a))` may return nil?

changluyi: image

changluyi: in splitIPRangeList

PR #2148, opened 2022-12-09
delete nc cmd in image

hongzhen-ma:

#### What type of this PR
- Bug fixes

Fixes #(issue-number)

### DIFF
--- a/dist/images/Dockerfile
+++ b/dist/images/Dockerfile
@@ -9,6 +9,9 @@ COPY grace_stop_ovn_controller /usr/share/ovn/scripts/grace_stop_ovn_controller
 
 WORKDIR /kube-ovn
 
+RUN rm -f /usr/bin/nc &&\
+    rm -f /usr/bin/netcat
+
 COPY kube-ovn-cmd /kube-ovn/kube-ovn-cmd
 COPY kube-ovn-webhook /kube-ovn/kube-ovn-webhook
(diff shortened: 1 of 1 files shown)


dist/images/Dockerfile
```diff
@@ -9,6 +9,9 @@ COPY grace_stop_ovn_controller /usr/share/ovn/scripts/grace_stop_ovn_controller
 
 WORKDIR /kube-ovn
 
+RUN rm -f /usr/bin/nc &&\
```
oilbeater, line 12: we should remove ncat from base

oilbeater (changes requested):

dist/images/Dockerfile
hongzhen-ma: Yes. The nc cmd is installed in nmap packet, so I delete the nmap packet in the last PR https://github.com/kubeovn/kube-ovn/pull/2141
But the nc cmd still exists because the ncat is installed in dockerfile.base.

The jira about security only reports cmd about nc and netcat, both are soft link from ncat.
image

So if we can keep ncat and delete the soft link cmd? It will be convennient if we want to use nc in the pod.
Or It's better to delete ncat totally?

oilbeater: ok

oilbeater (approved):

PR #2195, opened 2022-12-27
local ip bind to service

changluyi:

#### What type of this PR
- Features
for security purposes,  local services should bind to local IP but not 0.0.0.0 or ::
- Bug fixes
- Docs
- Tests

when deploying below deployment or ds :kube-ovn-controler/kube-ovn-monitor/kube-ovn-cni/ovn-central

add an env variable
            - name: ENABLE_BIND_LOCAL_IP
              value: "true"

it will work like below
```
root@kube-ovn-worker:/kube-ovn# netstat -tunlp | grep kube-ovn
tcp6       0      0 fc00:f853:ccd:e79:10665 :::*                    LISTEN      1805/./kube-ovn-dae
tcp6       0      0 fc00:f853:ccd:e79:10660 :::*                    LISTEN      1468/./kube-ovn-con
```
this feature will not work when the network is in the dual mode because golang can't support binding v4 and v6 addresses at the same time.

Fixes #(issue-number)
in jira ACP-20335

### DIFF
--- a/cmd/controller_health_check/controller_health_check.go
+++ b/cmd/controller_health_check/controller_health_check.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	kubeovnv1 "github.com/kubeovn/kube-ovn/pkg/apis/kubeovn/v1"
 	"github.com/kubeovn/kube-ovn/pkg/ovs"
 	"github.com/kubeovn/kube-ovn/pkg/util"
 )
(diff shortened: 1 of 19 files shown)


cmd/controller_health_check/controller_health_check.go
```diff
@@ -23,7 +24,22 @@ func CmdMain() {
 	if err := ovs.CheckAlive(); err != nil {
 		os.Exit(1)
 	}
-	conn, err := net.DialTimeout("tcp", "127.0.0.1:10660", 3*time.Second)
+
+	addr := "127.0.0.1:10660"
```
oilbeater, line 28: health_check should only bind to localhost

dist/images/install.sh
```diff
@@ -2023,6 +2023,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
```
oilbeater, line 2025: should be able to set ENABLE_BIND_LOCAL_IP from install.sh and helm

oilbeater (changes requested):

cmd/controller_health_check/controller_health_check.go
changluyi: if kube-ovn-controller bind local IP, eg 172.18.0.2:10660,   the controller health check service will get the fault "connection refused " when accessing the 127.0.0.1 10660

oilbeater: ok

dist/images/install.sh
changluyi: I set ENABLE_BIND_LOCAL_IP = true, on another pr , and it seems ok when running workflows

oilbeater: This option should be configurable

oilbeater (changes requested):

dist/images/install.sh
changluyi: it can be on/off, the pod will restart using a new env value.

changluyi: fixed

oilbeater (approved):

PR #2247, opened 2023-01-16
fix setting mtu for ovs internal port

zhangzujian:

#### What type of this PR

- Bug fixes

Changing MTU of ovs internal port with iproute2/netlink may not work, we should set interface mtu_request instead.

For OVS bridges, the MTU will be automatically adjusted, so there is no need to set MTU manually.


changluyi:
did the mirror nic port need to fix in the function configureMirrorLink ? I see the nic type is also openvswitch

zhangzujian:
> did the mirror nic port need to fix in the function configureMirrorLink ? I see the nic type is also openvswitch

fixed

changluyi (approved):

PR #2276, opened 2023-01-30
move enableLb to subnet

### LINKED ISSUE #1949: Move enable-lb option to subnet level
### Feature request

Now `enable-lb` is a global option, however we found that the ovn lb behavior doesn't match Kubernetes Service in some details. For example round-robin and sticky time are not support by ovn lb. But we still need ovn-lb to support Service basic functionality at the moment and we support hybrid mode that run both overlay and underlay subnet at the same time. We want to disable ovn-lb in overlay subnet and enable it in underlay subnet. So we need to move the `enable-lb` option from a global option to a subnet level option.

hongzhen-ma:

#### What type of this PR
- Features
- move enableLb to subnet

Fixes #1949

### DIFF
--- a/pkg/controller/subnet.go
+++ b/pkg/controller/subnet.go
@@ -93,6 +93,8 @@ func (c *Controller) enqueueUpdateSubnet(old, new interface{}) {
 		oldSubnet.Spec.EnableIPv6RA != newSubnet.Spec.EnableIPv6RA ||
 		oldSubnet.Spec.IPv6RAConfigs != newSubnet.Spec.IPv6RAConfigs ||
 		oldSubnet.Spec.Protocol != newSubnet.Spec.Protocol ||
+		(oldSubnet.Spec.EnableLb != nil && newSubnet.Spec.EnableLb == nil) ||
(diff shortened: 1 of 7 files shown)


pkg/controller/subnet.go
```diff
@@ -93,6 +93,8 @@ func (c *Controller) enqueueUpdateSubnet(old, new interface{}) {
 		oldSubnet.Spec.EnableIPv6RA != newSubnet.Spec.EnableIPv6RA ||
 		oldSubnet.Spec.IPv6RAConfigs != newSubnet.Spec.IPv6RAConfigs ||
 		oldSubnet.Spec.Protocol != newSubnet.Spec.Protocol ||
+		oldSubnet.Spec.EnableLb != nil && newSubnet.Spec.EnableLb == nil ||
```
zhangzujian, line 96:
```suggestion
		(oldSubnet.Spec.EnableLb != nil && newSubnet.Spec.EnableLb == nil) ||
```

hongzhen-ma: modified

zhangzujian (approved):

PR #2282, opened 2023-02-01
fix wrong network interface name in gateway check

qiutingjun:
- Bug fixes

fix wrong network interface name in gateway check


oilbeater:
Thanks! @qiutingjun

PR #2290, opened 2023-02-05
perform the gateway check but ignore the result when the annotation of subnet is ‘disableGatewayCheck=true’ to make sure of the first network packet

qiutingjun:
- Features

Under the underlay network mode, there are many situations will cause the underlay network arping fails, so there is a highly probability of gateway check failure.
If the gateway check is not performed, the pod or vm cannot be pinged from the outside after it is pulled up, because of there is no first network packet.
So I expect the gateway check is still performed but the result does not concerned when the annotation of subnet is ‘disableGatewayCheck=true’.


oilbeater:
The `waitNetworkReady` will try 200 times before timeout. If only need to send a first packet, we should not run the probe 200 times. `kube-ovn-cni` will hold a goroutine for each unfinished probe and it will also send lots of unnecessary packets into underlay network.

qiutingjun:
> The `waitNetworkReady` will try 200 times before timeout. If only need to send a first packet, we should not run the probe 200 times. `kube-ovn-cni` will hold a goroutine for each unfinished probe and it will also send lots of unnecessary packets into underlay network.

Thank you for reminding me about the setting of retry times，and I've added a parameter to control the max retries.

oilbeater (approved):

pkg/daemon/ovs_linux.go
```diff
@@ -339,36 +339,41 @@ func configureContainerNic(nicName, ifName string, ipAddr, gateway string, isDef
...
 
 		return nil
 	})
 }
 
-func waitNetworkReady(nic, ipAddr, gateway string, underlayGateway, verbose bool) error {
+func checkGatewayReady(gwCheckMode int, intr, ipAddr, gateway string, underlayGateway, verbose bool) error {
+	if gwCheckMode == gatewayCheckModeArpingNotConcerned || gwCheckMode == gatewayCheckModePingNotConcerned {
+		go waitNetworkReady(intr, ipAddr, gateway, underlayGateway, verbose, 1)
+		return nil
```
oilbeater, line 366: The waitNetworkReady with only one probe should return fast enough. No need to run in a goroutine and it will lead the lint error

oilbeater (changes requested):

oilbeater (approved):

oilbeater:
Thanks! @qiutingjun

PR #2297, opened 2023-02-06
fix: ovs-ovn should reboot now

lut777:

#### What type of this PR
- Bug fixes

Fixes #

Helm upgrade should change yaml for daemonset ovs-ovn, which should leads to reboot ovs-ovn pod.
But reboot policy for ovs-ovn is `ondelete`, so pod will not reboot.
Now a post-upgrade job will be revoked and reboot pod.


oilbeater:
The lint err should be fixed

lut777:
> 

Done.
PR for submariner install and test is involved unintentional.

oilbeater (approved):

PR #2316, opened 2023-02-09
Add the bgp router-id format check

KillMaster9:
#### What type of this PR
- Bug fixes

As a key parameter of bgp, router-id needs to be verified.
Currently, this parameter is not properly verified in the gobgp module. 
Validation after parsing configuration parameters reduces unnecessary grpc calls.


oilbeater (approved):

oilbeater:
Thanks! @KillMaster9

PR #2335, opened 2023-02-14
Add neighbor-address format check for kube-ovn-speaker

KillMaster9:
#### What type of this PR
- Bug fixes

Similar to pr #2316 
The neighbor-address parameter is not properly verified in the gobgp module.


oilbeater:
Thanks! @KillMaster9

Issue #2047, opened 2022-11-16
Support for specifying kubelet directory


lynn901:
### Feature request

Support for specifying kubelet directory

### Use case
In some k8s clusters, users specify the kubelet directory instead of using the default /var/lib/kubelet

oilbeater:
Why kube-ovn need to know the place of kubelet? I don't know any code in kube-ovn that will call kubelet directly.

oilbeater:
Thanks for the information. It's really should be configurable.

PR #2388, opened 2023-02-27
add support of user-defined kubelet directory

qiutingjun:
#### What type of this PR
- Features

Add configurations of user-defined kubelet directory to support users specifying the kubelet directory instead of using the default /var/lib/kubelet.

Fixes https://github.com/kubeovn/kube-ovn/issues/2047


oilbeater (approved):

oilbeater:
Thanks! @qiutingjun

PR #2476, opened 2023-03-13
Add jitter support to netem qos parameters

lime008:
Adds jitter parameter and annotation support for the netem QoS options.


hongzhen-ma:
@lime008 Thanks very much for the PR.

Api in windows build also need to be adapted
```
pkg/daemon/controller_windows.go:217: too many arguments in call to ovs.SetNetemQos
	have (string, string, string, string, string, string, string)
	want (string, string, string, string, string, string)
```

lime008:
The windows api should now also be adapted to include the jitter parameter.

hongzhen-ma (approved):

PR #2516, opened 2023-03-19
underlay: fix link name exchange

zhangzujian:

#### What type of this PR

- Bug fixes

In operating systems using systemd-networkd as network manager, we should wait for systemd-networkd to finish address/route configuration.

### DIFF
--- a/pkg/daemon/init_linux.go
+++ b/pkg/daemon/init_linux.go
@@ -45,6 +45,34 @@ func nmSetManaged(device string, managed bool) error {
 	return nil
 }
 
+// wait systemd-networkd to finish interface configuration
+func waitNetworkdConfiguration(linkIndex int) {
+	done := make(chan struct{})
+	ch := make(chan netlink.RouteUpdate)
+	if err := netlink.RouteSubscribe(ch, done); err != nil {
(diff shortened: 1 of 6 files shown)


pkg/daemon/init_linux.go
```diff
@@ -90,15 +90,43 @@ func changeProvideNicName(current, target string) (bool, error) {
 		klog.Errorf("failed to set link %s up: %v", target, err)
 		return false, err
 	}
+	klog.Infof("link %s has been renamed as %s", current, target)
+
+	// wait systemd-networkd to finish interface configuration
```
oilbeater, line 95: Does 50 millisecond enough for networkd to configure the interface? And this part should be moved to a separated function.

oilbeater (changes requested):

pkg/daemon/init_linux.go
zhangzujian: `systemd-networkd` monitors netlink events and responds quickly. 50ms is enough in VM tests.

oilbeater (approved):

oilbeater (approved):

PR #2525, opened 2023-03-20
add vip to webhook e2e

ShaPoHun:

#### What type of this PR
- Features
add vip to webhook e2e

### DIFF
--- a/test/e2e/framework/vip.go
+++ b/test/e2e/framework/vip.go
@@ -28,7 +28,7 @@ func (f *Framework) VipClient() *VipClient {
 func (c *VipClient) Get(name string) *apiv1.Vip {
 	vip, err := c.VipInterface.Get(context.TODO(), name, metav1.GetOptions{})
 	ExpectNoError(err)
-	return vip
+	return vip.DeepCopy()
 }
 
 // Create creates a new vip according to the framework specifications
--- a/test/e2e/framework/vip.go
+++ b/test/e2e/framework/vip.go
@@ -1,0 +1,87 @@
+package framework
+
+import (
+	"context"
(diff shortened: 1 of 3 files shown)


test/e2e/framework/vip.go
```diff
@@ -0,0 +1,87 @@
...
+	ExpectNoError(err, "Error creating vip")
+	return vip.DeepCopy()
+}
+
+// Patch patches the vip
+func (c *VipClient) Patch(original, modified *apiv1.Vip, timeout time.Duration) *apiv1.Vip {
+	patch, err := util.GenerateMergePatchPayload(original, modified)
+	ExpectNoError(err)
+
+	var patchedVlan *apiv1.Vip
```
hongzhen-ma, line 46: patchedVlan ？

test/e2e/framework/vip.go
```diff
@@ -0,0 +1,87 @@
...
+	return &VipClient{
+		f:            f,
+		VipInterface: f.KubeOVNClientSet.KubeovnV1().Vips(),
+	}
+}
+
+func (c *VipClient) Get(name string) *apiv1.Vip {
+	vip, err := c.VipInterface.Get(context.TODO(), name, metav1.GetOptions{})
+	ExpectNoError(err)
+	return vip
```
zhangzujian, line 31:
```suggestion
	return vip.DeepCopy()
```

zhangzujian:
From which version VIP is introduced? We should skip the test if the feature is not supported.

ShaPoHun:
> From which version VIP is introduced? We should skip the test if the feature is not supported.

Support for vip Webhooks starts with tag v1.11.2

zhangzujian:
Since webhook e2e is running only on the master branch, there is no need to check the version.

hongzhen-ma (approved):

PR #2544, opened 2023-03-23
fix lsp deletion failure when external-ids:ls is empty

zhangzujian:

#### What type of this PR

- Bug fixes

During the upgrade, LSP gc for terminated kube-ovn-pinger failed since the external ID is empty and will never be set in kube-ovn-controller initialization.


zhangzujian:
@gugulee Please review.

gugulee (approved):

PR #2554, opened 2023-03-26
fix update snat rules not effect correctly

qiutingjun:

#### What type of this PR
- Bug fixes

While update the internal CIDR of the snat rule, it will not really take effect in the vpc nat gateway pod, the rule k8s shows does not match the actual rule in gateway pod.

### DIFF
--- a/dist/images/install.sh
+++ b/dist/images/install.sh
@@ -739,6 +739,8 @@ spec:
                   type: string
                 redo:
                   type: string
+               internalCIDR:
+                  type: string
                 conditions:
                   type: array
                   items:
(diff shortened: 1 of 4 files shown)


bobz965 (approved):

hongzhen-ma (approved):

hongzhen-ma:
上边机器人 review意见第三条
There is a performance issue in the handleUpdateIptablesSnatRule function where it deletes and creates SNAT rules even if there is no change in the internal CIDR. This can be optimized to avoid unnecessary operations.

是否需要考虑下？

qiutingjun:
> 上边机器人 review意见第三条 There is a performance issue in the handleUpdateIptablesSnatRule function where it deletes and creates SNAT rules even if there is no change in the internal CIDR. This can be optimized to avoid unnecessary operations.
> 
> 是否需要考虑下？

The changes has been checked in function enqueueUpdateIptablesSnatRule.
If there are no changes, the rules will not be passed here for processing.

PR #3107, opened 2023-08-02
简化 ovn eip 类型

bobz965:
- [ ] Make sure you have followed [Kube-OVN Code Style](https://github.com/kubeovn/kube-ovn/blob/master/CODE_STYLE.md).

### What type of this PR
修复 bfd 路由无法新增多个
修复 vpc 更新路由未触发

### <samp>🤖 Generated by Copilot at 9cae868</samp>

Refactor the code to use new and simplified types for ovn eip and vip scenarios. This improves the readability and consistency of the code and the e2e test, and aligns with the ovn terminology. The main changes are in `pkg/controller`, `pkg/daemon`, `pkg/apis`, `pkg/util`, and `test/e2e`.

### <samp>🤖 Generated by Copilot at 9cae868</samp>

> _We are the masters of the ovn eip_
> _We use the `Lrp`, `Lsp`, and `Vip`_
> _We simplify the types and the names_
> _We crush the bugs and the flames_

### <samp>🤖 Generated by Copilot at 9cae868</samp>

*  Simplify the eip usage types from `eip`, `lrp`, and `node external gw` to `lrp`, `lsp`, and `nat` in the `OvnEipSpec` type definition ([link](https://github.com/kubeovn/kube-ovn/pull/3107/files?diff=unified&w=0#diff-69f75d53361877cc02ab6f413f306486d29287c6400b92b71c4c74f018f3234aL952-R955))
*  Use the new `Lrp` type instead of the old `LrpUsingEip` type when creating or updating the ovn eip cr for the logical router port in `createDefaultVpcLrpEip` and `handleAddVpcExternal` functions ([link](https://github.com/kubeovn/kube-ovn/pull/3107/files?diff=unified&w=0#diff-49e9f500bbdd1972f437e33413a4e9761aaa80c62cb2b31fe92b28c8b069333fL203-R203), [link](https://github.com/kubeovn/kube-ovn/pull/3107/files?diff=unified&w=0#diff-ae8a8b5d44d277a4bb96635ee325fe54b48e8bc10b24a79d4bb8e211b57131ecL815-R815))
[...]

PR #3302, opened 2023-10-15
Don't enqueue VPC update when DeletionTimestamp is zero

### LINKED ISSUE #3301: VPC update is always enqueued leading to huge number of updates when creating subnet
# Expected Behavior

When creating a subnet the VPC shouldn't be updated a large amount of times.

# Actual Behavior

Creating a subnet after a VPC already has two subnets leads to a huge amount of updates.

I assume that this has to do with the following line:

https://github.com/kubeovn/kube-ovn/blob/60ef1ad1f31d2974cb9462670b0ad26b1015be6f/pkg/controller/vpc.go#L47

Before #3107 this was (note the missing `!` in the line above):

https://github.com/kubeovn/kube-ovn/blob/68dc1c38023e4967e9644c89e72e63c605df8591/pkg/controller/vpc.go#L56

I'm unsure why the `!` should be removed here, as that leads to the `c.addOrUpdateVpcQueue.Add(key)` basically being called every time, which leads to a huge amount of (unnecessary) updates.

1. Create a VPC

  ```yaml
  kind: Vpc
  apiVersion: kubeovn.io/v1
  metadata:
    name: test-vpc
  ```

2. Create a subnet and bind it to the VPC. Check the `kube-ovn-controller` logs. Shortly after you created the subnet the logs calm down again.

  ```yaml
  apiVersion: kubeovn.io/v1
  kind: Subnet
  metadata:
    name: subnet-138
  spec:
    cidrBlock: 10.129.138.0/24,fd50:52:129:138::0/64
    default: false
    disableGatewayCheck: true
    vpc: test-vpc
  ```

3. Create a second subnet and bind it to the VPC. Check the `kube-ovn-controller` logs. Shortly after you created the subnet the logs calm down again.

  ```yaml
  apiVersion: kubeovn.io/v1
  kind: Subnet
  metadata:
    name: subnet-139
  spec:
    cidrBlock: 10.129.139.0/24,fd50:52:129:139::0/64
    default: false
    disableGatewayCheck: true
    vpc: test-vpc
  ```

4. Create a third subnet and bind it to the VPC.

  ```yaml
  apiVersion: kubeovn.io/v1
  kind: Subnet
  metadata:
    name: subnet-140
  spec:
    cidrBlock: 10.129.140.0/24,fd50:52:129:140::0/64
    default: false
    disableGatewayCheck: true
    vpc: test-vpc
  ```

5. Now watch the log output of the `kube-ovn-controller`. Notice that it takes ~3 minutes (on my machines) until the logs stop.

Check the [kubeovn-bug-report.log](https://github.com/kubeovn/kube-ovn/files/12910539/kubeovn-bug-report.log) I attached.
I annotated where I started to create the subnets.
You should note that adding `subnet-140` leads to an absurd amount of log lines.

# Additional Info

- Kubernetes version:

  ```
  Client Version: v1.28.2
  Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
  Server Version: v1.28.2
  ```

- kube-ovn version:
  
  ```
  v1.12.1
  ```

  **Output of `uname -r`:**
  ```
  Debian GNU/Linux 12 (bookworm)
  6.1.0-13-amd64
  ```

sirkrypt0:
We change the check to not enqueue a VPC update when the DeletionTimestamp is zero. Previously, the VPC update would basically always be enqueued, as the DeletionTimestamp would be zero until it was scheduled for deletion. This lead to a huge amount of enqueued VPC updates and therefore a lot of Kubernetes API requests and load on the API.

The logical not was removed in #3107 which therefore introduced this bug.

### What type of this PR

Bug Fix

### Which issue(s) this PR fixes:

Fixes #3301 

### <samp>🤖 Generated by Copilot at 01fc78b</samp>

Fix Vpc deletion bug by enqueuing update events when Vpc has deletion timestamp. Modify `pkg/controller/vpc.go` to implement this logic.

### <samp>🤖 Generated by Copilot at 01fc78b</samp>

> _`Vpc` deletion_
> _Enqueue update event now_
> _Fixes autumn bug_

### <samp>🤖 Generated by Copilot at 01fc78b</samp>

* Fix issue #1144 by enqueuing Vpc update events when the Vpc has a deletion timestamp ([link](https://github.com/kubeovn/kube-ovn/pull/3302/files?diff=unified&w=0#diff-ae8a8b5d44d277a4bb96635ee325fe54b48e8bc10b24a79d4bb8e211b57131ecL47-R47))

### DIFF
--- a/pkg/controller/vpc.go
+++ b/pkg/controller/vpc.go
@@ -44,7 +44,7 @@ func (c *Controller) enqueueUpdateVpc(oldObj, newObj interface{}) {
 	oldVpc := oldObj.(*kubeovnv1.Vpc)
 	newVpc := newObj.(*kubeovnv1.Vpc)
 
-	if newVpc.DeletionTimestamp.IsZero() ||
+	if !newVpc.DeletionTimestamp.IsZero() ||
 		!reflect.DeepEqual(oldVpc.Spec.Namespaces, newVpc.Spec.Namespaces) ||
 		!reflect.DeepEqual(oldVpc.Spec.StaticRoutes, newVpc.Spec.StaticRoutes) ||
 		!reflect.DeepEqual(oldVpc.Spec.PolicyRoutes, newVpc.Spec.PolicyRoutes) ||


pkg/controller/vpc.go
```diff
@@ -44,7 +44,7 @@ func (c *Controller) enqueueUpdateVpc(oldObj, newObj interface{}) {
 	oldVpc := oldObj.(*kubeovnv1.Vpc)
 	newVpc := newObj.(*kubeovnv1.Vpc)
 
-	if newVpc.DeletionTimestamp.IsZero() ||
+	if !newVpc.DeletionTimestamp.IsZero() ||
```
bobz965, line 47: maybe split the if !newVpc.DeletionTimestamp.IsZero() with the other is better.

if !newVpc.DeletionTimestamp.IsZero() {
// vpc is deleted
return nil

}

if ...(update while vpc is not deleted)

bobz965, line 47: How about just removing the !newVpc.DeletionTimestamp.IsZero(), I think the vpc `handleUpdate` has nothing to do with the delete logic.  all the clear logic should in the `handleDelVpc`

sirkrypt0: Looking around, [the code contains the same pattern in multiple places](https://github.com/search?q=repo%3Akubeovn%2Fkube-ovn%20.DeletionTimestamp.IsZero()&type=code), like

https://github.com/kubeovn/kube-ovn/blob/60ef1ad1f31d2974cb9462670b0ad26b1015be6f/pkg/controller/qos_policy.go#L55-L59

https://github.com/kubeovn/kube-ovn/blob/60ef1ad1f31d2974cb9462670b0ad26b1015be6f/pkg/controller/subnet.go#L70-L73

So there might be a reason for why this check exists? Although I currently tend to agree and don't really see a reason for why you'd want to enqueue an update, when the `DeletionTimestamp` has been set, i.e. the VPC is scheduled for deletion at some point later in time.

oilbeater: This pattern is aimed to prevent some race condition but I cannot recall the specific scenario. Maybe the controller receive an update event then  a deletion event, but it would process the deletion first then the update and lead to undesired garbage so we add a check before doing update. I think we can just keep the code like this as it can resolve this issue.

PR #3313, opened 2023-10-19
Nat reuse router port external ip

bobz965:

### What type of this PR

Features

- support ovn nat reuse lrp type eip
- support ovn nat specify IP or IP cidr
- ipam allocation should not return "nil" string, return ""(empty string) should be better
- update ovn nat e2e
- add independent vip e2e

limit gw nodes number to 2, in HA kind-init-ha env, we can have a normal worker node and two gw nodes.
use the ha env, we can tcpdump bfd packet between worker node and gw nodes in ovn `ha chassis group`  case in the future.

https://github.com/kubeovn/kube-ovn/issues/3287

### <samp>🤖 Generated by Copilot at ff9e3be</samp>

No summary available (Limit exceeded: required to process 52456 tokens, but only 50000 are allowed per call)

### <samp>🤖 Generated by Copilot at 3d06a9f</samp>

> _Sing, O Muse, of the valiant deeds of the ovn developers_
> _Who, with skill and wisdom, enhanced the vip and eip features_
> _They added tests and validations, and refined the error messages_
> _To make them clear and consistent, like the oracles of Delphi_

### <samp>🤖 Generated by Copilot at ff9e3be</samp>

No walkthrough available (Limit exceeded: required to process 52456 tokens, but only 50000 are allowed per call)


charts/templates/kube-ovn-crd.yaml
```diff
@@ -980,8 +980,6 @@ spec:
                   type: string
                 v4Ip:
                   type: string
-                macAddress:
```
oilbeater, line 983: 这个字段为什么要移除

cmd/cmdmain.go
```diff
@@ -47,11 +47,11 @@ func dumpProfile() {
 				klog.Errorf("failed to create cpu profile file: %v", err)
 				return
 			}
-			defer f.Close()
 			if err = pprof.StartCPUProfile(f); err != nil {
 				klog.Errorf("failed to start cpu profile: %v", err)
 				return
 			}
+			defer f.Close()
```
oilbeater, line 54: 为什么 defer 要往后移

bobz965: 这个是因为有个 G307 lint err，放在后面就不会有报错了

charts/templates/kube-ovn-crd.yaml
bobz965: fip 的 mac 只有在分布式场景下是有意义的，和 pod ip 的 mac 一致，而在集中式网关看到的snat出去的包都是 lrp 的 mac。

所以如果记录了，也可能存在集中式场景下的 mac 无意义的情况。 
由于 在  pod ip 那里可查 mac ，所以不再记录了

oilbeater (approved):

PR #3368, opened 2023-11-02
fix:  the ip repeated sometime when you change the subnet excludeIps

watermelon-brother:
## What type of this PR
Only support add reserved ip,because it will cause the ip repeated sometime when you add reserved ip and delete after sometime.

ip repeated problem

## WHAT
I highly recommend that you can deploy the webhook pod to verify the cr modify.

## HOW

Modify the subnet.V4CIDR or subnet.V4Gw is not recommended for use,because it will cause the same problem.

### DIFF
--- a/pkg/ipam/ipam.go
+++ b/pkg/ipam/ipam.go
@@ -193,89 +193,147 @@ func (ipam *IPAM) AddOrUpdateSubnet(name, cidrStr, gw string, excludeIps []strin
 			klog.Errorf("failed to parse v6 exclude ips %v", v6ExcludeIps)
 			return err
 		}
-		if (protocol == kubeovnv1.ProtocolDual || protocol == kubeovnv1.ProtocolIPv4) &&
-			(subnet.V4CIDR.String() != v4cidrStr || subnet.V4Gw != v4Gw || !subnet.V4Reserved.Equal(v4Reserved)) {
-			_, cidr, _ := net.ParseCIDR(v4cidrStr)
-			subnet.V4CIDR = cidr
-			subnet.V4Reserved = v4Reserved
-			firstIP, _ := util.FirstIP(v4cidrStr)
-			lastIP, _ := util.LastIP(v4cidrStr)
-			ips, _ := NewIPRangeListFrom(fmt.Sprintf("%s..%s", firstIP, lastIP))
-			subnet.V4Free = ips.Separate(subnet.V4Reserved)
-			subnet.V4Available = subnet.V4Free.Clone()
-			subnet.V4Using = subnet.V4Using.Intersect(ips)
-			subnet.V4Gw = v4Gw
+		if protocol == kubeovnv1.ProtocolDual || protocol == kubeovnv1.ProtocolIPv4 {
+			if subnet.V4CIDR.String() != v4cidrStr || subnet.V4Gw != v4Gw {
+				_, cidr, _ := net.ParseCIDR(v4cidrStr)
+				subnet.V4CIDR = cidr
+				subnet.V4Reserved = v4Reserved
+				firstIP, _ := util.FirstIP(v4cidrStr)
+				lastIP, _ := util.LastIP(v4cidrStr)
+				ips, _ := NewIPRangeListFrom(fmt.Sprintf("%s..%s", firstIP, lastIP))
+				subnet.V4Free = ips.Separate(subnet.V4Reserved)
+				subnet.V4Available = subnet.V4Free.Clone()
+				subnet.V4Using = subnet.V4Using.Intersect(ips)
+				subnet.V4Gw = v4Gw
 
-			pool := subnet.IPPools[""]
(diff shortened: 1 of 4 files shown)


pkg/util/validator.go
```diff
@@ -127,6 +127,18 @@ func ValidateSubnet(subnet kubeovnv1.Subnet) error {
 	return nil
 }
 
+func ValidateSubnetUpdate(subnet kubeovnv1.Subnet, oldSubnet kubeovnv1.Subnet) error {
+	err := ValidateSubnet(subnet)
+	if err != nil {
+		return err
+	}
+	// only support add excludeIps
```
bobz965, line 135: why only add ip into excludeIps ?

pkg/ipam/ipam.go
```diff
@@ -193,89 +193,143 @@ func (ipam *IPAM) AddOrUpdateSubnet(name, cidrStr, gw string, excludeIps []strin
...
-			for nicName, ip := range subnet.V4NicToIP {
-				if !ips.Contains(ip) {
-					podName := subnet.V4IPToPod[ip.String()]
-					klog.Errorf("%s address %s not in subnet %s new cidr %s", podName, ip.String(), name, cidrStr)
-					delete(subnet.V4NicToIP, nicName)
-					delete(subnet.V4IPToPod, ip.String())
-					continue
+			if !subnet.V4Reserved.Equal(v4Reserved) {
+				newAddV4Reserved := v4Reserved.Separate(subnet.V4Reserved)
+				// only support add reserved ip,because it will cause the ip repeated sometime when you add reserved ip and delete after sometime.
```
bobz965, line 241:
```suggestion
				// only support add reserved ip, because it will cause the ip repeated sometime when you add reserved ip and delete after sometime.
```

pkg/ipam/ipam.go
```diff
@@ -193,89 +193,143 @@ func (ipam *IPAM) AddOrUpdateSubnet(name, cidrStr, gw string, excludeIps []strin
...
-			for nicName, ip := range subnet.V6NicToIP {
-				if !ips.Contains(ip) {
-					podName := subnet.V6IPToPod[ip.String()]
-					klog.Errorf("%s address %s not in subnet %s new cidr %s", podName, ip.String(), name, cidrStr)
-					delete(subnet.V6NicToIP, nicName)
-					delete(subnet.V6IPToPod, ip.String())
-					continue
+			if !subnet.V6Reserved.Equal(v6Reserved) {
+				newAddV6Reserved := v6Reserved.Separate(subnet.V6Reserved)
+				// only support add reserved ip,because it will cause the ip repeated sometime when you add reserved ip and delete after sometime.
```
bobz965, line 312:
```suggestion
				// only support add reserved ip, because it will cause the ip repeated sometime when you add reserved ip and delete after sometime.
```

pkg/util/validator.go
watermelon-brother: If a vm use ip 1.1.1.1,you add 1.1.1.1 to excludeIps list.you delete the ip 1.1.1.1 from the excludeIps list at times.
If you do it ,you can find subnet.status.v4availableIPrange has 1.1.1.1 appear again，if a new vm use it , ip repeated problem is occurs.

bobz965: 一般来说，update 操作应该能够支持 add 和 delete 的更新。如果没能支持 delete，应该就是一个功能缺失的 bug。

watermelon-brother: Just think about what would happen if you support deletion.

pkg/util/validator.go
```diff
@@ -127,6 +127,18 @@ func ValidateSubnet(subnet kubeovnv1.Subnet) error {
 	return nil
 }
 
+func ValidateSubnetUpdate(subnet kubeovnv1.Subnet, oldSubnet kubeovnv1.Subnet) error {
+	err := ValidateSubnet(subnet)
+	if err != nil {
+		return err
+	}
```
zhangzujian, line 134:
```suggestion
	if err := ValidateSubnet(subnet); err != nil {
		return err
	}
```

pkg/util/validator.go
watermelon-brother: If you want delete  excludeIps, I suggest as below:
1. you must sure the ip is not vm used.
2. After you delete the excludeIps ip ,you can restart the ovn-controller.

zhangzujian: So there is a bug about excludeIPs update. We should fix it.

zhangzujian (changes requested):
We should fix the bug instead of disabling excludeIPs update/deletion.

oilbeater:
@watermelon-brother can you fix the conflict and the review comment

oilbeater:
We need to reconsider how to update ipam when excludeIPs changes, use https://github.com/kubeovn/kube-ovn/issues/3574 to follow up

PR #3439, opened 2023-11-17
fix: release VF assigned to the Pod back to the host network namespace

### LINKED ISSUE #3437: When the pod using the sriov interface is deleted, although the sriov interface has been returned to host network namespace, but the sriov interface has not been restored to its original name.
# Bug Report
when delete a pod using the sriov interface

## Expected Behavior
the sriov interface has been returned to host network namespace, 
and the sriov interface has been restored to its original name

## Actual Behavior
the sriov interface has been returned to host network namespace, and the sriov interface has not been restored to its original name

## Steps to Reproduce the Problem

1.kubectl create a pod using the sriov interface
2.kubectl delete the pod in step 1
3.On the node where the pod is located, use ip a to check the network port name and find that the network port name is the pod‘s interface name before like eth0, not the original network port name in the host network namespace of this node.

guof888:
# Pull Request

## What type of this PR

fix: release VF assigned to the Pod back to the host network namespace when delete pod

Fixes #3437 

## WHAT
Add a symmetric operation to configure the siriov interface (setupSriovInterface) for the pod: release the sriovs interface

## HOW
1.move VF to the host network namespace
2.renames the VF device from `eth0` to its original name


oilbeater:
@guof888 please add sign off in the commit to fix the DCO issues

PR #3748, opened 2024-02-20
fix: should use domain-qualified finalizer name

bobz965:
# Pull Request

## What type of this PR

- Bug fixes

image

Fixes #(issue-number)

### DIFF
--- a/pkg/controller/init.go
+++ b/pkg/controller/init.go
@@ -830,6 +830,7 @@ func (c *Controller) initNodeChassis() error {
 
 func (c *Controller) syncFinalizers() error {
 	// migrate depreciated finalizer to new finalizer
+	klog.Info("start to sync finalizers")
 	if err := c.syncIPFinalizer(); err != nil {
 		klog.Errorf("failed to sync ip finalizer: %v", err)
 		return err
@@ -878,5 +879,6 @@ func (c *Controller) syncFinalizers() error {
 		klog.Errorf("failed to sync iptables snat finalizer: %v", err)
 		return err
 	}
+	klog.Info("sync finalizers done")
 	return nil
 }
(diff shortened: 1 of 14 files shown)


zhangzujian:
We should migrate the finalizer field during initialization.

bobz965:
> We should migrate the finalizer field during initialization.

ok，

pkg/util/const.go
```diff
@@ -3,7 +3,8 @@ package util
 const (
 	CniTypeName = "kube-ovn"
 
-	ControllerName = "kube-ovn-controller"
+	DepreciatedFinalizerName = "kube-ovn-controller"
+	FinalizerName            = "kubeovn.io/finalizer"
```
zhangzujian, line 7:
```suggestion
	FinalizerName            = "kubeovn.io/kube-ovn-controller"
```

pkg/controller/init.go
```diff
@@ -882,3 +884,18 @@ func (c *Controller) syncFinalizers() error {
...
+		controllerutil.AddFinalizer(newObj, util.KubeOVNControllerFinalizer)
+		patch, err := util.GenerateMergePatchPayload(cachedObj, newObj)
+		if err != nil {
+			klog.Errorf("failed to generate patch payload for %s, %v", newObj.GetName(), err)
+			return nil, err
+		}
+		return patch, nil
+	}
+	return nil, nil
+}
```
zhangzujian, line 901: We can use `https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#Client` to simplify codes.

zhangzujian (approved):

pkg/controller/init.go
bobz965: i tried, but it is hard for now.

PR #4004, opened 2024-05-11
fix: Resolved the hidden issue with zombie processes

fanriming:
# Pull Request

## What type of this PR

- Bug fixes

When a container process restarts abnormally, the 1st process that is taken over by exec cannot be adopted, leading to zombie process issues.

### DIFF
--- a/dist/images/start-controller.sh
+++ b/dist/images/start-controller.sh
@@ -32,6 +32,6 @@ function gen_conn_str {
 nb_addr="$(gen_conn_str 6641)"
 sb_addr="$(gen_conn_str 6642)"
 
-exec ./kube-ovn-controller --ovn-nb-addr="$nb_addr" \
+./kube-ovn-controller --ovn-nb-addr="$nb_addr" \
                            --ovn-sb-addr="$sb_addr" \
                            $@
--- a/dist/images/start-ic-controller.sh
+++ b/dist/images/start-ic-controller.sh
@@ -55,6 +55,6 @@ if [ -z "${OVN_NB_DAEMON}" ]; then
   exit 1
 fi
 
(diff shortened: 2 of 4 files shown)


zhangzujian (approved):

zhangzujian:
Could you please attach details of the zombie processes?

fanriming:
> Could you please attach details of the zombie processes?

image

zhangzujian:
> Correction: The child process cannot be adopted。
> 
> image

What's the parent process of the `ovn-nbctl`? And how does the patch work?

`ovn-nbctl` has been replaced by libovsdb now. What version are you using?

fanriming:
> > Correction: The child process cannot be adopted。
> > image
> 
> What's the parent process of the `ovn-nbctl`? And how does the patch work?
> 
> `ovn-nbctl` has been replaced by libovsdb now. What version are you using?



> > Correction: The child process cannot be adopted。
> > image
> 
> What's the parent process of the `ovn-nbctl`? And how does the patch work?
> 
> `ovn-nbctl` has been replaced by libovsdb now. What version are you using?

The case in the screenshot may be special, and we are considering compatibility with older versions of the cluster, so ovn-nbctl and libovsdb are used together. I mainly want to express that when the golang program is the first process, the child process created by it will cause a zombie process after an abnormal exit. 
We found that if entrypoint is a golang executable (entrypoint executes commands from process 1 in the container), then the process does not have the ability of process 1 in the traditional operating system to recover resources from the orphan waitpid process and the orphan process when it ends. But bash does.

More details refer to the following link: 
https://segmentfault.com/a/1190000044567340
https://segmentfault.com/a/1190000044596111

PR #4127, opened 2024-06-05
add ovn0 default route

hongzhen-ma:
# Pull Request

## What type of this PR

1、dual-stack cluster
2、IPv6 route for ovn0 on node is deleted in some case, so loopOvn0Check failed and kube-ovn-cni pod crash
3、ovn0 nic has already exist and will not be created again. 
4、The route is added when pod change to Running state and after startWorkers. But before that, the ping check for gateway is failed since IPv6 route on node is deleted and has not been added.
5、So the route for ovn0 should be added before gateway check. 

- Features
- Bug fixes

Fixes #(issue-number)


oilbeater:
need an e2e test

oilbeater (approved):

PR #4140, opened 2024-06-06
speaker: add option to set node IP to override pod IP

### LINKED ISSUE #4139: [BUG] speaker: invalid next hop ip address when running pod without host network mode
### Kube-OVN Version

v1.12.16

### Kubernetes Version

v1.27.11

all kernels

### Description

When the speaker pod cannot run in host network mode, the default router ID and the source IP address of some routes is set to the pod IP, which is in a different subnet than the actual router IP address, causing invalid next hop IP address for those routes.

### Steps To Reproduce

- Run speaker with `hostNetwork: false`

### Current Behavior

Invalid next hop IP address for some routes which is the IP address of the pod 

### Expected Behavior

The IP address of the node should be set as the next hop for those routes

yaa110:
# Pull Request

## What type of this PR

- Features

Sometimes the speaker pod cannot run in host network mode, causing the default router ID and the source IP address of some routes to be set to the pod IP, which is in a different subnet than the actual router IP address. By introducing a new flag for the node IP, we can override the invalid router IP address.

Fixes #4139

### DIFF
--- a/pkg/speaker/subnet.go
+++ b/pkg/speaker/subnet.go
@@ -302,7 +302,7 @@ func (c *Controller) getNlriAndAttrs(route string) (*anypb.Any, [][]*anypb.Any, 
 			Origin: 0,
 		})
 		a2, _ := anypb.New(&bgpapi.NextHopAttribute{
-			NextHop: getNextHopAttribute(addr, c.config.RouterID),
+			NextHop: c.getNextHopAttribute(addr),
 		})
 		attrs = append(attrs, []*anypb.Any{a1, a2})
 	}
@@ -348,11 +348,17 @@ func getNextHopFromPathAttributes(attrs []bgp.PathAttributeInterface) net.IP {
 	return nil
 }
 
-func getNextHopAttribute(neighborAddress, routeID string) string {
-	nextHop := routeID
+func (c *Controller) getNextHopAttribute(neighborAddress string) string {
+	nextHop := c.config.RouterID
 	routes, err := netlink.RouteGet(net.ParseIP(neighborAddress))
 	if err == nil && len(routes) == 1 && routes[0].Src != nil {
 		nextHop = routes[0].Src.String()
+	}
+	proto := util.CheckProtocol(nextHop)
+	nextHopIP := net.ParseIP(nextHop)
+	nodeIP := c.config.NodeIPs[proto]
+	if nodeIP != nil && nextHopIP != nil && nextHopIP.Equal(c.config.PodIPs[proto]) {
+		nextHop = nodeIP.String()
 	}
 	return nextHop
 }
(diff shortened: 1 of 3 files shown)


pkg/speaker/subnet.go
```diff
@@ -348,11 +348,14 @@ func getNextHopFromPathAttributes(attrs []bgp.PathAttributeInterface) net.IP {
...
-	nextHop := routeID
+func (c *Controller) getNextHopAttribute(neighborAddress string) string {
+	nextHop := c.config.RouterID
 	routes, err := netlink.RouteGet(net.ParseIP(neighborAddress))
 	if err == nil && len(routes) == 1 && routes[0].Src != nil {
 		nextHop = routes[0].Src.String()
 	}
+	if c.config.NodeIP != "" && nextHop == c.config.PodIP {
+		nextHop = c.config.NodeIP
+	}
```
zhangzujian, line 359: We should consider deployment in duak-stack clusters.

yaa110: done

pkg/speaker/subnet.go
```diff
@@ -348,11 +348,18 @@ func getNextHopFromPathAttributes(attrs []bgp.PathAttributeInterface) net.IP {
...
 	if err == nil && len(routes) == 1 && routes[0].Src != nil {
 		nextHop = routes[0].Src.String()
 	}
+	proto := util.CheckProtocol(nextHop)
+	nextHopIP := net.ParseIP(nextHop)
+	nodeIP := c.config.NodeIPs[proto]
+	podIP := c.config.PodIPs[proto]
+	if nodeIP != nil && podIP != nil && nextHopIP != nil && nextHopIP.Equal(podIP) {
+		nextHop = nodeIP.String()
+	}
```
zhangzujian, line 363: I think we should use nodeIP as nextHtop if the nodeIP is not nil. No need to check podIP.

yaa110: done

pkg/speaker/subnet.go
```diff
@@ -348,11 +348,17 @@ func getNextHopFromPathAttributes(attrs []bgp.PathAttributeInterface) net.IP {
 	return nil
 }
 
-func getNextHopAttribute(neighborAddress, routeID string) string {
-	nextHop := routeID
+func (c *Controller) getNextHopAttribute(neighborAddress string) string {
```
zhangzujian, line 351: I mean this function should be like this:

```go
func (c *Controller) getNextHopAttribute(neighborAddress string) string {
	proto := util.CheckProtocol(neighborAddress)
	if nodeIP := c.config.NodeIPs[proto]; nodeIP != nil {
		return nodeIP.String()
	}

	nextHop := c.config.RouterID
	routes, err := netlink.RouteGet(net.ParseIP(neighborAddress))
	if err == nil && len(routes) == 1 && routes[0].Src != nil {
		nextHop = routes[0].Src.String()
	}
	return nextHop
}
```

yaa110: what if the route has a different source from the podIP? I meant, we should only use nodeIP when the nextHop equals podIP.

Issue #4217, opened 2024-06-24
[Feature Request] Announce NAT-GW EIP over BGP


SkalaNetworks:
### Description

When adding EIPs from an "external subnet", the act of receiving the traffic for our EIPs is ensured by ARP on the L2 segment between the "external subnet" GW and the NAT-GW from Kube-OVN.
A more suited way of announcing those IPs would be BGP. This is especially useful in highly flexible environments like the ones of cloud providers.
This would be analog to the already present feature of announcing Pod IPs/Subnet IPs to a BGP speaker, except the NAT-GW would dynamically announce the EIPs everytime one is added to it.

### Who will benefit from this feature?

People in need of flexibility, where routing to a GW isn't always the best way to do things (the GW needs to have an IP in the subnet, potentially wasting a public IP for public ranges..)

SkalaNetworks:
The way I see it, it could not work because of this:
- New EIP added to a NAT-GW
- The speakers start announcing this IP, the traffic for those EIPs starts routing to the nodes hosting the speakers
- The traffic needs to head to the NAT-GW now, the problem is the NAT-GW only has one IP from the point of view of Kubernetes, even though it might have many more because the EIPs are added directly to the eth0 interface in the Pod
- The node doesn't know where to route that traffic because the IP isn't visible from K8s point of view

Other way of doing it,
- Speakers announce every EIP with nexthop set to the NAT-GW
- NAT-GW needs to be on the same L2 segment as the BGP router the speakers are peering to

Yet another way
- Speakers announce every EIP with next set to NAT-GW
- Speakers announce the IP (used as nexthop for EIPs) of the NAT-GW
- BGP router peered to the speakers has recursive nexhop activated and knows that EIP 10.X.Y.Z has nexthop A.B.C.D and A.B.C.D is reachable by using one of the nodes hosting the speakers as nexthop

This is obviously less optimal but whatever.

What do you guys think?

bobz965:
I think ovn eip is a better way.

zhangzujian:
We can add a speaker process in the vpc nat gw pod and announce eip(s) to the "external subnet" GW.

SkalaNetworks:
That sounds like a good idea, a few questions on my part:
- EIPs need to be part of the "external subnet", correct? If yes, that would be limiting. Ideally EIPs could be from multiple subnets of "public ranges", and then announced to the GW of the external subnet through BGP.
[...]

PR #4285, opened 2024-07-11
Add BGP capabilities to the NAT GW

SkalaNetworks:
# Pull Request

## What type of this PR

- Introduces a flag to the BGP speaker to turn it into a "NAT GW mode", this makes the speaker announce EIPs linked to the GW the speaker is currently in. This disables announcement of Pod IPs/Services/Subnets by the speaker. Note: this mode can only be set inside a NAT GW as the pod will search for a specific new env variable in the GW to determine in which one it is running.
- Introduces a new container to the NAT GW if the experimental BGP option is set in the NAT GW configmap, this container contains a BGP speaker that will announce EIPs added to the GW if they have the correct annotation

Fixes [4217](https://github.com/kubeovn/kube-ovn/issues/4217)

This PR is highly experimental, do not merge in master.

To test:

We need to have a NAD for the interface that the speaker will use to talk to the apiserver.
For that we follow the exact same pattern as for the vpc-dns

Create a NAD with access to kube-ovn default network:
```yaml
apiVersion: "k8s.cni.cncf.io/v1"
kind: NetworkAttachmentDefinition
metadata:
  name: ovn-nad
  namespace: default
spec:
  config: '{
      "cniVersion": "0.3.0",
      "type": "kube-ovn",
      "server_socket": "/run/openvswitch/kube-ovn-daemon.sock",
      "provider": "ovn-nad.default.ovn"
    }'
```

Edit the ovn-default subnet to use the NAD as a provider.

Create new permissions for the gateway to poll the apiserver:
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    kubernetes.io/bootstrapping: rbac-defaults
  name: system:vpc-nat-gw
rules:
  - apiGroups:
    - ""
    resources:
    - services
    - pods
    verbs:
    - list
    - watch
  - apiGroups:
    - kubeovn.io
    resources:
    - iptables-eips
    - subnets
    - vpc-nat-gateways
    verbs:
    - list
    - watch
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  annotations:
    rbac.authorization.kubernetes.io/autoupdate: "true"
  labels:
    kubernetes.io/bootstrapping: rbac-defaults
  name: vpc-nat-gw
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: system:vpc-nat-gw
subjects:
- kind: ServiceAccount
  name: vpc-nat-gw
  namespace: kube-system
---
apiVersion: v1
kind: ServiceAccount
metadata:
  name: vpc-nat-gw
  namespace: kube-system
```

Launch a NAT GW and mark its EIP with the BGP annotation:
`ovn.kubernetes.io/bgp: "true"`

### DIFF
--- a/charts/kube-ovn/templates/kube-ovn-crd.yaml
+++ b/charts/kube-ovn/templates/kube-ovn-crd.yaml
@@ -503,6 +503,15 @@ spec:
                     type: string
                 qosPolicy:
                   type: string
+                bgpSpeaker:
+                  type: object
+                  properties:
+                    enabled:
+                      type: boolean
+                    parameters:
+                      type: array
+                      items:
+                        type: string
                 tolerations:
                   type: array
                   items:
--- a/charts/kube-ovn/templates/kube-ovn-crd.yaml
+++ b/charts/kube-ovn/templates/kube-ovn-crd.yaml
@@ -508,7 +508,23 @@ spec:
                   properties:
                     enabled:
                       type: boolean
-                    parameters:
+                    asn:
+                      type: integer
+                    remoteAsn:
+                      type: integer
+                    neighbors:
+                      type: array
+                      items:
+                        type: string
+                    holdTime:
+                      type: string
+                    routerId:
+                      type: string
+                    password:
+                      type: string
+                    enableGracefulRestart:
+                      type: boolean
+                    extraArgs:
                       type: array
                       items:
                         type: string
--- a/pkg/controller/vpc_nat.go
+++ b/pkg/controller/vpc_nat.go
@@ -11,6 +11,7 @@ import (
 var (
 	vpcNatImage             = ""
 	vpcNatGwBgpSpeakerImage = ""
+	vpcNatAPINadName        = ""
 	vpcNatAPINadProvider    = ""
 )
 
@@ -34,6 +35,9 @@ func (c *Controller) resyncVpcNatConfig() {
 	// Image for the BGP sidecar of the gateway (optional)
 	vpcNatGwBgpSpeakerImage = cm.Data["bgpSpeakerImage"]
 
-	// NetworkAttachmentDefinition for the BGP speaker to call the API server
+	// NetworkAttachmentDefinition name for the BGP speaker to call the API server
+	vpcNatAPINadName = cm.Data["apiNadName"]
+
+	// NetworkAttachmentDefinition provider for the BGP speaker to call the API server
 	vpcNatAPINadProvider = cm.Data["apiNadProvider"]
 }
--- a/pkg/controller/vpc_nat.go
+++ b/pkg/controller/vpc_nat.go
@@ -29,7 +29,6 @@ func (c *Controller) resyncVpcNatImage() {
 
 	enableBgpSpeaker, exist := cm.Data["enableBgpSpeaker"]
 	if exist && enableBgpSpeaker == "true" {
-		klog.V(5).Infof("experimental BGP speaker enabled")
 		vpcNatEnableBgpSpeaker = true
 	}
 }
--- a/pkg/controller/vpc_nat.go
+++ b/pkg/controller/vpc_nat.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"fmt"
+
 	"github.com/kubeovn/kube-ovn/pkg/util"
 	"k8s.io/klog/v2"
 )
--- a/pkg/controller/vpc_nat.go
+++ b/pkg/controller/vpc_nat.go
@@ -11,7 +11,7 @@ import (
 var (
 	vpcNatImage             = ""
 	vpcNatGwBgpSpeakerImage = ""
-	vpcNatApiNadProvider    = ""
+	vpcNatAPINadProvider    = ""
 )
 
 func (c *Controller) resyncVpcNatConfig() {
@@ -35,5 +35,5 @@ func (c *Controller) resyncVpcNatConfig() {
 	vpcNatGwBgpSpeakerImage = cm.Data["bgpSpeakerImage"]
 
 	// NetworkAttachmentDefinition for the BGP speaker to call the API server
-	vpcNatApiNadProvider = cm.Data["apiNadProvider"]
+	vpcNatAPINadProvider = cm.Data["apiNadProvider"]
 }
--- a/pkg/controller/vpc_nat.go
+++ b/pkg/controller/vpc_nat.go
@@ -9,9 +9,8 @@ import (
 )
 
 var (
-	vpcNatImage              = ""
-	vpcNatGwEnableBgpSpeaker = false
-	vpcNatGwBgpSpeakerImage  = ""
+	vpcNatImage             = ""
+	vpcNatGwBgpSpeakerImage = ""
 )
 
 func (c *Controller) resyncVpcNatImage() {
@@ -30,16 +29,5 @@ func (c *Controller) resyncVpcNatImage() {
 	}
 	vpcNatImage = image
 
-	// Check BGP is enabled on the NAT GW, if yes, verify required parameters are present
-	enableBgpSpeaker, exist := cm.Data["enableBgpSpeaker"]
-	if exist && enableBgpSpeaker == "true" {
-		vpcNatGwEnableBgpSpeaker = true
-
-		vpcNatGwBgpSpeakerImage, exist = cm.Data["bgpSpeakerImage"]
-		if !exist {
-			err = fmt.Errorf("%s should have bgp speaker image field if bgp enabled", util.VpcNatConfig)
-			klog.Error(err)
-			return
-		}
-	}
+	vpcNatGwBgpSpeakerImage = cm.Data["bgpSpeakerImage"]
 }
--- a/pkg/controller/vpc_nat.go
+++ b/pkg/controller/vpc_nat.go
@@ -11,9 +11,10 @@ import (
 var (
 	vpcNatImage             = ""
 	vpcNatGwBgpSpeakerImage = ""
+	vpcNatApiNadProvider    = ""
 )
 
-func (c *Controller) resyncVpcNatImage() {
+func (c *Controller) resyncVpcNatConfig() {
 	cm, err := c.configMapsLister.ConfigMaps(c.config.PodNamespace).Get(util.VpcNatConfig)
 	if err != nil {
 		err = fmt.Errorf("failed to get ovn-vpc-nat-config, %w", err)
@@ -21,6 +22,7 @@ func (c *Controller) resyncVpcNatImage() {
 		return
 	}
 
+	// Image we're using to provision the NAT gateways
 	image, exist := cm.Data["image"]
 	if !exist {
 		err = fmt.Errorf("%s should have image field", util.VpcNatConfig)
@@ -29,5 +31,9 @@ func (c *Controller) resyncVpcNatImage() {
 	}
 	vpcNatImage = image
 
+	// Image for the BGP sidecar of the gateway (optional)
 	vpcNatGwBgpSpeakerImage = cm.Data["bgpSpeakerImage"]
+
+	// NetworkAttachmentDefinition for the BGP speaker to call the API server
+	vpcNatApiNadProvider = cm.Data["apiNadProvider"]
 }
--- a/pkg/controller/vpc_nat.go
+++ b/pkg/controller/vpc_nat.go
@@ -2,13 +2,14 @@ package controller
 
 import (
 	"fmt"
-
-	"k8s.io/klog/v2"
-
 	"github.com/kubeovn/kube-ovn/pkg/util"
+	"k8s.io/klog/v2"
 )
 
-var vpcNatImage = ""
+var (
+	vpcNatImage            = ""
+	vpcNatEnableBgpSpeaker = false
+)
 
 func (c *Controller) resyncVpcNatImage() {
 	cm, err := c.configMapsLister.ConfigMaps(c.config.PodNamespace).Get(util.VpcNatConfig)
@@ -17,6 +18,7 @@ func (c *Controller) resyncVpcNatImage() {
 		klog.Error(err)
 		return
 	}
+
 	image, exist := cm.Data["image"]
 	if !exist {
 		err = fmt.Errorf("%s should have image field", util.VpcNatConfig)
@@ -24,4 +26,10 @@ func (c *Controller) resyncVpcNatImage() {
 		return
 	}
 	vpcNatImage = image
+
+	enableBgpSpeaker, exist := cm.Data["enableBgpSpeaker"]
+	if exist && enableBgpSpeaker == "true" {
+		klog.V(5).Infof("experimental BGP speaker enabled")
+		vpcNatEnableBgpSpeaker = true
+	}
 }
--- a/pkg/controller/vpc_nat.go
+++ b/pkg/controller/vpc_nat.go
@@ -7,8 +7,9 @@ import (
 )
 
 var (
-	vpcNatImage            = ""
-	vpcNatEnableBgpSpeaker = false
+	vpcNatImage              = ""
+	vpcNatGwEnableBgpSpeaker = false
+	vpcNatGwBgpSpeakerImage  = ""
 )
 
 func (c *Controller) resyncVpcNatImage() {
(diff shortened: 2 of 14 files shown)


SkalaNetworks:
This is in progress, I'll fix conflicts in a few hours, but you can at least see how I got the thing to work for now

zhangzujian:
Great job! 🎉

zhangzujian:
Please add Signed-off-by in the commit message and rebase the master branch.


```txt
<COMMIT MESSAGE>

Signed-off-by: Author Name <authoremail@example.com>
```

SkalaNetworks:
Yeah I've done that but not pushed it yet, i'm seeing an awful lot of messages such as
"E0712 04:50:36.827065       7 vpc_nat_gateway.go:719] failed to ExecuteCommandInContainer, errOutput: Error: any valid address is expected rather than "dev".
failed to exec command "ip route replace default via  dev net1"

in the controller logs, idk if my code broke something or not ever since i rebased the changes you've done yesterday on the nat gateway or if it's something else, I need to investigate

SkalaNetworks:
Ah, you modified a bunch of stuff in the shell script of the nat-gateway image and the pullPolicy is IfNotPresent, I had an old version running. This is now fixed, everything seems to be working.

I'll add a few lines to make the image used as a speaker a parameter, and find a way to configure the BGP neighbors. Do you think we should make this a parameter of the VpcNatGateway CRD or just something adjusted from the nat-gw configmap ? @zhangzujian

zhangzujian:
CRD would be better.

SkalaNetworks:
Added a CRD to enable the bgp speaker per gw and to override the parameters sent to kube-ovn-speakers.

pkg/speaker/bgp.go
```diff
@@ -0,0 +1,242 @@
...
+
+	kubeovnv1 "github.com/kubeovn/kube-ovn/pkg/apis/kubeovn/v1"
+	"github.com/kubeovn/kube-ovn/pkg/util"
+	bgpapi "github.com/osrg/gobgp/v3/api"
+	bgpapiutil "github.com/osrg/gobgp/v3/pkg/apiutil"
+	"github.com/osrg/gobgp/v3/pkg/packet/bgp"
+	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
+	"google.golang.org/protobuf/types/known/anypb"
+	"k8s.io/klog/v2"
```
zhangzujian, line 16:
```suggestion
	bgpapi "github.com/osrg/gobgp/v3/api"
	bgpapiutil "github.com/osrg/gobgp/v3/pkg/apiutil"
	"github.com/osrg/gobgp/v3/pkg/packet/bgp"
	"github.com/vishvananda/netlink"
	"golang.org/x/sys/unix"
	"google.golang.org/protobuf/types/known/anypb"
	"k8s.io/klog/v2"

	kubeovnv1 "github.com/kubeovn/kube-ovn/pkg/apis/kubeovn/v1"
	"github.com/kubeovn/kube-ovn/pkg/util"
```

pkg/speaker/bgp.go
```diff
@@ -0,0 +1,242 @@
...
+)
+
+var (
+	maskMap = map[string]int{kubeovnv1.ProtocolIPv4: 32, kubeovnv1.ProtocolIPv6: 128}
+)
```
zhangzujian, line 21: Please run `make gofumpt` to format the files.

pkg/speaker/controller.go
```diff
@@ -3,6 +3,8 @@ package speaker
 import (
 	"time"
 
+	"github.com/kubeovn/kube-ovn/pkg/util"
```
zhangzujian, line 6: This import should be put together with other internal imports.

pkg/speaker/eip.go
```diff
@@ -0,0 +1,54 @@
+package speaker
+
+import (
+	"fmt"
+
+	v1 "github.com/kubeovn/kube-ovn/pkg/apis/kubeovn/v1"
+	"github.com/kubeovn/kube-ovn/pkg/util"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
```
zhangzujian, line 9:
```suggestion
	"k8s.io/apimachinery/pkg/labels"
	"k8s.io/apimachinery/pkg/selection"

	v1 "github.com/kubeovn/kube-ovn/pkg/apis/kubeovn/v1"
	"github.com/kubeovn/kube-ovn/pkg/util"
```

pkg/speaker/utils.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	"fmt"
+	"os"
+	"strconv"
+	"strings"
+
+	kubeovnv1 "github.com/kubeovn/kube-ovn/pkg/apis/kubeovn/v1"
+	"github.com/kubeovn/kube-ovn/pkg/util"
+	bgpapi "github.com/osrg/gobgp/v3/api"
+	v1 "k8s.io/api/core/v1"
```
zhangzujian, line 12:
```suggestion
	bgpapi "github.com/osrg/gobgp/v3/api"
	v1 "k8s.io/api/core/v1"

	kubeovnv1 "github.com/kubeovn/kube-ovn/pkg/apis/kubeovn/v1"
	"github.com/kubeovn/kube-ovn/pkg/util"
```

pkg/speaker/utils.go
```diff
@@ -0,0 +1,110 @@
...
+// parseRoute returns the prefix and length of the prefix (in bits) by parsing the received route
+// If no prefix is mentioned in the route (e.g 1.1.1.1 instead of 1.1.1.1/32), the prefix length
+// is assumed to be 32 bits
+func parseRoute(route string) (string, uint32, error) {
+	var prefixLen uint32 = 32
+	prefix := route
+	if strings.Contains(route, "/") {
+		prefix = strings.Split(route, "/")[0]
+		strLen := strings.Split(route, "/")[1]
+		intLen, err := strconv.ParseInt(strLen, 10, 32)
```
zhangzujian, line 83:
```suggestion
		intLen, err := strconv.ParseUint(strLen, 10, 32)
```

SkalaNetworks:
Hi, thanks for the review, I'll make the commits.
Also linking this PR: [https://github.com/kubeovn/docs/pull/181](), as gofumpt is not installed by default

pkg/controller/vpc_nat.go
```diff
@@ -3,12 +3,15 @@ package controller
 import (
 	"fmt"
 
-	"k8s.io/klog/v2"
-
 	"github.com/kubeovn/kube-ovn/pkg/util"
+	"k8s.io/klog/v2"
```
zhangzujian, line 7: Internal imports should be put at the end. Please revert this change.

SkalaNetworks:
```go
import (
	"fmt"

	"k8s.io/klog/v2"

	"github.com/kubeovn/kube-ovn/pkg/util"
)
```

Like this? My IDE is configured in goimports "package" mode just like your linter seems to be, but it doesn't order imports like that.

charts/kube-ovn/templates/kube-ovn-crd.yaml
```diff
@@ -503,6 +503,15 @@ spec:
...
                   type: string
+                bgpSpeaker:
+                  type: object
+                  properties:
+                    enabled:
+                      type: boolean
+                    parameters:
+                      type: array
+                      items:
+                        type: string
```
zhangzujian, line 514: How about split the `parameters` property into multiple fields for easier use? We can refer to [BGPPeer configuration](https://github.com/metallb/metallb/blob/a207a6e5a26dfa10d8552f099d3eb93eaa095fec/api/v1beta2/bgppeer_types.go#L25) used by MetalLB.

pkg/controller/vpc_nat.go
```diff
@@ -17,11 +21,25 @@ func (c *Controller) resyncVpcNatImage() {
...
 		err = fmt.Errorf("%s should have image field", util.VpcNatConfig)
 		klog.Error(err)
 		return
 	}
 	vpcNatImage = image
+
+	// Check BGP is enabled on the NAT GW, if yes, verify required parameters are present
+	enableBgpSpeaker, exist := cm.Data["enableBgpSpeaker"]
+	if exist && enableBgpSpeaker == "true" {
+		vpcNatGwEnableBgpSpeaker = true
```
zhangzujian, line 36: I think we should move this switch into the vpc nat gateway CRD.

SkalaNetworks: it's mainly here as an ultimate failswitch because this is experimental, I guess we can get rid of it, the code seems safe enough to not accidentally inject the speaker sidecar

charts/kube-ovn/templates/kube-ovn-crd.yaml
SkalaNetworks: Sure, I'll extract the parameters the speaker already supports from the metallb config

SkalaNetworks: done, i still left the extraArgs just in case

pkg/controller/vpc_nat.go
SkalaNetworks: I removed it entirely

SkalaNetworks:
@zhangzujian Hi, PR is now fully green!

zhangzujian:
```txt
E0720 08:07:53.667742      13 vpc_nat_gateway.go:294] failed to create statefulset 'vpc-nat-gw-gw1', err: StatefulSet.apps "vpc-nat-gw-gw1" is invalid: spec.template.annotations: Invalid value: ".kubernetes.io/routes": prefix part a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

Please fix it.

SkalaNetworks:
The error is caused by your configuration lacking the NAD Provider (same setup as the one needed by the VPC DNS), I was relying on the same configmap as the VPC-DNS one by mistake. Now the VPC NATGW configmap has a field to provide a networkattachementdefinition so that the speaker can actually talk with the API server.

See the PR's first message for the complete setup of the NAD.

You need the field `apiNadProvider: ovn-nad.default.ovn` in your ovn-vpc-nat-config configmap (or any other value that points to a NAD connected to the default subnet)

I added an error message to make the problem more explicit.

SkalaNetworks:
I fixed the conflict introduced by the latest commit on main

SkalaNetworks:
Well apparently there's also a new linter now, so I need to do some changes on the fmt part..

zhangzujian:
I'm kind of busy recently. Will review the commits later.

zhangzujian:
```yaml
k8s.v1.cni.cncf.io/networks: kube-system/ovn-vpc-external-network, default/
```

Still fails. Please fix it.

pkg/controller/vpc_nat_gateway.go
```diff
@@ -747,6 +790,18 @@ func (c *Controller) genNatGwStatefulSet(gw *kubeovnv1.VpcNatGateway, oldSts *v1
 		util.LogicalSwitchAnnotation:     gw.Spec.Subnet,
 		util.IPAddressAnnotation:         gw.Spec.LanIP,
 	}
+
+	if gw.Spec.BgpSpeaker.Enabled { // Add an interface that can reach the API server
+		defaultSubnet, err := c.subnetsLister.Get(c.config.DefaultLogicalSwitch)
```
zhangzujian, line 795: Since we have introduced configMap option `apiNadProvider`, we should use the nad instead of hard-coded default subnet.

zhangzujian: We should use function `findSubnetByNetworkAttachmentDefinition` instead of using the default subnet directly.

SkalaNetworks: The Kubernetes apiserver runs in the default subnet doesn't it? The NAD just happens to be connected to that subnet. Do you want me to:
- Lookup every subnet in the cluster
- Determine which one has a provider equal to our NAD
- Get its gateway

That can be an option

zhangzujian: > The Kubernetes apiserver runs in the default subnet doesn't it?

K8s apiserver runs in control plane nodes with host network. It can be accessed from subnets in the default vpc.

zhangzujian: The NAD can be pointed to ANY subnet which is running in the default vpc.

SkalaNetworks: Cool, should kube-ovn have a default NAD installed in the default VPC for kube-dns and the NAT GW? That would be extremely handy.

zhangzujian: Here is an example I'm using:

```yaml
apiVersion: "k8s.cni.cncf.io/v1"
kind: NetworkAttachmentDefinition
metadata:
  name: ovn-nad
  namespace: default
spec:
  config: '{
      "cniVersion": "0.3.0",
      "type": "kube-ovn",
      "server_socket": "/run/openvswitch/kube-ovn-daemon.sock",
      "provider": "ovn-nad.default.ovn"
    }'
---
apiVersion: kubeovn.io/v1
kind: Subnet
metadata:
  name: vpc-apiserver-subnet
spec:
  protocol: IPv4
  cidrBlock: 100.100.100.0/24
  provider: ovn-nad.default.ovn
```

zhangzujian: > should kube-ovn have a default NAD installed in the default VPC for kube-dns and the NAT GW?

NAD resources can be created only when the CRD is installed in the cluster, while the CRD should be installed by users.

SkalaNetworks:
> ```yaml
> k8s.v1.cni.cncf.io/networks: kube-system/ovn-vpc-external-network, default/
> ```
> 
> Still fails. Please fix it.

New ConfigMap option:

```yaml
apiVersion: v1
data:
  apiNadName: ovn-nad
  apiNadProvider: ovn-nad.default.ovn
  bgpSpeakerImage: docker.io/kubeovn/vpc-nat-gateway:v1.13.0
  image: docker.io/kubeovn/vpc-nat-gateway:v1.13.0
kind: ConfigMap
metadata:
  name: ovn-vpc-nat-config
  namespace: kube-system
```

This should fix it.
I don't really understand why I need to provide the "networks" by their NAD name, but configuration for routes etc... with the full provider name. Is there a specific reason.

pkg/controller/vpc_nat_gateway.go
```diff
@@ -735,6 +736,52 @@ func (c *Controller) execNatGwRules(pod *corev1.Pod, operation string, rules []s
 	return nil
 }
 
+func (c *Controller) setNatGwInterface(annotations map[string]string, externalNetwork string, defaultSubnet *kubeovnv1.Subnet) error {
+	if vpcNatAPINadName == "" {
+		return errors.New("no NetworkAttachmentDefinition provided to access apiserver, check configmap ovn-vpc-nat-config and field 'apiNadName'")
+	}
+
+	nad := fmt.Sprintf("%s/%s, %s/%s", c.config.PodNamespace, externalNetwork, corev1.NamespaceDefault, vpcNatAPINadName)
```
zhangzujian, line 744: If the NAD namspace is always `default`, there is no need to set both `apiNadName` and `apiNadProvider` in the configMap since `<apiNadProvider>` equals `<apiNadName>.default.ovn` or `<apiNadName>.default`.

How about getting NAD name and namespace from the NAD provider?

zhangzujian:
>I don't really understand why I need to provide the "networks" by their NAD name, but configuration for routes etc... with the full provider name. Is there a specific reason.

I am not very clear what you mean. What's the best implement in your mind?

SkalaNetworks:
> > I don't really understand why I need to provide the "networks" by their NAD name, but configuration for routes etc... with the full provider name. Is there a specific reason.
> 
> I am not very clear what you mean. What's the best implement in your mind?

I probably don't understand what providers are from the point of view of Kube-OVN.
To me, they basically symbolize a subnet attached to a NAD (so, to a specific interface in the Pod). The Kube-OVN API reference doc says the value of the "provider" field in the subnet field makes Kube-OVN find the NetworkAttachementDefinition.

```
Default value is ovn. In the case of multiple NICs, the value is <name>.<namespace> of the NetworkAttachmentDefinition, Kube-OVN will use this information to find the corresponding subnet resource
```

But my NAD has a provider called "ovn-nad.default.ovn" not ovn-nad.default, which makes me think Kube-OVN actually finds the NAD based on the name of the provider, not by decomposing it into name + namespace? Or am I mistaken?

SkalaNetworks:
Ok, seems like the "dot ovn" in the provider is a special case handled by "findSubnetByNetworkAttachmentDefinition"

zhangzujian:
> But my NAD has a provider called "ovn-nad.default.ovn" not ovn-nad.default, which makes me think Kube-OVN actually finds the NAD based on the name of the provider, not by decomposing it into name + namespace? Or am I mistaken?

NAD with a suffix `.ovn` means both datapath and IPAM are provided by Kube-OVN. NAD without that suffix means the subnet is used only to provide IPAM for other CNI plugins.

zhangzujian (approved):

oilbeater:
@SkalaNetworks can you add a doc to describe this feature?

SkalaNetworks:
I'll be doing it ASAP, I'm also adding the feature mentionned by @zhangzujian to make the NAD name/provider one parameter.

SkalaNetworks:
Related PR: https://github.com/kubeovn/kube-ovn/pull/4352

PR #4326, opened 2024-07-22
security: run as non-root user

zhangzujian:
# Pull Request

- [x] Make sure you have followed [Kube-OVN Code Style](https://github.com/kubeovn/kube-ovn/blob/master/CODE_STYLE.md).

## What type of this PR

Examples of user facing changes:

- Features
- Bug fixes
- Docs
- Tests

## Which issue(s) this PR fixes

Fixes #(issue-number)

PR #4363, opened 2024-08-05
Fix DISABLE_MODULES_MANAGEMENT

kvaps:
There is a regression introduced by https://github.com/kubeovn/kube-ovn/pull/4326

Let's move this logic into `start-ovn.sh` script

### DIFF
--- a/charts/kube-ovn/templates/ovncni-ds.yaml
+++ b/charts/kube-ovn/templates/ovncni-ds.yaml
@@ -126,7 +126,9 @@ spec:
               - NET_BIND_SERVICE
               - NET_RAW
               - SYS_ADMIN
+              {{- if not .Values.DISABLE_MODULES_MANAGEMENT }}
               - SYS_MODULE
+              {{- end }}
               - SYS_NICE
         env:
           - name: ENABLE_SSL
--- a/charts/kube-ovn/templates/ovsovn-ds.yaml
+++ b/charts/kube-ovn/templates/ovsovn-ds.yaml
@@ -81,18 +81,7 @@ spec:
           {{- if .Values.DPDK }}
           command: ["/kube-ovn/start-ovs-dpdk.sh"]
           {{- else }}
-          command:
-          {{- if .Values.DISABLE_MODULES_MANAGEMENT }}
-          - /bin/sh
-          - -ec
-          - |
-              ln -sf /bin/true /usr/sbin/modprobe
-              ln -sf /bin/true /usr/sbin/modinfo
-              ln -sf /bin/true /usr/sbin/rmmod
-              exec /kube-ovn/start-ovs.sh
-          {{- else }}
-          - /kube-ovn/start-ovs.sh
-          {{- end }}
+          command: ["/kube-ovn/start-ovs.sh"]
           {{- end }}
           securityContext:
             runAsUser: 65534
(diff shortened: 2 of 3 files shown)


charts/kube-ovn/templates/ovncni-ds.yaml
```diff
@@ -126,7 +126,9 @@ spec:
               - NET_BIND_SERVICE
               - NET_RAW
               - SYS_ADMIN
+              {{- if not .Values.DISABLE_MODULES_MANAGEMENT }}
               - SYS_MODULE
+              {{- end }}
```
zhangzujian, line 131: kube-ovn-cni will fail to start if we drop capability SYS_MODULE.

charts/kube-ovn/templates/ovsovn-ds.yaml
```diff
@@ -81,18 +81,7 @@ spec:
...
-              ln -sf /bin/true /usr/sbin/modinfo
-              ln -sf /bin/true /usr/sbin/rmmod
-              exec /kube-ovn/start-ovs.sh
-          {{- else }}
-          - /kube-ovn/start-ovs.sh
-          {{- end }}
+          command: ["/kube-ovn/start-ovs.sh"]
           {{- end }}
           securityContext:
             runAsUser: 65534
```
zhangzujian, line 87: I think the best way to fix the problem is to set `runAsUser=0` if `DISABLE_MODULES_MANAGEMENT` is `true`.

charts/kube-ovn/templates/ovsovn-ds.yaml
```diff
@@ -81,18 +81,7 @@ spec:
           {{- if .Values.DPDK }}
           command: ["/kube-ovn/start-ovs-dpdk.sh"]
           {{- else }}
-          command:
-          {{- if .Values.DISABLE_MODULES_MANAGEMENT }}
```
zhangzujian, line 85: Why we introduced `DISABLE_MODULES_MANAGEMENT`? Is it related to https://github.com/kubeovn/kube-ovn/issues/4286？

kvaps:
Closed in favor https://github.com/kubeovn/kube-ovn/pull/4363

charts/kube-ovn/templates/ovncni-ds.yaml
kvaps: Yeah you're right, this makes ovn totally incompatible with Talos Linux, because it is not allowed managing system modules by containers.

https://www.talos.dev/v1.7/learn-more/process-capabilities/

charts/kube-ovn/templates/ovsovn-ds.yaml
kvaps: Because it is the only way for running kube-ovn on Talos Linux
https://github.com/kubeovn/kube-ovn/blob/ed16ce5a2b8fb4288c88d87fefb174530796a007/charts/kube-ovn/README.md?plain=1#L24

cybercoder: I have deployed my simple ovs daemonset inside talos and works fine. The problem is : ```modinfo``` is lying because it fetches the list of known modules by reading the ```/lib/modules/$(uname -r)/modules.*``` files, which are usually updated with ```depmod```.

```depmod -a``` can't run in talos and modinfo does not knows about it. This does not prevent us from loading the module.

PR #4425, opened 2024-08-21
fix: support ptp networks with ipv4 /31 netmask and ipv6 /127 netmask 

cnvergence:
# Pull Request

## What type of this PR

- Bug fixes

It would be valuable for kube-ovn to support ptp networks, to create external subnets between kube-ovn and the router, following [RFC 3021 ](https://datatracker.ietf.org/doc/html/rfc3021) and [RFC 6164](https://www.rfc-editor.org/rfc/rfc6164).

Fixes 
- https://github.com/kubeovn/kube-ovn/issues/4426


bobz965 (approved):

pkg/util/net.go
```diff
@@ -81,21 +81,32 @@ func SubnetBroadcast(subnet string) string {
 		length = 128
 	}
 	maskLength, _ := cidr.Mask.Size()
+	if maskLength == 31 || maskLength == 127 {
+		return ""
+	}
```
zhangzujian, line 86: What if the `cidr` is an ipv6 one with prefix 31?

cnvergence: Sorry, wanted to add also ipv6 and did mistake

pkg/util/net.go
```diff
@@ -75,45 +75,73 @@ func SubnetNumber(subnet string) string {
...
+			return big.NewInt(0).Sub(size, big.NewInt(1))
+		}
+	default:
+		if maskLength == 31 {
+			return big.NewInt(0).Sub(size, big.NewInt(1))
+		}
+	}
+
+	return big.NewInt(0).Sub(size, big.NewInt(2))
+}
```
zhangzujian, line 143:
```suggestion
func getCIDRSize(length uint, maskLength int) *big.Int {
	size := big.NewInt(0).Lsh(big.NewInt(1), length-uint(maskLength))
	if maskLength+1 == int(length) {
		return big.NewInt(0).Sub(size, big.NewInt(1))
	}
	return big.NewInt(0).Sub(size, big.NewInt(2))
}
```

pkg/util/net.go
```diff
@@ -75,45 +75,73 @@ func SubnetNumber(subnet string) string {
...
+		if maskLength == 31 {
+			return ""
+		}
 		length = 32
 	} else {
+		if maskLength == 127 {
+			return ""
+		}
 		length = 128
 	}
```
zhangzujian, line 89:
```suggestion
	maskLength, length := cidr.Mask.Size()
	if maskLength+1 == length {
		return ""
	}
```

pkg/util/net.go
```diff
@@ -75,45 +75,73 @@ func SubnetNumber(subnet string) string {
...
+	if proto == kubeovnv1.ProtocolIPv4 {
 		length = 32
 	} else {
 		length = 128
 	}
 	maskLength, _ := cidr.Mask.Size()
 	ipInt := IP2BigInt(cidr.IP.String())
-	size := big.NewInt(0).Lsh(big.NewInt(1), length-uint(maskLength))
-	size = big.NewInt(0).Sub(size, big.NewInt(2))
+	size := getCIDRSize(length, maskLength, proto)
```
zhangzujian, line 125:
```go
	maskLength, length := cidr.Mask.Size()
	ipInt := IP2BigInt(cidr.IP.String())
	size := getCIDRSize(length, maskLength)
```

pkg/util/net.go
```diff
@@ -178,8 +206,11 @@ func CheckProtocol(address string) string {
 
 func AddressCount(network *net.IPNet) float64 {
 	prefixLen, bits := network.Mask.Size()
-	if bits-prefixLen < 2 {
-		return 0
+	// Special case handling for /31 and /32 subnets
+	if bits-prefixLen == 1 {
+		return 2 // /31 subnet
+	} else if bits-prefixLen == 0 {
+		return 1 // /32 subnet
```
zhangzujian, line 213: Why should address count be `1` for `/32` or `/128` networks? When and where will a `/32` or `/128` be used?

cnvergence: I have added it for the util function as well, IMO it would be good to validate against this case, as I think it won't be used in kube-ovn subnets. wdyt?

zhangzujian:
Subnet checks fail for /32 subnets:

```yaml
status:
  activateGateway: ""
  conditions:
  - lastTransitionTime: "2024-08-29T01:52:46Z"
    lastUpdateTime: "2024-08-29T01:52:46Z"
    message: 'validate gateway 1.1.1.0 for cidr 1.1.1.0/31 failed: 1.1.1.0 is the
      network number ip in cidr 1.1.1.0/31'
    reason: ValidateLogicalSwitchFailed
    status: "True"
    type: Error
```

We need to fix it.

cnvergence:
Thanks for the review, I have added validation test cases @zhangzujian

zhangzujian:
We need update IPAM to support p2p networks.

cnvergence:
hi @zhangzujian, I have added test cases for IPAM in the pkg/ipam
could you verify where we should add the support for ptp in ipam?

zhangzujian:
> hi @zhangzujian, I have added test cases for IPAM in the pkg/ipam could you verify where we should add the support for ptp in ipam?

I tried the following subnet and deployment but the kube-ovn could not handle the pod creation:

```yaml
apiVersion: kubeovn.io/v1
kind: ProviderNetwork
metadata:
  name: ethx
spec:
  defaultInterface: ethx
---
apiVersion: kubeovn.io/v1
kind: Vlan
metadata:
  name: vlan0
spec:
  id: 0
  provider: ethx
---
apiVersion: kubeovn.io/v1
kind: Subnet
metadata:
  name: s1
spec:
  protocol: IPv4
  cidrBlock: 1.1.1.0/31
  vlan: vlan0
  disableGatewayCheck: true
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: kubeovn
  namespace: default
spec:
  selector:
    matchLabels:
      app: kubeovn
  replicas: 2
  template:
    metadata:
      labels:
        app: kubeovn
      annotations:
        ovn.kubernetes.io/logical_switch: s1
        ovn.kubernetes.io/ip_address: 1.1.1.0,1.1.1.1
    spec:
      containers:
      - name: pod
        imagePullPolicy: Never
        # image: kubeovn/kube-ovn:v1.13.0
        image: kubeovn/kube-ovn:dev
        securityContext:
          privileged: true
        command:
        - sleep
        - infinity
```

If this kind of usage is wrong, please tell me how to use /31 subnets correctly.

cnvergence:
> ```yaml
>   provider: ethx
> ```

I see, this could cause a problem with trying to assign 1.1.1.0 to a Pod, which might be a problem.
Fot the Point-to-point network you would use this external subnet you have created with 1.1.1.0/31, excluding 1.1.1.0, as this will be configured as a gateway and have one available IP for your pod 1.1.1.1

zhangzujian (approved):

PR #4437, opened 2024-08-24
feat(helm): new chart design

### LINKED ISSUE #4330: [Feature Request] Restructuring and introducion best practices for the helm chart
### Description

Currently, the helm chart is quite a mess in my point of view. Mostly it is about the structure and naming of all the values.

I suggest the following improvements:
- define all variables in lowerCamelCase 
- combine all configuration-related vars into one map (to access later for example ".Values.config.logDir" )
- introduce variables for all the Kubernetes-related things for each component (image, labels, annotations, strategy, affinities etc..)
- Document the variables with [helm-docs](https://github.com/norwoodj/helm-docs)

In general, I'm a big fan of the charts from Bitnami or Grafana for example.

**I'm open to contribute those improvements with a PR :)**

Everyone who uses helm for the installation.

### Anything else?

_No response_

SkalaNetworks:
This PR is about issue [#4330](https://github.com/kubeovn/kube-ovn/issues/4330) 

This is a draft, it shouldn't change any resource name as to not break existing installations, but if someone wants to update using the Helm Chart, they'll obviously have to adapt their values.

Fixes #4330

### DIFF
--- a/charts/kube-ovn-v2/values.yaml
+++ b/charts/kube-ovn-v2/values.yaml
@@ -1,3 +1,5 @@
+# -- Global configuration.
+# @section -- Global parameters
 global:
   registry:
     address: docker.io/kubeovn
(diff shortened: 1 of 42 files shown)


SkalaNetworks:
Note that I have no clue what the difference is between DPDK and DPDK-hybrid, the DPDK docs only mention the hybrid version as far as I know

oujonny:
hey @SkalaNetworks what's the state of the PR?

SkalaNetworks:
Hey @oujonny , It's pretty much a fully ready Helm Chart, I need to integrate the changes that have been done to the original in this PR too and put back the labels that are matched by the CLI.

We actually use this chart in production. 

The thing is, the changes in values make the 2 charts completely incompatible. So rolling out this chart needs a notice on how to migrate values and a deprecation notice of the old chart.

I'll update the chart with new changes shortly

SkalaNetworks:
Hey @zhangzujian, do you have an opinion on what we can do with this "new" Helm chart? I say new considering it's kind of a breaking change, users need to adapt their values to the chart. The work isn't even finished, a lot of work could be done to make the values prettier.

I tried not to change them too much as not to kill clusters if someone tries to upgrade their chart to mine, but it is far from a guarantee. 

Maybe release this chart as a V2 alongside the first one and slowly transition to it? It should be easier to maintain as I tried to follow Helm best practices and the standard Helm structure.

I'm starting to work a bit more on network stuff that relates to Kube-OVN so I'll continue modifying the chart pretty much weekly.

zhangzujian:
> Maybe release this chart as a V2 alongside the first one and slowly transition to it? It should be easier to maintain as I tried to follow Helm best practices and the standard Helm structure.

I think a new v2 version would be better. We can deprecate the current version and remove it in the future.

SkalaNetworks:
What's left to do:
- automate the doc generation
- make a nicer readme

SkalaNetworks:
@zbb88888 @zhangzujian I think this PR is ok to me right now, maybe we could get it merged? That way we always modify both Helm charts in this repo when something is added, and I don't have to port the changes myself in my fork!

I also think we should get the chart published to artifacthub

charts/kube-ovn-v2/values.yaml
```diff
@@ -0,0 +1,719 @@
...
+# -- Namespace in which the CNI is deployed.
+# @section -- Global parameters
+namespace: kube-system
+
+# -- Comma-separated list of IPs for each master node.
+# @section -- Global parameters
+masterNodes: ""
+# -- Label used to auto-identify masters.
+# @section -- Global parameters
+masterNodesLabel: "kube-ovn/role=master"
```
zhangzujian, line 40:
```suggestion
masterNodesLabel:
  kube-ovn/role: master
```

Let's make this value a KV pairs.

charts/kube-ovn-v2/values.yaml
```diff
@@ -0,0 +1,719 @@
...
+# @section -- Global parameters
+fullnameOverride: ""
+
+# -- Namespace in which the CNI is deployed.
+# @section -- Global parameters
+namespace: kube-system
+
+# -- Comma-separated list of IPs for each master node.
+# @section -- Global parameters
+masterNodes: ""
```
zhangzujian, line 37:
```suggestion
masterNodes: []
```

charts/kube-ovn-v2/values.yaml
```diff
@@ -0,0 +1,719 @@
...
+  annotations: {}
+  # -- Labels to be added to all top-level agent objects (resources under templates/agent)
+  # @section -- CNI agent configuration
+  labels: {}
+  # -- Annotations to be added to the agent pods (kube-ovn-cni)
+  # @section -- CNI agent configuration
+  podAnnotations: {}
+  # -- Labels to be added to the agent pods (kube-ovn-cni)
+  # @section -- CNI agent configuration
+  podLabels: {}
```
zhangzujian, line 618:
```suggestion
  template:
    annotations: {}
    labels: {}
```

I think this will make it more clear.

charts/kube-ovn-v2/values.yaml
```diff
@@ -0,0 +1,719 @@
...
+  # -- ""
+  # @section -- Network parameters of the CNI
+  enableCompact: false
+
+# -- Features of Kube-OVN we wish to enable/disable.
+# @section -- Opt-in/out Features
+features:
+  ENABLE_LB: true
+  ENABLE_NP: true
+  ENABLE_EXTERNAL_VPC: true
```
zhangzujian, line 170:
```suggestion
  EnableLoadBalacner: true
  EnableNetworkPolicy: true
  EnableExternalVPC: true
```

Let's use k8s feature-gate style names.

charts/kube-ovn-v2/values.yaml
```diff
@@ -0,0 +1,719 @@
...
+  HW_OFFLOAD: false
+  ENABLE_LB_SVC: false
+  ENABLE_KEEP_VM_IP: true
+  LS_DNAT_MOD_DL_DST: true
+  LS_CT_SKIP_DST_LPORT_IPS: true
+  CHECK_GATEWAY: true
+  LOGICAL_GATEWAY: false
```
zhangzujian, line 177: These values/features are for the default subnet, so they should be moved to `networking.pods`.

charts/kube-ovn-v2/values.yaml
SkalaNetworks: Done. I made it a list and modified the helper functions.

charts/kube-ovn-v2/values.yaml
SkalaNetworks: Done. Each label given in the dictionnary is matched. So it's not a labelSelector, you don't need all the labels to match to be elected a master. But I think it's a better alternative.

charts/kube-ovn-v2/values.yaml
SkalaNetworks: I copied that concept from Cilium's Helm Chart, I think the comment is pretty clear on what this value edits, but I can modify each daemon's value to reflect your change if you want to.

The idea is that each daemon has a bunch of config for it (service accounts, roles, pods...) and the "annotations/labels" value add annotations and labels to every resource of that daemon. And podLabels/podAnnotations only add to the actual pod of the daemon.

charts/kube-ovn-v2/values.yaml
SkalaNetworks: I'm gonna work on this. I'll still use the standard camel case syntax to respect Helm standards.
There's a lot of work to do on the old "func" dict, I didn't touch it too much because I'm not sure what is still relevant or not.

charts/kube-ovn-v2/values.yaml
SkalaNetworks: I moved it.

charts/kube-ovn-v2/values.yaml
SkalaNetworks: @zhangzujian https://github.com/kubeovn/kube-ovn/pull/4437/commits/34574c575a85eae5b51c8653c457e8892692b637

this commit addresses some of the feature flags, I left some of them because I have no idea what they do. If you want to add some more comments to each flag to make the values more bearable to new users, feel free to give me feedback

zhangzujian:
There are still some jobs we need to do, including chart adjustment/modification, document updates and CI testing. This PR is an initial commit of the new chart, and I think we can merge it for now.

@SkalaNetworks Thanks a lot for your contribution!

PR #4557, opened 2024-09-23
Make NAT gateway generation prefix a variable

SkalaNetworks:
This makes the prefix in front of Pods/STS generated for NAT gateways a configurable variable from the ConfigMap.

### DIFF
--- a/pkg/controller/vpc_nat.go
+++ b/pkg/controller/vpc_nat.go
@@ -27,6 +27,8 @@ func (c *Controller) resyncVpcNatConfig() {
 	vpcNatGwNamePrefix := cm.Data["natGwNamePrefix"]
 	if vpcNatGwNamePrefix != "" {
 		util.VpcNatGwNamePrefix = vpcNatGwNamePrefix
+	} else {
+		util.VpcNatGwNamePrefix = util.VpcNatGwNameDefaultPrefix
 	}
 
 	// Image we're using to provision the NAT gateways
--- a/pkg/controller/vpc_nat.go
+++ b/pkg/controller/vpc_nat.go
@@ -22,6 +22,13 @@ func (c *Controller) resyncVpcNatConfig() {
 		return
 	}
 
+	// Prefix used to generate the name of the StatefulSet/Pods for a NAT gateway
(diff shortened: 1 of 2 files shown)


pkg/util/vpc_nat_gateway.go
```diff
@@ -2,10 +2,17 @@ package util
 
 import "fmt"
 
+var (
```
bobz965, line 5: how about using a const?

SkalaNetworks: Allows for changing the prefix on the fly from other packages, such as the one managing the configmap for the nat gateway

bobz965: ok

bobz965 (approved):

bobz965:
@SkalaNetworks please fix： 
image

pkg/controller/vpc_nat.go
```diff
@@ -22,6 +22,13 @@ func (c *Controller) resyncVpcNatConfig() {
 		return
 	}
 
+	// Prefix used to generate the name of the StatefulSet/Pods for a NAT gateway
+	// By default it is equal to the value contained in 'util.VpcNatGwNamePrefix'
+	vpcNatGwNamePrefix := cm.Data["natGwNamePrefix"]
+	if vpcNatGwNamePrefix != "" {
+		util.VpcNatGwNamePrefix = vpcNatGwNamePrefix
+	}
```
zhangzujian, line 30: What if `natGwNamePrefix` was set back to empty?

SkalaNetworks: Bad things happen I guess :) I'll fix that

SkalaNetworks:
Hey, sorry I had a lot of stuff on my plates, just fixed this so the prefix resets back to default if the value is changed to default.

Note that changing a prefix AFTER NAT gateways have been provisioned is a destructive operation as Kube-OVN will fail to track the old gateways (it uses labels/names, and they don't match the new prefix anymore). I'll document that in the docs if this is merged.

zhangzujian (approved):

PR #4714, opened 2024-11-08
[bugfix] Optimize node gc method

cmdy:
# Pull Request

## What type of this PR

- Bug fixes

Fixes #(issue-number)
1. When the controller is stopped, the node is deleted, and then the controller is restarted, the logical router policy with a priority of 29000 is not properly cleared. 
image
2. Although ovn-controller will skip the error in the picture, it will cause some retries，this causes the ovn-controller to start slowly and the flow table information to be delivered slowly in a large-scale cluster(2000+ node,15w+ pod,100+ subnet).
3. #4194 This PR caused the node deletion logic in the gc node to not be actually executed.

### DIFF
--- a/pkg/controller/gc.go
+++ b/pkg/controller/gc.go
@@ -233,7 +233,7 @@ func (c *Controller) gcNode() error {
 		if strings.HasPrefix(ip.Name, util.NodeLspPrefix) && !strings.Contains(ip.Name, ".") {
 			if node := ip.Name[len(util.NodeLspPrefix):]; !nodeNames.Has(node) {
 				klog.Infof("gc node %s", node)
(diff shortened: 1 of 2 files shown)


oilbeater:
@cmdy the u2o e2e failed. Seems like some policy rules get lost.

cmdy:
> @cmdy the u2o e2e failed. Seems like some policy rules get lost.

In release-1.12, e2e test file not found, what should I do to pass the check

oilbeater:
It has been backport in https://github.com/kubeovn/kube-ovn/commit/b23c8da43aff75d3cce8aa7c703d8d6bb5a2e488

PR #4724, opened 2024-11-11
lb-svc is compatible with dual stack subnets and supports attachment network for device plugin types

coldzerofear:
lb-svc is compatible with dual stack subnets
lb-svc supports attachment network for device plugin types

# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)

### DIFF
--- a/pkg/controller/service.go
+++ b/pkg/controller/service.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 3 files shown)


oilbeater:
@coldzerofear, could you introduce the scenarios in which it is necessary to add device plugin devices to the lb-svc pods? I'm considering whether the vpc-nat-gateway also needs a similar configuration.

coldzerofear:
> @coldzerofear, could you introduce the scenarios in which it is necessary to add device plugin devices to the lb-svc pods? I'm considering whether the vpc-nat-gateway also needs a similar configuration.

Some network providers use device plugins to allocate devices or enhance scheduling, multus supports passing deviceID assigned by device plugins to CNI for binding container networks. Although it is uncertain how many CNIs can benefit from it, I think it should be compatible with these features

coldzerofear:
> @coldzerofear, could you introduce the scenarios in which it is necessary to add device plugin devices to the lb-svc pods? I'm considering whether the vpc-nat-gateway also needs a similar configuration.

I have another suggestion, hoping that lb's pod can have more configurable items, such as resource request size, so that lb pod can run with guaranteed QoS

oilbeater:
> I have another suggestion, hoping that lb's pod can have more configurable items, such as resource request size, so that lb pod can run with guaranteed QoS

That make sense. I think we may need a template way to deploy lb-svc pod and vpc-nat-gateway.

oilbeater (approved):

PR #4828, opened 2024-12-13
cni: do not exit if the sysctl variable does not exist or can not be set

zhangzujian:
# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)

### DIFF
--- a/cmd/cni/sysctl.go
+++ b/cmd/cni/sysctl.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/containernetworking/plugins/pkg/ns"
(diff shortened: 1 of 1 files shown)


oilbeater (approved):

patriziobassi:
@zhangzujian shouldn't we add a log entry too when error occurs?

zhangzujian:
> @zhangzujian shouldn't we add a log entry too when error occurs?

It's exexcuted in the CNI plugin, not kube-ovn-cni. Any logging will NOT passed to kubelet if the CNI command succeeds. We need a new mechanism to store the logging to a file on the host, or to pass it to kube-ovn-cni.

Issue #4841, opened 2024-12-17
[BUG] Controller crashes if ENABLE_NP is false and not specifying a gateway node


SkalaNetworks:
### Kube-OVN Version

v1.13/v1.14

### Kubernetes Version

v1.31.0

### Operation-system/Kernel Version

/

### Description

The controller crashes when submitting a new Subnet to the K8S API through the Go SDK if network policies have been disabled.

The reason is a panic due to a nil pointer found here: https://github.com/kubeovn/kube-ovn/blob/77650e5931c708dc39bba1792cbe93d9eb368613/pkg/controller/subnet.go#L72

We try to use a npLister, which isn't initialized due to network policies being disabled. The only way to not trigger that call is for the above condition to evaluate to false. My example code doesn't trigger that code if I set the gateway manually, but if I don't specify the gateway, it seems Kube-OVN accepts the Subnet as-is, and a millisecond later edits the Gateway to the first IP in the subnet.

This means there's 2 events in quick succession:
- Creation (triggered by my code)
- Modification to set the gateway to the first IP (triggered by Kube-OVN)

And then it crashes. I don't understand how my call from Go code triggers the modification of the gateway, but doing it using kubectl doesn't. If I don't specify the gateway in a YAML file and submit it, no crash happens.

I am unsure if my way of creating Subnets using the SDK is the right one considering this bug, but it's the standard K8S way.

I also don't know why NPs need to be re-enqued if the gateway changes, but adding a condition that fails to check if npLister is nil would be a quick fix for this problem.

### Steps To Reproduce

1. Deploy latest Kube-OVN build
2. Write go code that creates a Subnet using the Go SDK:
```go
	subnet := v1.Subnet{
		ObjectMeta: metav1.ObjectMeta{Name: "this-crashes-controller"},
		Spec: v1.SubnetSpec{
			Vpc:        "dummy",
			Default:    true,
			CIDRBlock:   "10.60.0.0/24",
			NatOutgoing: false,
			//Gateway:     "10.60.0.1", // We are not specifying a gateway, if we remove the comment, it doesn't crash anymore
			GatewayType: "", 
			Protocol:    "IPv4",
		},
	}

	return kubernetes.KubeOvnClient.KubeovnV1().Subnets().Create(context.Background(), &subnet, metav1.CreateOptions{})
```
3. Witness the controller crashing when that code is called

### Current Behavior

Controller crashes

### Expected Behavior

Controller doesn't crash

[...]

PR #4848, opened 2024-12-18
fix(controller/subnet): controller crashes on subnets if gateway is u…

SkalaNetworks:
This PR references #4841

### DIFF
--- a/pkg/controller/subnet.go
+++ b/pkg/controller/subnet.go
@@ -50,14 +50,16 @@ func (c *Controller) enqueueUpdateSubnet(oldObj, newObj interface{}) {
 	newSubnet := newObj.(*kubeovnv1.Subnet)
 	key := cache.MetaObjectToName(newSubnet).String()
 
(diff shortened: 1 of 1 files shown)


pkg/controller/subnet.go
```diff
@@ -50,14 +50,16 @@ func (c *Controller) enqueueUpdateSubnet(oldObj, newObj interface{}) {
...
-	if newSubnet.Spec.Gateway != oldSubnet.Spec.Gateway ||
-		newSubnet.Status.U2OInterconnectionIP != oldSubnet.Status.U2OInterconnectionIP {
-		policies, err := c.npsLister.List(labels.Everything())
-		if err != nil {
-			klog.Errorf("failed to list network policies: %v", err)
-		} else {
-			for _, np := range policies {
-				c.enqueueAddNp(np)
+	// Trigger network policy refresh only if they are enabled, otherwise the lister will be nil
+	if c.npsLister != nil {
```
bobz965, line 54: how about using ?

if c.npsLister == nil {
log
return
}

Issue #4922, opened 2025-01-13
[BUG] IP address leak if adding EIP fails


cruickshankpg:
### Kube-OVN Version

v1.12.22

### Kubernetes Version

v1.28.6

### Operation-system/Kernel Version

"Ubuntu 22.04.5 LTS" 6.8.0-47-generic

### Description

If an EIP fails to be added to a NAT gateway it allocates an IP. If the EIP is deleted before the NAT gateway comes up then the IP address allocation is leaked in IPAM. Restarting the kube-ovn-controller leader removes the leak 

```
0110 16:34:03.854233       7 vpc_nat_gw_eip.go:208] handle add iptables eip poison-eip
I0110 16:34:03.854278       7 ipam.go:60] allocate v4 10.146.46.215, v6 , mac 72:b4:28:77:2c:b0 for poison-eip from subnet ovn-vpc-external-network
E0110 16:34:03.854395       7 vpc_nat_gw_eip.go:425] pod.apps "doesn-not-exist" not found
E0110 16:34:03.854405       7 vpc_nat_gw_eip.go:247] failed to create eip 'poison-eip' in pod, pod.apps "doesn-not-exist" not found
E0110 16:34:03.854470       7 vpc_nat_gw_eip.go:110] error syncing 'poison-eip': pod.apps "doesn-not-exist" not found, requeuing
```

The IP address is added to the external subnet's status' `v4usingIPrange` and removed from the the `v4availableIPrange` but `v4usingIPs` is not updated.

```
  v4availableIPrange: 10.146.46.204-10.146.46.211,10.146.46.217-10.146.46.254
  v4availableIPs: 48
  v4usingIPrange: 10.146.46.212-10.146.46.216
  v4usingIPs: 3
```

### Steps To Reproduce

Create an eip

```
cat <<EOF | kubectl apply -f -
apiVersion: kubeovn.io/v1
kind: IptablesEIP
metadata:
  name: poison-eip
spec:
  natGwDp: does-not-exist
  externalSubnet: ovn-vpc-external-network
EOF
```

Wait a moment for it to get programmed

Delete the eip

```
kubectl delete eip poison-eip
```

Inspect the subnet status

### Current Behavior

EIP IP allocation leaked

### Expected Behavior

EIP IP allocation not leaked

cruickshankpg:
[...]

PR #4924, opened 2025-01-14
make sure gw pod exist before eip creation

zbb88888:
# Pull Request

## What type of this PR

- Bug fixes

https://github.com/kubeovn/kube-ovn/issues/4922

Fixes #(issue-number)


cruickshankpg:
thanks for this. I think this will prevent the leak when the EIP is misconfigured. There's still the risk that the nat gateway is broken in some way and the exec in pod fails which may cause the IP to leak but that's more of an edge case

zbb88888:
> thanks for this. I think this will prevent the leak when the EIP is misconfigured. There's still the risk that the nat gateway is broken in some way and the exec in pod fails which may cause the IP to leak but that's more of an edge case

yeah, if the pod crashes, and can not run up later, the EIP is leaked. how about talking about theses cases in detail later, we will try to help fix them too.

zbb88888:
@oilbeater  这个ut降低跟这个pr应该也没关系

oilbeater (approved):

PR #4933, opened 2025-01-16
Remove kubevirt/pkg/controller dependency

xrstf:
# Pull Request

## What type of this PR
- Refactoring

By importing `kubevirt.io/kubevirt`, KubeOVN also has to import and deal with an incredible amount of legacy cruft. Compared to `kubevirt.io/api`, the Go dependencies of the main kubevirt module are not so very much up-to-date.

KubeOVN only uses a single struct from the kubevirt codebase, and of that struct just very little. I propose that it is cheaper and in the long-term more maintainable to copy-paste the required code into this repository. This gets rid of the heavy dependency and quite a few `replace` directives. This in turn helps other projects that import KubeOVN into their codebase.

Since both projects are Apache Licensed, I would consider this an okay-ish hack with more upsides than downsides.


oilbeater:
@xrstf the conflict need to be resolved.

oilbeater:
It has been merged by https://github.com/kubeovn/kube-ovn/pull/5065, thanks @xrstf

PR #5014, opened 2025-02-19
fix logging in ovn lr/ls gc

zhangzujian:
we should print lr/ls names instead of ovn nb objects.

# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)


cloudnull (approved):
LGTM and the improved logging messages in GC are very welcomed.

PR #5141, opened 2025-04-10
Deleting FIP sets EIP to ready

### LINKED ISSUE #5114: [BUG] Deleting FIP overwrites EIP ready state
### Kube-OVN Version

v1.13.4

### Kubernetes Version

v1.28.6

"Ubuntu 22.04.5 LTS" 6.8.0-47-generic

### Description

Deleting a FIP triggers a reset of the associated EIP. Resetting an EIP sets `status.ready: true` even if the EIP is not yet ready. The EIP update handler then does not program the EIP on the NAT GW.

We hit this following a NAT GW being rescheduled to a new node, this triggers all FIPs and EIPs to be marked as not ready so they can be programmed on the new pod.  A FIP was deleted and recreated while all this was going on but the associated EIP was never added to the new gateway pod.

### Steps To Reproduce

It's a race condition so it can be difficult to consistently recreate this bug.

1. Create vpc nat gw
2. Create a lot of EIPs and FIPs on vpc nat gw
3. Delete vpc nat gw pod
4. Delete all FIPs
5. Exec onto vpc nat gw pod and check ip addresses on net1

### Current Behavior

Deleting a FIP while an EIP is not yet programmed prevents the EIP ever being programmed.

### Expected Behavior

Deleting a FIP while an EIP is not yet programmed has no impact on EIP being programmed.

andrewlee1089:
# Pull Request

## What type of this PR

- Previously when deleting a FIP, the function `handleResetOvnEip` was called.
- `handleResetOvnEip` called `patchOvnEipStatus` to set the "Ready"
  field to true, regardless of the current EIP.
- This fixes it so when `handleResetOvnEip` is called, the "Ready" is
  field of the EIP is unchanged 

I have rejected a more ambitious fix where all calls of the `handleResetOvnEip` function set the EIP status to not be ready.

Fixes #5114

### DIFF
--- a/pkg/controller/ovn_eip.go
+++ b/pkg/controller/ovn_eip.go
@@ -193,7 +193,7 @@ func (c *Controller) handleResetOvnEip(key string) error {
 		return nil
 	}
 	klog.Infof("handle reset ovn eip %s", cachedEip.Name)
-	if err := c.patchOvnEipStatus(key, true); err != nil {
+	if err := c.patchOvnEipStatus(key, false); err != nil {
 		klog.Errorf("failed to reset nat for eip %s, %v", key, err)
 		return err
 	}
@@ -364,7 +364,7 @@ func (c *Controller) createOrUpdateOvnEipCR(key, subnet, v4ip, v6ip, mac, usageT
 	return nil
 }
 
-func (c *Controller) patchOvnEipStatus(key string, ready bool) error {
+func (c *Controller) patchOvnEipStatus(key string, markEIPAsReady bool) error {
 	cachedOvnEip, err := c.ovnEipsLister.Get(key)
 	if err != nil {
 		klog.Errorf("failed to get cached ovn eip '%s', %v", key, err)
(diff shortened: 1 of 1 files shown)


andrewlee1089:
I'm not entirely sure of the implications of this PR - hence why this is in draft state.

zbb88888:
maybe you are right?
1. iptables-eip has the ready status, so I use the `ready` in ovn-eip status. but ovn-eip is just a ip, it has no process of `setting ip on the nic`
2. the ready status can help to make sure the ovn-eip v4 and v6 ip is already exist
3. the ip crd has no ready status, so you can just remove the `ready` status.

andrewlee1089:
@zbb88888 I can't get the pipeline to pass (I think it is failing for an unrelated reason) but I think this PR makes sense now. Please can you review?

pkg/controller/ovn_eip.go
```diff
@@ -364,17 +364,19 @@ func (c *Controller) createOrUpdateOvnEipCR(key, subnet, v4ip, v6ip, mac, usageT
 	return nil
 }
 
-func (c *Controller) patchOvnEipStatus(key string, ready bool) error {
+func (c *Controller) patchOvnEipStatus(key string, ready, changeReady bool) error {
```
zbb88888, line 367: how about removing  `ready, changeReady` and eip status `ready`,  keep the ovn-eip just as IP crd？ keep it simple

andrewlee1089:
Hi @zbb88888 , I've finally got CI to pass and I think I am happy with the scope of this fix.

Are you able to review?

zbb88888 (approved):

zbb88888:
> Hi @zbb88888 , I've finally got CI to pass and I think I am happy with the scope of this fix.
> 
> Are you able to review?

ok, LGTM, Thanks

PR #5144, opened 2025-04-10
Support /32 tunnel-source

netdever:
# Pull Request

## What type of this PR

Feature

## PR Description
 
There exists a use case for sourcing tunnel traffic from a /32 address assigned to the lo interface of the node. The lo interface typically comes with a localhost address (ie. 127.0.0.1). This change excludes localhost addresses from acting as the tunnel source. It also permits using a /32 address for the tunnel source if the host-tunnel-src variable is set to true (false by default).


oilbeater:
@netdever the DCO check failed, can you sign off the commit?

netdever:
@oilbeater Rebased and signed off.

oilbeater:
Thanks! @netdever

PR #5287, opened 2025-05-29
fix podcidr route still use join ip as src ip

changluyi:
# Pull Request

## What type of this PR
企业微信截图_17484211222774
when upgrade from lower version to 1.12.29 , the route is not auto replace.

the podcidr route on node should use nodeip as src ip

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)

### DIFF
--- a/pkg/daemon/controller_linux.go
+++ b/pkg/daemon/controller_linux.go
@@ -401,10 +401,9 @@ func routeDiff(nodeNicRoutes, allRoutes []netlink.Route, cidrs, joinCIDR []strin
 		}
 
 		found := false
-		_, cidrNet, _ := net.ParseCIDR(c)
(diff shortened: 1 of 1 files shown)


pkg/daemon/controller_linux.go
```diff
@@ -401,26 +401,30 @@ func routeDiff(nodeNicRoutes, allRoutes []netlink.Route, cidrs, joinCIDR []strin
 		}
 
 		found := false
+		_, cidrNet, _ := net.ParseCIDR(c)
```
oilbeater, line 404: c 已经是 cidrNet.String() 格式了不需要再转换

oilbeater (changes requested):

PR #5349, opened 2025-06-11
fix(slr): address family and familypolicy isn't correct

SkalaNetworks:
# Pull Request

## What type of this PR

Bug fix on the SwitchLBRules.

The ipFamilies and ipFamilyPolicy aren't honored for the VIP. Because the clusterIP isn't set, K8S determines the family using its defaults. 

So if the cluster was installed with a ServiceCIDR that contains the IPv4 range first and the IPv6 second, the family will be set to IPv4 by default. This works in nearly every case.

But if the ServiceCIDR contains the IPv6 range first and the IPv4 range second, the Service is created with ipFamily IPv6, eventhough the VIP might be an IPv4. This breaks the SLR and no endpoint is generated for the service.

The VIP might contain IPv4/IPv6 or a mix of the two. Depending on the content of the VIP, we detect if the Service should be created with ipFamilies IPv4, IPv6 or both. We also compute the ipFamilyPolicy to SingleStack or Dual depending on the number of ipFamilies.


pkg/controller/switch_lb_rule.go
```diff
@@ -406,3 +416,35 @@ func generateEndpoints(slr *kubeovnv1.SwitchLBRule, oldEps *corev1.Endpoints) *c
...
+	if isV4 { // One IP in the VIP is v4
+		families = append(families, corev1.IPv4Protocol)
+	}
+
+	policy = corev1.IPFamilyPolicySingleStack
+	if isV4 && isV6 {
+		policy = corev1.IPFamilyPolicyPreferDualStack
+	}
+
+	return families, policy
```
zhangzujian:
```suggestion
	// Check every IP in the VIP, assess if it is an IPv6 or an IPv4
	ipFamilies := set.New[corev1.IPFamily]()
	for ip := range strings.SplitSeq(vip, ",") {
		switch util.GetIPAddressProtocol(ip) {
		case kubeovnv1.ProtocolIPv6:
			ipFamilies.Insert(corev1.IPv6Protocol)
		case kubeovnv1.ProtocolIPv4:
			ipFamilies.Insert(corev1.IPv4Protocol)
		}
	}

	policy = corev1.IPFamilyPolicySingleStack
	if ipFamilies.Len() > 1 {
		policy = corev1.IPFamilyPolicyPreferDualStack
	}

	return ipFamilies.SortedList(), policy
```

SkalaNetworks: @zhangzujian done

pkg/controller/switch_lb_rule.go
```diff
@@ -406,3 +417,25 @@ func generateEndpoints(slr *kubeovnv1.SwitchLBRule, oldEps *corev1.Endpoints) *c
...
 }
+
+// getIPFamilies returns the IP families (IPv6/IPv4) for a set of IPs within a VIP
+// This function is used to correctly construct the Service of a SwitchLBRule
+// It also returns the corresponding IPFamilyPolicy to set in the Service
+func getIPFamilies(vip string) (families []corev1.IPFamily, policy corev1.IPFamilyPolicy) {
+	// Check every IP in the VIP, assess if it is an IPv6 or an IPv4
+	ipFamilies := set.New[corev1.IPFamily]()
+	for ip := range strings.SplitSeq(vip, ",") {
+		switch util.GetIPAddressProtocol(ip) {
```
zhangzujian: Can we use `utli.CheckProtocol()`?

SkalaNetworks: @zhangzujian sure, done

PR #5376, opened 2025-06-19
fix(vip): support vips/slrs targeting subnets with non-default pr…

SkalaNetworks:
VIPs for SLRs get garbage collected by accident because their portName at creation (which includes the provider of the subnet) doesn't match the default `.ovn` provider used during the GC check.

The GC now checks LSPs for VIPs using the same name that should have been computed at creation of the LSP.


pkg/controller/gc.go
```diff
@@ -395,7 +395,19 @@ func (c *Controller) markAndCleanLSP() error {
...
+			if subnetName == "" {
+				klog.Errorf("failed to gc vip '%s', subnet should be set", vip.Name)
+				continue
+			}
+
+			subnet, err := c.subnetsLister.Get(subnetName)
+			if err != nil {
+				klog.Errorf("failed to get subnet %s: %v", subnetName, err)
+				continue
+			}
```
zhangzujian, line 408: Should we check whether the subnet is deleted or not found?

SkalaNetworks: @zhangzujian I think we shouldn't, and we should abort at any error.

I know you can't delete Subnets as long as they have IPs, but this is a VIP, so I'm not sure it's handled the same way.

It means that technically, the VIP could exist but not its Subnet. It would fail here because we can't retrieve the provider. 

If we can't retrieve the provider, we can't compute the full portName for that VIP (it isn't stored anywhere, status or else..). In that case, we should GC the LSP because it has obviously become useless.

By aborting with an error here, the LSP will never be matched against the vipsMap (because nothing is added to the vipsMap) and it should be deleted later in the code.

zbb88888: I think the vip should be in a subnet

SkalaNetworks: @zbb88888 https://github.com/kubeovn/kube-ovn/blob/5a66b6626f2cf01203e5575f9767349ff5a923c7/pkg/controller/vip.go#L71

This is the code that is triggered when a VIP is added. It checks if it has a Subnet, errors out if it doesn't. This is the same code that I imported in the GC for consistency.

Now, VIPs not only work on  "ovn" providers, but also on custom ones

zbb88888: ok

PR #5386, opened 2025-06-21
Add network attachment informer to controller

### LINKED ISSUE #5353: [BUG] Slow controller processing due to fetching network attachment definitions
### Kube-OVN Version

v1.13.12

### Kubernetes Version

v1.28.6

Ubuntu 22.04.5 LTS 6.8.0-52-generic

### Description

In our test cluster we are seeing high work queue depths (~100) for AddSubnets and long delays actually adding subnets leading to errors provisioning VMs as IPAM isn't aware of the subnets yet. 

We are using multus and all subnets have a corresponding NetworkAttachmentDefinition. Our test cases spin up new subnets and NADs for each test so there is relatively steady churn.

Looking at a CPU profile we're spending 19% of total CPU time in `handleAddOrUpdateSubnet` and then 11% of total time getting network attachment definitions.

Image

Image

It doesn't look like the controller caches NetworkAttachmentDefinitions so getting them becomes really expensive.

### Steps To Reproduce

Create subnets that use network attachment definitions.

### Current Behavior

Controller makes a network call whenever it needs to read a network attachment definition.

### Expected Behavior

Controller uses a cache.

cruickshankpg:
# Pull Request

## What type of this PR

- Bug fixes

This allows caching of network attachment definitions and should improve performance when using multus.

Fixes #5353

### DIFF
--- a/charts/kube-ovn-v2/templates/rbac/ovn-CR.yaml
+++ b/charts/kube-ovn-v2/templates/rbac/ovn-CR.yaml
@@ -84,6 +84,8 @@ rules:
       - network-attachment-definitions
     verbs:
       - get
+      - list
+      - watch
   - apiGroups:
       - ""
       - networking.k8s.io
--- a/pkg/controller/network_attachment.go
+++ b/pkg/controller/network_attachment.go
@@ -1,0 +1,79 @@
+package controller
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"gopkg.in/k8snetworkplumbingwg/multus-cni.v4/pkg/logging"
+	multustypes "gopkg.in/k8snetworkplumbingwg/multus-cni.v4/pkg/types"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+
+	"github.com/kubeovn/kube-ovn/pkg/util"
+)
+
+func (c *Controller) isNetAttachCRDInstalled() (bool, error) {
+	resources, err := c.config.AttachNetClient.Discovery().ServerResourcesForGroupVersion("k8s.cni.cncf.io/v1")
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return false, nil
+		}
+		return false, fmt.Errorf("getting server resources for k8s.cni.cncf.io/v1; %w", err)
+	}
+
+	for _, resource := range resources.APIResources {
+		if resource.Name == "network-attachment-definitions" {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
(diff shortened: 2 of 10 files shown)


charts/kube-ovn-v2/templates/rbac/ovn-CR.yaml
```diff
@@ -84,6 +84,7 @@ rules:
       - network-attachment-definitions
     verbs:
       - get
+      - list
```
zhangzujian, line 87: Missing `watch`.

oilbeater:
@cruickshankpg the ut failed and you can check https://github.com/kubeovn/kube-ovn/pull/5390 to refactor multus informer. If multus is installed after kube-ovn, we need a mechanism to auto discover it and start the informer.

cruickshankpg:
> @cruickshankpg the ut failed and you can check #5390 to refactor multus informer. If multus is installed after kube-ovn, we need a mechanism to auto discover it and start the informer.

OK I've changed it to always start the informer but only to check if it has synced in the event the CRD has installed. The consequence is there is now a benign error that comes up every 30s from the informer if the CRD doesn't exist.

oilbeater:
> The consequence is there is now a benign error that comes up every 30s from the informer if the CRD doesn't exist.

You can refer to [this implementation](https://github.com/kubeovn/kube-ovn/blob/master/pkg/controller/kubevirt.go#L183), which uses a separate go routine to loop the CRD existent and then starts the list and watch.

pkg/controller/network_attachment.go
```diff
@@ -0,0 +1,78 @@
...
+func (c *Controller) StartNetAttachInformerFactory(ctx context.Context) {
+	ticker := time.NewTicker(10 * time.Second)
+	go func() {
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ticker.C:
+				exists, err := c.isNetAttachCRDInstalled()
+				if err != nil {
+					klog.Errorf("checking network attachment CRD exists: %w", err)
```
oilbeater: klog doesn't support `%w`

oilbeater (changes requested):

pkg/controller/network_attachment.go
```diff
@@ -0,0 +1,79 @@
...
+			select {
+			case <-ticker.C:
+				exists, err := c.isNetAttachCRDInstalled()
+				if err != nil {
+					klog.Errorf("checking network attachment CRD exists: %v", err)
+					continue
+				}
+
+				if exists {
+					klog.Info("Start kubevirt informer")
```
oilbeater:
```suggestion
					klog.Info("Start attachment informer")
```

oilbeater (approved):

oilbeater:
It's more difficult than we expected before. Thanks, @cruickshankpg !

PR #5414, opened 2025-07-01
feat(endpoint_slice): refactor and support arbitrary providers

SkalaNetworks:
# Pull Request

## What type of this PR

- Features

This PR refactors the way endpoint slices are handled. It makes the logic of detecting the VPC/subnet targetted by a service more straightforward and supports custom providers instead of just the "ovn" one.

Support for specifying "endpoints" instead of selectors is also improved, with custom providers added to them. The refactor will make the implementation of healthchecks for them easier.

Note that this PR doesn't fix the obvious issue where a service targets pods that are in different VPC/subnets. It also only handles the first network of every Pod. For this, we'd need more checks and to use EndpointSlice mirrors.

### DIFF
--- a/pkg/controller/switch_lb_rule_test.go
+++ b/pkg/controller/switch_lb_rule_test.go
@@ -4,6 +4,11 @@ import (
 	"reflect"
 	"testing"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	kubeovnv1 "github.com/kubeovn/kube-ovn/pkg/apis/kubeovn/v1"
+	"github.com/kubeovn/kube-ovn/pkg/util"
+
 	corev1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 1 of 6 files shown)


SkalaNetworks:
/cc @zhangzujian @oilbeater 

This should help for people using custom providers. Especially useful on multi-attached VMs that have many network interfaces of type Kube-OVN. 

I'm gonna add healthchecks on custom endpoints next (right now, only selectors are supported).

SkalaNetworks:
@zhangzujian @oilbeater I've prepared a PR to add healthchecks on static endpoints https://github.com/kubeovn/kube-ovn/pull/5435

I'll wait on this one to be merged to get it out of draft as it depends on code present here.

pkg/controller/switch_lb_rule_test.go
```diff
@@ -4,6 +4,11 @@ import (
 	"reflect"
 	"testing"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
zhangzujian: This package import should be placed together with `corev1 "k8s.io/api/core/v1"`.

zhangzujian:
Could you please add an e2e test case for this feature?

SkalaNetworks:
> Could you please add an e2e test case for this feature?

will do

test/e2e/kube-ovn/switch_lb_rule/switch_lb_rule.go
```diff
@@ -120,10 +114,40 @@ var _ = framework.Describe("[group:slr]", func() {
...
+
+		ginkgo.By("Creating custom overlay subnet")
+
+		var provider string
+		if customProvider {
+			provider = fmt.Sprintf("%s.%s.%s", subnetName, f.Namespace.Name, util.OvnProvider)
+		}
+
+		annotations := make(map[string]string)
+		if customProvider {
```
zhangzujian:
```suggestion
		if customProvider {
			f.SkipVersionPriorTo(1, 15, "This feature was introduced in v1.15")
```

SkalaNetworks: Done

test/e2e/kube-ovn/switch_lb_rule/switch_lb_rule.go
```diff
@@ -120,10 +114,39 @@ var _ = framework.Describe("[group:slr]", func() {
...
+		} else {
+			annotations[util.LogicalSwitchAnnotation] = subnetName
+		}
+
+		overlaySubnet := framework.MakeSubnet(subnetName, "", overlaySubnetCidr, "", vpcName, provider, nil, nil, nil)
+		_ = subnetClient.CreateSync(overlaySubnet)
+
+		if customProvider {
+			nad := framework.MakeOVNNetworkAttachmentDefinition(subnetName, f.Namespace.Namespace, provider, []request.Route{})
+			nadClient.Create(nad)
```
zhangzujian, line 153: This test case requires Multus-CNI. We can move it to `test/e2e/multus/e2e_test.go`.

SkalaNetworks: @zhangzujian  Considering it's gonna run pretty much exactly the same tests as the standard SLRs that don't use Multus, how do you envision it? I'd like to minimize code duplication.

Maybe we can force Multus installations for SLR tests?

zhangzujian: We can check whether the CRD exists and skip this test case if not found.

Also, we can add a ginkgo label such as `multus-cni` that indicates this case requires Multus-CNI.

SkalaNetworks: I implemented an if to check if Multus is installed. I'll see for the label.

Issue #5318, opened 2025-06-05
[BUG] underlay u2o routing stopped working in 1.14


patriziobassi:
### Kube-OVN Version

v1.14.0

### Kubernetes Version

v.1.30.13

### Operation-system/Kernel Version

5.15.0-119-generic

### Description

we have cluster k8s with kubeov 1.12.6 in underlay mode and u2o enabled.

we can create pods, they get the underlay ip address and if inside that pod we do 
curl http://<dns-external> we can get to the destination.

upgrading the cluster to 1.14 with the same layout do not work: the dns is resolved (so the u2o to the coredns overlay pod is fine) but that we se ARP traffic is lost between the logical gw (u2o gw) and the real subnet gateway. so the pod cannot reach the destination.

this is a regression, we tought something was broken during the cni update from 1.12 to 1.14 so we scratched the whole cluster re-deploying 1.14 since the beginning but no luck.

### Steps To Reproduce

.

### Current Behavior

.

### Expected Behavior

.

patriziobassi:
Hi, i tested again v.1.14.0 (from dockerhub) and still fails.

So i manually applied https://github.com/kubeovn/kube-ovn/pull/5192 patch and recompiled but it is still not fixing the problem.

So basically when u2o is enabled underlay pods can resolve names (using coredns deployed in the default overlay) but they cannot talk with the real gateway. when we disable u2o, the traffic with the gateway works, but of course we cannot reach overlay services.

I tried a stable 1.13x release and it seems to have the same issue, but i did very few tests there.

When using v1.12.6 both same scenarios work, so it looks like a regression.

changluyi:
We have made a change to drop the arp of u2o logical ip.
https://github.com/kubeovn/kube-ovn/pull/4150/files

patriziobassi:
Hi,

is there a reason for it? can we re-enable or workaround? because it completly broke our infra which is working in v1.12.6 (i saw it was merged in v1.12.17).

this explains why 1.13 and 1.14 are broken for us.

changluyi:
At that time, the u2o gatewayIP was exposed to the switch, which was equivalent to the switch receiving the same ARP from all ports connected to the node. Some switches with strict FDB configuration would alarm, thinking that the same MAC should not have multiple ports. In addition, the u2oInterconnect gateway was considered to be internal to the ovn system and should not be exposed to the outside.

changluyi:
[...]

PR #5419, opened 2025-07-02
controller: do not set mac binding age when the vpc is connected to e…

zhangzujian:
…xternal network

# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)

### DIFF
--- a/pkg/controller/vpc.go
+++ b/pkg/controller/vpc.go
@@ -268,7 +268,7 @@ func (c *Controller) handleAddOrUpdateVpc(key string) error {
 		return err
 	}
 
-	if err = c.createVpcRouter(key); err != nil {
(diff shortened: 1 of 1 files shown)


changluyi (approved):

zbb88888:
@zhangzujian 大佬，U2O 场景下会有同样的问题么？

patriziobassi:
@zhangzujian this will *not* fix the case where we have a underlay network attached to the default vpc (for u2o) as that network will no be marked as external (no need for nat gw), so bug https://github.com/kubeovn/kube-ovn/issues/5318 will not be fixed

zhangzujian:
@patriziobassi This patch is NOT a fix for the problem you metioned above.

PR #5504, opened 2025-07-22
panic install if labeled nodes are not found

oilbeater:
# Pull Request

## What type of this PR

- Bug fixes

Fixes #(issue-number)


zhangzujian (approved):

abhishek-pandey-1:
@oilbeater This change means that we can't use `helm template` or helm install with client-side dry-run with the kubeovn charts. Any thoughts on how we can fix that? 

I this it is acceptable for the function to not fail and just return nothing if the len of the nodes returned is zero. If we actually find nodes, then maybe what you have works fine? This will enable `helm template` and dry-run installs to work.

PR #5605, opened 2025-08-14
support for vpc nat gateway without eip

### LINKED ISSUE #5480: [Feature Request] Initialize VPC NAT Gateway without underlay IP
### Description

For historical reasons, the VPC NAT Gateway was initialized with a default underlay IP for egress. However, we later evolved to use EIPs for finer granular control while keeping the default underlay IP. Currently, the VPC NAT Gateway still requires an additional underlay IP, but no functionality actually depends on it, which wastes an underlay IP. We can modify the initialization logic to create the interface without an IP to optimize the implementation.

Whoever assigns public IPs to a VPC will save the IP for future use.

### Anything else?

_No response_

oilbeater:
# Pull Request

## What type of this PR

- Features

Fixes #5480

### DIFF
--- a/charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
+++ b/charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
@@ -506,6 +506,8 @@ spec:
                     type: string
                 qosPolicy:
                   type: string
+                noDefaultEIP:
+                  type: boolean
                 bgpSpeaker:
                   type: object
                   properties:
--- a/charts/kube-ovn/templates/kube-ovn-crd.yaml
+++ b/charts/kube-ovn/templates/kube-ovn-crd.yaml
@@ -506,6 +506,8 @@ spec:
                     type: string
                 qosPolicy:
                   type: string
+                noDefaultEIP:
+                  type: boolean
                 bgpSpeaker:
                   type: object
                   properties:
--- a/dist/images/install.sh
+++ b/dist/images/install.sh
@@ -751,6 +751,8 @@ spec:
                     type: string
                 qosPolicy:
                   type: string
+                noDefaultEIP:
+                  type: boolean
                 bgpSpeaker:
                   type: object
                   properties:
--- a/dist/images/vpcnatgateway/nat-gateway.sh
+++ b/dist/images/vpcnatgateway/nat-gateway.sh
@@ -137,7 +137,7 @@ function init() {
 
     # Send gratuitous ARP for all the IPs on the external network interface at initialization
     # This is especially useful to update the MAC of the nexthop we announce to the BGP speaker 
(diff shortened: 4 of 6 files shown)


SkalaNetworks:
I'm curious about this feature but I'm not I understand how it works. I'm also working on the GW right now.

What I understand:
Instead of passing the gateway through annotations, it is passed through env variables. The withoutEip flag only disables the annotation, but the bash script that sets the gateways is always triggered.

It also doesn't disable EIPs from being created on the gateway.

Is the goal to simply forward traffic? Can't the GW already do that?

oilbeater:
> Instead of passing the gateway through annotations, it is passed through env variables. The withoutEip flag only disables the annotation, but the bash script that sets the gateways is always triggered.

It took me a long journey to get this at least work. Here's what I tried:

When `withoutEIP` is true, I also set `annotations[fmt.Sprintf(util.AllocatedAnnotationTemplate, subnet.Spec.Provider)] = "true"` to prevent IPAM from allocating an address for this gateway on `net1`.

(Initially, I tried removing the IPAM config from the Multus config, but that didn't work due to some logical issues in the implementation, so I used this workaround instead.)

However, if `net1` has no IP address, Multus fails because it expects the IPAM plugin to return at least one IP. So when using this mode, still need to remove the IPAM config from the Multus configuration.

But removing the IPAM config means the route information for `net1` can't be passed to Multus. I then tried using Multus's `default-route` annotation.

The problem is that Multus requires `net1` to have an IP address before setting the `default-route`. In the end, I made it work by only setting the default route when an EIP is added to the gateway through env.

SkalaNetworks:
So the option ensures net1 has no IP at all? Does ARP work if it doesn't?

oilbeater:
> So the option ensures net1 has no IP at all? Does ARP work if it doesn't?

The user can still add EIPs to gw later through `IptablesEIP` CR, which is the common practice. In this use case, the default allocated IP address is unnecessary and goes to waste.

SkalaNetworks:
> The user can still add EIPs to gw later through IptablesEIP CR, which is the common practice. In this use case, the default allocated IP address is unnecessary and goes to waste.

True, the IP is wasted. That is unless the NAT GW is running in BGP mode. If that's the case, the default allocated IP is used as the nexthop, which is why I just submitted this PR to add gratuitous ARP on them (https://github.com/kubeovn/kube-ovn/pull/5607)

I'm a bit confused by the naming of the parameter, why is it called `withoutEIP`? Maybe something like `withoutExternalNetworkIP`? It's longer but if we look at the documentation, this IP is always part of the "externalNetwork".

oilbeater:
> If that's the case, the default allocated IP is used as the nexthop, which is why I just submitted this PR to add gratuitous ARP on them

For backward compatibility, this parameter defaults to false.

> I'm a bit confused by the naming of the parameter, why is it called withoutEIP? Maybe something like withoutExternalNetworkIP? 

There is still some ongoing work. I'll work on clarifying the naming to make it more intuitive.

Issue #5360, opened 2025-06-13
[Feature Request] Extend the capability of Kube-OVN for it to be used as a non-primary CNI plugin


starbops:
### Description

The [Harvester HCI project](https://github.com/harvester/harvester) is currently undergoing a significant integration effort to incorporate Kube-OVN into its ecosystem, providing users with advanced networking features ([ref](https://github.com/harvester/harvester/issues/7397)). However, not every part of the integration plan works as smoothly as expected. We have tried very hard to overcome these issues on our side, but unfortunately, they seem to require support or changes from the upstream, i.e., in the Kube-OVN project.

Currently, we've identified several areas that could pose issues for the integration effort. These are rooted in the premise that Kube-OVN is the primary CNI plugin in a Kubernetes cluster. To name a few:

#### ovn-cni initialization failure

This was described in #5078 in detail and has already been resolved.

#### Make the default Subnet `ovn-default` optional

Since the pod network is still using the CNI plugin other than Kube-OVN, it seems unnecessary to keep the `ovn-default` Subnet. Otherwise, all the pods running on the cluster consume IP addresses from the `ovn-default` Subnet (though it does no harm actually). For instance, the pod honors the IP allocation from Canal but not Kube-OVN, leaving those Kube-OVN-generated annotations ineffective (but still occupied an IP address from the Subnet):

```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    cni.projectcalico.org/containerID: 6859c02d1c3d49f74b7a4193a6f04d0b2e34a56bfa0b82ee5a0c791a3af477f0
    cni.projectcalico.org/podIP: 10.52.0.32/32   # ip allocated from canal
    cni.projectcalico.org/podIPs: 10.52.0.32/32
    k8s.v1.cni.cncf.io/network-status: |-
      [{
          "name": "k8s-pod-network",
          "ips": [
              "10.52.0.32"                       # the actual pod ip
          ],
          "default": true,
          "dns": {}
      }]
    ovn.kubernetes.io/allocated: "true"
    ovn.kubernetes.io/cidr: 10.52.0.0/16
    ovn.kubernetes.io/gateway: 10.52.0.1
    ovn.kubernetes.io/ip_address: 10.52.0.77     # not being used
    ovn.kubernetes.io/logical_router: ovn-cluster
    ovn.kubernetes.io/logical_switch: ovn-default
    ovn.kubernetes.io/mac_address: c6:f6:0b:86:67:10
[...]

PR #5618, opened 2025-08-18
Feat: Added non primary cni mode support for Kube-OVN

cleoz97:
# Pull Request

## What type of this PR

- Features

The changes enable handling of secondary IP addresses by updating endpoints, NAT gateway configuration, pod network extraction, and associated scripts when non-primary CNI mode is activated.

- endpoint_slice.go: Introduced a function to replace endpoint addresses with pod secondary IPs and added logic in endpoint update handlers to support non-primary CNI mode.
- vpc_nat_gateway.go: Updated NAT gateway initialization and StatefulSet generation to extract network interface names from annotations, configure interfaces for secondary CNI, and added a helper function for retrieving the subnet provider.
- nat-gateway.sh: Revised the NAT gateway script to support custom VPC and external interfaces through validated input and persistent configuration.
- pod.go: Added functionality to extract NAD interface names from network status annotations and adjusted pod network extraction to bypass default networks in non-primary CNI mode.
- config.go and const.go: Added a configuration flag for non-primary CNI mode and updated related annotation templates.

Attempt to support #5360

### DIFF
--- a/dist/images/vpcnatgateway/nat-gateway.sh
+++ b/dist/images/vpcnatgateway/nat-gateway.sh
@@ -2,9 +2,9 @@
 # Network interfaces configuration
 
 # Read interfaces from persistent file
-if [ -f /usr/bin/kube-ovn/nat-gateway.env ]; then
+if [ -f /etc/kube-ovn/nat-gateway.env ]; then
     # shellcheck disable=SC1091
-    source /usr/bin/kube-ovn/nat-gateway.env
+    source /etc/kube-ovn/nat-gateway.env
 fi
 # Default interfaces
 VPC_INTERFACE=${VPC_INTERFACE:-"eth0"}
@@ -78,19 +78,28 @@ function check_inited() {
 
 function init() {
     interfaces=$1
+    echo "init $interfaces"
     if [ -n "$interfaces" ]; then
+        # First, remove all spaces around commas
+        interfaces=$(echo "$interfaces" | sed 's/[[:space:]]*,[[:space:]]*/,/g')
         IFS=',' read -r -a interface_array <<< "$interfaces"
         if [ ${#interface_array[@]} -ne 2 ]; then
             >&2 echo "Error: Expected two interfaces separated by a comma (e.g., net1,net2)"
             exit 1
         fi
-        # Trim whitespace
-        VPC_INTERFACE="${interface_array[0]// /}"
-        EXTERNAL_INTERFACE="${interface_array[1]// /}"
+        
+        # Trim any remaining leading and trailing whitespace
+        VPC_INTERFACE=$(echo "${interface_array[0]}" | sed 's/^[[:space:]]*//;s/[[:space:]]*$//')
+        EXTERNAL_INTERFACE=$(echo "${interface_array[1]}" | sed 's/^[[:space:]]*//;s/[[:space:]]*$//')
+        
+        # Validate interface names are not empty after trimming
+        if [ -z "$VPC_INTERFACE" ] || [ -z "$EXTERNAL_INTERFACE" ]; then
+            >&2 echo "Error: Interface names cannot be empty"
+            exit 1
+        fi
     fi
     echo "using VPC interface: $VPC_INTERFACE"
     echo "using External interface: $EXTERNAL_INTERFACE"
-
     # check if interfaces exist
     if ! ip link show "$VPC_INTERFACE" >/dev/null 2>&1; then
         >&2 echo "Error: VPC interface '$VPC_INTERFACE' not found"
@@ -101,8 +110,9 @@ function init() {
         exit 1
     fi
     # Store interfaces persistently
-    echo "VPC_INTERFACE=$VPC_INTERFACE" > /usr/bin/kube-ovn/nat-gateway.env
-    echo "EXTERNAL_INTERFACE=$EXTERNAL_INTERFACE" >> /usr/bin/kube-ovn/nat-gateway.env
+    mkdir -p /etc/kube-ovn
+    echo "VPC_INTERFACE=$VPC_INTERFACE" > /etc/kube-ovn/nat-gateway.env
+    echo "EXTERNAL_INTERFACE=$EXTERNAL_INTERFACE" >> /etc/kube-ovn/nat-gateway.env
 
     # run once is enough
     $iptables_save_cmd | grep DNAT_FILTER && exit 0
@@ -493,9 +503,8 @@ opt=$1
 case $opt in
     init)
         # get user interfaces if provided from input
-        interfaces=$(echo $rules | awk '{print $1}')
-        echo "init $interfaces"
-        init $interfaces
+        interfaces="$rules"
+        init "$interfaces"
         ;;
     subnet-route-add)
         echo "subnet-route-add $rules"
--- a/dist/images/vpcnatgateway/nat-gateway.sh
+++ b/dist/images/vpcnatgateway/nat-gateway.sh
@@ -1,4 +1,9 @@
 #!/usr/bin/env bash
+# TODO: Interfaces used in this script are hard coded. Should we make them configurable?
+#       For now, we assume the following:
+#       - net1 is the interface that connects to the VPC
+#       - net2 is the interface that connects to the external network
+#       - A VPC gateway cannot be part of more than one VPC and external network
 
 iptables_cmd=$(which iptables)
 iptables_save_cmd=$(which iptables-save)
@@ -67,7 +72,7 @@ function add_vpc_internal_route() {
         cidr=${arr[0]}
         nextHop=${arr[1]}
 
-        exec_cmd "ip route replace $cidr via $nextHop dev eth0"
+        exec_cmd "ip route replace $cidr via $nextHop dev net1"
     done
 }
 
@@ -79,7 +84,7 @@ function del_vpc_internal_route() {
         arr=(${rule//,/ })
         cidr=${arr[0]}
 
-        exec_cmd "ip route del $cidr dev eth0"
+        exec_cmd "ip route del $cidr dev net1"
     done
 }
 
@@ -92,9 +97,9 @@ function del_vpc_external_route() {
     # do
     #     arr=(${rule//,/ })
     #     cidr=${arr[0]}
-    #     exec_cmd "ip route del $cidr dev net1"
+    #     exec_cmd "ip route del $cidr dev net2"
     #     sleep 1
-    #     exec_cmd "ip route del default dev net1"
+    #     exec_cmd "ip route del default dev net2"
     # done
 }
 
@@ -117,9 +122,9 @@ function del_eip() {
     do
         arr=(${rule//,/ })
         eip=${arr[0]}
-        ipCidr=`ip addr show net1 | grep $eip | awk '{print $2 }'`
+        ipCidr=`ip addr show net2 | grep $eip | awk '{print $2 }'`
         if [ -n "$ipCidr" ]; then
-            exec_cmd "ip addr del $ipCidr dev net1"
+            exec_cmd "ip addr del $ipCidr dev net2"
         fi
     done
 }
@@ -169,7 +174,7 @@ function add_snat() {
         randomFullyOption=${arr[2]}
         # check if already exist
         $iptables_save_cmd | grep SHARED_SNAT | grep "\-s $internalCIDR" | grep "source $eip" && exit 0
-        exec_cmd "$iptables_cmd -t nat -A SHARED_SNAT -o net1 -s $internalCIDR -j SNAT --to-source $eip $randomFullyOption"
+        exec_cmd "$iptables_cmd -t nat -A SHARED_SNAT -o net2 -s $internalCIDR -j SNAT --to-source $eip $randomFullyOption"
     done
 }
 function del_snat() {
@@ -230,14 +235,14 @@ function del_dnat() {
 
 
 # example usage:
-# delete_tc_u32_filter "net1" "1:0" "192.168.1.1" "src"
+# delete_tc_u32_filter "net2" "1:0" "192.168.1.1" "src"
 function delete_tc_u32_filter() {
     dev=$1
     qdisc_id=$2
     cidr=$3
     matchDirection=$4
 
-    # tc -p -s -d filter show dev net1 parent $qdisc_id
+    # tc -p -s -d filter show dev net2 parent $qdisc_id
     # filter protocol ip pref 10 u32 chain 0
     # filter protocol ip pref 10 u32 chain 0 fh 800: ht divisor 1
     # filter protocol ip pref 10 u32 chain 0 fh 800::800 order 2048 key ht 800 bkt 0 *flowid :1 not_in_hw
@@ -265,8 +270,8 @@ function delete_tc_u32_filter() {
 
 function eip_ingress_qos_add() {
     # ingress:
-    # external --> net1 --> qos -->
-    # dst ip is iptables eip on net1
+    # external --> net2 --> qos -->
+    # dst ip is iptables eip on net2
     for rule in $@
     do
         arr=(${rule//,/ })
@@ -274,7 +279,7 @@ function eip_ingress_qos_add() {
         priority=${arr[1]}
         rate=${arr[2]}
         burst=${arr[3]}
-        dev="net1"
+        dev="net2"
         matchDirection="dst"
         tc qdisc add dev $dev ingress 2>/dev/nul || true
         # get qdisc id
@@ -290,8 +295,8 @@ function eip_ingress_qos_add() {
 
 function eip_egress_qos_add() {
     # egress:
-    # net1 --> qos --> external
-    # src ip is iptables eip on net1
+    # net2 --> qos --> external
+    # src ip is iptables eip on net2
     for rule in $@
     do
         arr=(${rule//,/ })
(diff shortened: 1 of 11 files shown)


pkg/controller/endpoint_slice.go
```diff
@@ -131,6 +131,22 @@ func (c *Controller) handleUpdateEndpointSlice(key string) error {
...
+	// If Kube-OVN is running in secondary CNI mode, the endpoint IPs should be derived from the network attachment definitions
+	// This overwrite can be removed if endpoint construction accounts for network attachment IP address
+	// TODO: Identify how endpoints are constructed, by default, endpoints has IP address of eth0 inteface
+	if c.config.EnableNonPrimaryCNI {
+		var pods []*v1.Pod
+		if pods, err = c.podsLister.Pods(namespace).List(labels.Set(svc.Spec.Selector).AsSelector()); err != nil {
+			klog.Errorf("failed to get pods for service %s in namespace %s: %v", name, namespace, err)
+			return err
+		}
+		err = c.replaceEndpointAddressesWithSecondaryIPs(endpointSlices, pods)
```
oilbeater, line 143: We need to make a deepCopy of the endpointSlices; otherwise, the Informer might change them asynchronously.

pkg/controller/endpoint_slice.go
```diff
@@ -131,6 +131,22 @@ func (c *Controller) handleUpdateEndpointSlice(key string) error {
 		}
 	}
 
+	// If Kube-OVN is running in secondary CNI mode, the endpoint IPs should be derived from the network attachment definitions
```
oilbeater, line 134: Do you have any plans to create a separate controller for generating EndpointSlices for network attachments? I believe this could benefit the ecosystem beyond just Kube-OVN users.

pkg/controller/pod.go
```diff
@@ -2396,3 +2430,30 @@ func setPodRoutesAnnotation(annotations map[string]string, provider string, rout
 
 	return nil
 }
+
+func (c *Controller) getNadInterfaceFromNetworkStatusAnnotation(networkStatus string, nadName string) (string, error) {
```
oilbeater: This can be moved to `util/network_attachment.go` and use the [NetworkStatus Type](https://github.com/k8snetworkplumbingwg/network-attachment-definition-client/blob/master/pkg/apis/k8s.cni.cncf.io/v1/types.go#L105C6-L105C19)

oilbeater:
@cleoz97 Thanks for this, please rebase master and run `make lint` to resolve the ci issues.

dist/images/vpcnatgateway/nat-gateway.sh
```diff
@@ -1,4 +1,14 @@
 #!/usr/bin/env bash
+# Network interfaces configuration
+
+# Read interfaces from persistent file
+if [ -f /usr/bin/kube-ovn/nat-gateway.env ]; then
+    # shellcheck disable=SC1091
+    source /usr/bin/kube-ovn/nat-gateway.env
```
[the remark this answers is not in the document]
SkalaNetworks: Probably relevant, but a nit

dist/images/vpcnatgateway/nat-gateway.sh
```diff
@@ -8,6 +18,46 @@ if iptables-legacy -t nat -S INPUT 1 2>/dev/null; then
...
+    echo "  qos-del                  - Delete QoS rule"
+    echo "  eip-ingress-qos-add      - Add EIP ingress QoS"
+    echo "  eip-egress-qos-add       - Add EIP egress QoS"
+    echo "  eip-ingress-qos-del      - Delete EIP ingress QoS"
+    echo "  eip-egress-qos-del       - Delete EIP egress QoS"
+    echo "  get-iptables-version     - Show iptables version"
+    echo ""
+    echo "Examples:"
+    echo "  # Use custom interfaces"
+    echo "  $0 init net1, net2"
```
SkalaNetworks, line 52: Isn't it "net1,net2" and not "net1, net2" 

per your script that handles splitting the arg

dist/images/vpcnatgateway/nat-gateway.sh
```diff
@@ -27,6 +77,33 @@ function check_inited() {
...
 function init() {
+    interfaces=$1
+    if [ -n "$interfaces" ]; then
+        IFS=',' read -r -a interface_array <<< "$interfaces"
+        if [ ${#interface_array[@]} -ne 2 ]; then
+            >&2 echo "Error: Expected two interfaces separated by a comma (e.g., net1,net2)"
+            exit 1
+        fi
+        # Trim whitespace
+        VPC_INTERFACE="${interface_array[0]// /}"
```
SkalaNetworks: Can you get whitespaces? Considering the script is bash, unless the parameter is quoted, won't an input like 

init net1, net2 

yield:
 firstparam=net1,
 secondparam=net2

I tested in a sandbox and this is what it did

SkalaNetworks: Maybe just make it 2 distinct parameter and don't bother with the ","?

dist/images/vpcnatgateway/nat-gateway.sh
```diff
@@ -411,8 +488,10 @@ rules=${@:2:${#}}
 opt=$1
 case $opt in
     init)
-        echo "init $rules"
-        init $rules
```
SkalaNetworks, line 419: Interesting to see we passed rules that weren't used in the init function

pkg/controller/config.go
```diff
@@ -212,6 +215,8 @@ func ParseFlags() (*Configuration, error) {
 		argTLSMinVersion   = pflag.String("tls-min-version", "", "The minimum TLS version to use for secure serving. Supported values: TLS10, TLS11, TLS12, TLS13. If not set, the default is used based on the Go version.")
 		argTLSMaxVersion   = pflag.String("tls-max-version", "", "The maximum TLS version to use for secure serving. Supported values: TLS10, TLS11, TLS12, TLS13. If not set, the default is used based on the Go version.")
 		argTLSCipherSuites = pflag.StringSlice("tls-cipher-suites", nil, "Comma-separated list of TLS cipher suite names to use for secure serving (e.g., 'TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384'). Names must match Go's crypto/tls package. See Go documentation for available suites. If not set, defaults are used. Users are responsible for selecting secure cipher suites.")
+
+		argNonPrimaryCNI = pflag.Bool("non-primary-cni-mode", false, "Use Kube-OVN in non primary cni mode. When true, Kube-OVN will only manage the network for network attachment definitions")
```
SkalaNetworks, line 219: Could you edit the helm charts in both versions with a new value to enable/disable this behaviour?

pkg/controller/pod.go
```diff
@@ -513,8 +519,21 @@ func (c *Controller) handleAddOrUpdatePod(key string) (err error) {
 		}
 	}
 
-	if vpcGwName, isVpcNatGw := pod.Annotations[util.VpcNatGatewayAnnotation]; isVpcNatGw {
+	providerName := util.OvnProvider
+	// In non-primary CNI mode, we extract the provider from the pod's annotation.
```
SkalaNetworks: nit: maybe factorize this as you're using it below in an identical form

pkg/controller/endpoint_slice.go
cleoz97: is it not guaranteed every time informer changes endpoint slices, handleUpdateEndpointSlice is called?

pkg/controller/endpoint_slice.go
cleoz97: Amazing thought.

At the moment, we have nothing decided to watch for such IPs, but i think it could be a very useful feature since today endpoints do not have secondary IPs.

I will get back to you on it

dist/images/vpcnatgateway/nat-gateway.sh
cleoz97: fixed script to handle various input formats

dist/images/vpcnatgateway/nat-gateway.sh
cleoz97: I have fixed the script to handle various valid input strings.

parameter-based inputs will have changes elsewhere too, making it difficult to maintain changes.
New changes should account for all cases

pkg/controller/pod.go
cleoz97: done

dist/images/vpcnatgateway/nat-gateway.sh
cleoz97: yes, I guess it's for consistency between all calls

pkg/controller/config.go
cleoz97: done

pkg/controller/vpc_nat_gateway.go
```diff
@@ -611,13 +660,17 @@ func (c *Controller) handleUpdateNatGwSubnetRoute(natGwKey string) error {
...
+	// For each subnet provider, generate vpc cidr annotation
+	patch := util.KVPatch{}
+	for provider, cidrs := range newProviderCIDRMap {
+		cidrBytes, err := json.Marshal(cidrs)
+		if err != nil {
+			klog.Errorf("marshal eip annotation failed %v", err)
+			return err
+		}
+		patch[fmt.Sprintf(util.VpcCIDRsAnnotationTemplate, provider)] = string(cidrBytes)
 	}
```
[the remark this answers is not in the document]
cleoz97: now handling stale vpc cidr annotations

pkg/controller/endpoint_slice.go
```diff
@@ -257,6 +273,71 @@ func (c *Controller) handleUpdateEndpointSlice(key string) error {
...
+	replacedEndpoints := make(map[string]bool)
+	// iterate over all endpoint slices
+	for i, endpoint := range endpointSlice {
+		// iterate over all endpoints in the slice
+		for j, ep := range endpoint.Endpoints {
+			if ep.TargetRef != nil && ep.TargetRef.Kind == "Pod" {
+				podName := ep.TargetRef.Name
+				for _, pod := range pods {
+					// Find the matching pod
+					if podName == pod.Name {
```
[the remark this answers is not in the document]
cleoz97: optimized using map

dist/images/vpcnatgateway/nat-gateway.sh
```diff
@@ -27,6 +77,44 @@ function check_inited() {
...
+        
+        # Validate interface names are not empty after trimming
+        if [ -z "$VPC_INTERFACE" ] || [ -z "$EXTERNAL_INTERFACE" ]; then
+            >&2 echo "Error: Interface names cannot be empty"
+            exit 1
+        fi
+    fi
+    echo "using VPC interface: $VPC_INTERFACE"
+    echo "using External interface: $EXTERNAL_INTERFACE"
+    exit 0
```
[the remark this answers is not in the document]
cleoz97: removed debug code

pkg/controller/endpoint_slice.go
cleoz97: on after thought are we talking about race condition here?

oilbeater: When you replace the cache in the informer, other handlers might read the updated endpoint, which may not be what they expect. Additionally, the informer could modify the endpoint after you've replaced it but before you actually use it.

SkalaNetworks:
@cleoz97 Hi, please take a look at the CI.

 Step build kube-ovn seems to fail. 

Running `make lint` should remove errors about formatting.

pkg/controller/endpoint_slice.go
cleoz97: now using deep copy when we replace endpoint IPs

cleoz97:
> @cleoz97 Hi, please take a look at the CI.
> 
> Step build kube-ovn seems to fail.
> 
> Running `make lint` should remove errors about formatting.

I have fixed the lint errors scoped to my changes. However, I do see other lint errors being reported in other files, I'm not sure what is causing those

oilbeater:
@cleoz97 I fixed the lint issue to make the E2E run. And the vpc nat gateway e2e failed, please take a look.

pkg/controller/endpoint_slice.go
```diff
@@ -257,6 +273,77 @@ func (c *Controller) handleUpdateEndpointSlice(key string) error {
...
+					found := false
+					for k, address := range ep.Addresses {
+						// skip if this specific endpoint has already been replaced
+						if replacedEndpoints[address] {
+							continue
+						}
+						// Endpoints received here are from cache and can be either primary or secondary network IP address
+						// Replace the primary network IP address with the secondary network IP address
+						// Secondary IP address is already seen when the endpoint has been already created/updated
+						if address == pod.Status.PodIP || address == ipAddress {
```
nijosmsft: why do we need to recopy if its already equal? can we just break?

pkg/controller/endpoint_slice.go
```diff
@@ -257,6 +273,77 @@ func (c *Controller) handleUpdateEndpointSlice(key string) error {
...
+					ipAddress := pod.Annotations[fmt.Sprintf(util.IPAddressAnnotationTemplate, providers[0])]
+					if ipAddress == "" {
+						klog.Errorf("failed to get pod %s/%s ip address", pod.Namespace, pod.Name)
+						return fmt.Errorf("failed to get pod %s/%s ip address", pod.Namespace, pod.Name)
+					}
+					// Update the matching endpoint with the secondary IP address
+					found := false
+					for k, address := range ep.Addresses {
+						// skip if this specific endpoint has already been replaced
+						if replacedEndpoints[address] {
```
nijosmsft: why is this needed?

pkg/controller/pod.go
```diff
@@ -2396,3 +2408,34 @@ func setPodRoutesAnnotation(annotations map[string]string, provider string, rout
...
+		return false, ""
+	}
+	if pod.Annotations == nil {
+		return false, ""
+	}
+	// default provider
+	providerName := util.OvnProvider
+	// get providers
+	providers, err := c.getPodProviders(pod)
+	if err != nil {
```
nijosmsft: should we use default provider instead of returning error?

nijosmsft:
hi @oilbeater, thanks a lot for reviewing the changes. I have a question about test requirements. We will be adding E2E tests to validate the scenarios. Other than that, are there any test requirements we should be aware of?

pkg/controller/pod.go
```diff
@@ -2396,3 +2408,34 @@ func setPodRoutesAnnotation(annotations map[string]string, provider string, rout
...
+		klog.Errorf("failed to get pod %s/%s providers, %v", pod.Namespace, pod.Name, err)
+		return false, ""
+	}
+	if len(providers) > 0 {
+		// use the first provider
+		providerName = providers[0]
```
nijosmsft: In `getPodKubeovnNets`, `attachmentNets` are added first then default ovn. As we are using `providers[0]` I think when secondary CNI mode is not enabled, `providerName` will end up as the macvlan NAD?

pkg/controller/pod.go
nijosmsft: > should we use default provider instead of returning error?

I am thinking from a regression pov. Another thought, should we just use `util.OvnProvider` if c.config.EnableNonPrimaryCNI == false ?

oilbeater:
> hi @oilbeater, thanks a lot for reviewing the changes. I have a question about test requirements. We will be adding E2E tests to validate the scenarios. Other than that, are there any test requirements we should be aware of?

There are no specific test requirements beyond that. You can start by covering the basic usage scenarios and add more test cases later based on any issues we find.

pkg/controller/endpoint_slice.go
cleoz97: refactored the code for clarity

pkg/controller/pod.go
cleoz97: we now use util.OvnProvider as default. but we still return error if no providers are found in non primary cni mode

pkg/controller/pod.go
cleoz97: we only call getPodKubeovnNets when it is running as non primary cni mode

cleoz97:
@oilbeater do we expect any failure to e2e tests?
Last 2 runs have been a deployment failure in different test cases.

Is there a cmd to run all e2e testcases?

cleoz97:
@oilbeater

I could not reproduce this locally for this specific test case using the same make commands
I could also see that the failure is due to kube-ovn-cni pod ovn0 address not being allocated, however other test cases seem to pass

Any suggestions?

cleoz97:
@oilbeater looks like all test passed in last 2 runs, thanks for the retrigger

For this PR to be merged, what other requirements must be satisfied? Do we want to include e2e tests along this PR or a separate one

We have few unit and e2e tests written

oilbeater:
@cleoz97 Thanks for the update! I think we should also update the documentation to introduce this feature, but both docs and tests can be handled in separate PRs. This PR is ready to merge as is.

Yu-Jack:
@oilbeater May I know which release version will include this feature? Thanks.

oilbeater:
@Yu-Jack all the features added to the main branch will only be included in the next minor release, which in this case is v1.15.0. We don't have a fixed release schedule yet, and to be frank, the official release isn't that significant if the feature isn't widely used by the community.  If you're relying on this feature, I recommend testing it now to see if it meets your needs and to check for any bugs or limitations. That way, we can make adjustments before the official release.

Issue #5470, opened 2025-07-11
[Feature Request] VPC edge router


abasitt:
### Description

I have been playing with VPC egress gateway, IMO it's a great feature and i love the BFD and HA features in it. 

Can we expand the same egress gateway and have an option to run it as edge router?

Why edge router?
- We want to natively advertise subnets from vpc to a network in a different VRF.
- Accept network routes in to edge router. This will need a new controller in the bgp speaker to push the learn routes in to the egress gateway fib
- Push learned routes in the egress gateway in to the VPC. This will need a new controller in the bgp speaker to push learn routes in to the vpc routing table.

Hope this make sense ? Thank you.

oilbeater:
Thanks for the great idea! I like the concept, but egress gateway is currently focused on pod egress control. The edge router functionality would be better served by a new dedicated CRD that could reuse some egress gateway logic. Would you be interested in contributing this feature?

abasitt:
I am thinking about working on it. Looking at the code, I think there are two possible ways to handle it.

- New CRDS and a logic of creating edge-router similar to egress gateway. This sounds simple but we may also end up with duplicate code.
- Modify egress gateway logic little and add new CR with parentref to configure bgp. 

Any thoughts on implementation @oilbeater @zhangzujian ?

donghee-park-rsk:
Hello, I implemented this feature CR bgp-edge-router, bgp-edge-router-advertisements and gobgp-configs.
Can you look and give an advice @oilbeater @zhangzujian ?
Here is the code : https://github.com/donghee-park-rsk/kube-ovn/tree/edge-router-dev
Branch : edge-router-dev

oilbeater:
@donghee-park-rsk Thanks for reaching out to us. Since this is a long series of commits, could you please write a document about the design and usage? That would make it easier for us to review.

donghee-park-rsk:
@oilbeater Sure, when I complete the docs, i will share it :)

donghee-park-rsk:
@oilbeater Here is the document for bgp-edge-router what I implemented. Please check and feel free to give a feedback.
https://github.com/kubeovn/docs/blob/4593ffe7ab9c0fdff963825343f4c9b65c7ade19/docs/vpc/bgp-edge-router.en.md

oilbeater:
[...]

PR #5717, opened 2025-09-11
feat(BGP edge router) : BGP per VPC

donghee-park-rsk:
# Pull Request

## What type of this PR

- Features
   BGP edge router allowing bgp per VPC
- Docs
   https://github.com/kubeovn/docs/blob/4593ffe7ab9c0fdff963825343f4c9b65c7ade19/docs/vpc/bgp-edge-router.en.md

Feature Request #5470


oilbeater:
@donghee-park-rsk Recently, I was investigating how to integrate Kube-OVN with EVPN and found that EVPN is mostly BGP with Vxlan. When I looked into GoBGP, I discovered that it only handles control plane information, whereas FRR can also handle part of the data plane information, like the Linux FDB and IP routes. So, I think FRR would be a better solution for BGP to reduce the coding required on the Kube-OVN side. What do you think about it?

donghee-park-rsk:
> @donghee-park-rsk Recently, I was investigating how to integrate Kube-OVN with EVPN and found that EVPN is mostly BGP with Vxlan. When I looked into GoBGP, I discovered that it only handles control plane information, whereas FRR can also handle part of the data plane information, like the Linux FDB and IP routes. So, I think FRR would be a better solution for BGP to reduce the coding required on the Kube-OVN side. What do you think about it?

@oilbeater I agree with you. 
The reason I use GoBGP was, I only want to bgp feature. If one pod has VPN and BGP features then using FRR would be better choice like your opinion.

PR #5741, opened 2025-09-20
fix(netpol): don't add default block twice for dualstacks

SkalaNetworks:
# Pull Request

## What type of this PR

- Bug fixes
- Tests

When running dualstack, netpols add the "block" default ACL for both protocols (v6/v4), leading to error messages when functions check if there's duplicates.

Instead of adding the ACL each time, we now only do it only once for egress and ingress.

Fixes #(5736)


oilbeater:
One IPv6 networkpolicy test case keeps failing, please take a look.

SkalaNetworks:
I won't be able to look into this for the next 2 weeks but I'll do it as soon as I get the chance.

It is strange that it only fails for IPv6 and on the underlay, I'll triple check everything.

SkalaNetworks:
@oilbeater I don't know what's happening there but I cannot get this to reproduce on my env. Could you try cloning the branch and executing the tests on your side please?

I did confirm that the tests on Github succeed if the policy doesn't exist, so that's definitely the cause. But when running the E2E tests on my side with IPv6 only, the tests are successful.

EDIT: I merged master into the PR and somehow the faulty test is gone, but now netpols don't seem to block anything. Very strange.

oilbeater:
I will take a look.

SkalaNetworks:
Looks like the default drop doesn't get applied anymore if the policy of the NetworkPolicy is not both Egress/Ingress.

I may have found an issue with the check logic that doesn't generate ACL ops if something is already present. I'm trying a fix.

SkalaNetworks:
@oilbeater Looking at the test, I don't get how it should even pass (but somehow, locally, it does).

We're creating a network policy to allow only traffic to all IPv4s (0.0.0.0/0). But we're trying to fetch the API server with IPv6.

The description of the test mentions it should run as "hostNetwork"

 > should be able to access svc with backend host network pod after any other ingress network policy rules created

But the definition of the Pod doesn't have any setting.

EDIT: Looking at the behaviour of the ACLs, there's ACLs auto-added to allow communication between the nodes and the pods to conform with Kubernetes rules (and to allow healthchecks). If the API server runs on one of these nodes (which is the case with kind), the traffic will be let go.

SkalaNetworks:
Is there a way for me to re-launch only one of the steps of the GitHub pipeline? The issue is (for me) only present on the GH env and I'd like to try to reproduce it somehow.

oilbeater:
> Is there a way for me to re-launch only one of the steps of the GitHub pipeline? The issue is (for me) only present on the GH env and I'd like to try to reproduce it somehow.

Currently, there's no simple way to rerun just one step. You can try removing other E2E steps in `/.github/workflows/build-x86-image.yaml` first to isolate the issue.

SkalaNetworks:
I'm out of ideas, the tests are flaking and have 0 consistent behaviour.

It's always the underlay that fails, sometimes IPv6, sometimes IPv4, sometimes DualStack. I can't get a single failure on my local env in any of these scenarios.

Could it be that those tests have always been flaky and because I'm modying this part of the code it shows up?

SkalaNetworks:
it works when hostnetwork is true like the title of the test says, but not without it...

SkalaNetworks:
Looking at the logs of the controller during the GH test

image


It looks like the E2E test is broken somehow when working in underlay mode. The provider network is not found?

pkg/controller/network_policy.go
```diff
@@ -174,6 +174,13 @@ func (c *Controller) handleUpdateNp(key string) error {
 
 	if hasIngressRule(np) {
 		for _, protocol := range protocolSet.List() {
+			blockACLOps, err := c.OVNNbClient.UpdateDefaultBlockACLOps(key, pgName, ovnnb.ACLDirectionToLport, logEnable)
```
oilbeater, line 177: I think we can move this part out of the for loop to make sure it only be added once.

SkalaNetworks: That's the whole point of my PR, but it didn't work when out of the loop. I just found out why.

SOmetimes, the E2E tests generate netpols with selectors that don't match anything. When that's th case, the "protocolSet" list is empty, and the DefaultBlock is not added. 

This is actually the behaviour  we want, but it's a side-effect.

My implementation adds the block in any possible case, even when the portgroup is inexistant. And that seems to make everything explode. I'm pushing a fix right now.


This was pretty hard to debug because I couldn't easily see at first that the cases that were blowing up in the pipeline were due to selectors that didn't match anything?

SkalaNetworks: @oilbeater tests passed 3 times in a row when avoiding to put the blocking ACL on empty portgroups. So I think this problem is fixed and the PR can be merged.

PR #5745, opened 2025-09-22
feat(netpol): authorize l3 protocols

SkalaNetworks:
# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)

### DIFF
--- a/pkg/ovs/ovn-nb-acl.go
+++ b/pkg/ovs/ovn-nb-acl.go
@@ -54,7 +54,7 @@ func setACLName(acl *ovnnb.ACL, name string) {
 }
 
 // UpdateDefaultBlockACLOps returns operations to update/create the default block ACL
-func (c *OVNNbClient) UpdateDefaultBlockACLOps(npName, pgName, direction string, loggingEnabled bool) ([]ovsdb.Operation, error) {
+func (c *OVNNbClient) UpdateDefaultBlockACLOps(npName, pgName, direction string, loggingEnabled, lax bool) ([]ovsdb.Operation, error) {
 	portDirection := "outport"
 	priority := util.IngressDefaultDrop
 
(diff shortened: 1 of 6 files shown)


SkalaNetworks:
I'll wait for https://github.com/kubeovn/kube-ovn/pull/5741 to be merged to factorize the code

I'll also add a global switch to enable this or to keep the default behaviour. And also an annotation on NetworkPolicies to act as a toggle.

SkalaNetworks:
@oilbeater new PR to address the discussion we had on Slack

There's now a global controller switch and an annotation on network policies to configure the enforcement of network policies.

When the enforcement is "standard", everything works just like it always has: only what is in the policy is allowed, the rest is blocked (including ICMP)

When the enforcement is "lax", some protocols are excluded from being blocked. Those protocols are considered somewhat essential in a traditional network (especially one that has VMs in it): ARP/NDP (through ICMPv6)/ICMP/DHCP.

That setting is especially useful when using VMs through Kubevirt, as they may rely on DHCP to get their MTU/IP/DNS and more from Kubevirt or Kube-OVN. ICMP is unnecessary in the standard Kubernetes networking model, but very important for VMs to do PMTU discovery.

pkg/ovs/ovn-nb-acl.go
```diff
@@ -99,6 +99,87 @@ func (c *OVNNbClient) UpdateDefaultBlockACLOps(netpol, pgName, direction string,
 	return ops, nil
 }
 
+// UpdateDefaultBlockExceptionsACLOps updates the exceptions to the default block ACLs of a NetworkPolicy
+// to allow ARP/ICMP/DHCP. We may allow these protocols because the standard for NetworkPolicies expects to only block
+// based on the IP/TCP/UDP/SCTP protocols.
+func (c *OVNNbClient) UpdateDefaultBlockExceptionsACLOps(npName, pgName, npNamespace, direction, protocol string) ([]ovsdb.Operation, error) {
```
oilbeater: Based on the supported predicates listed in the OVN documentation https://man7.org/linux/man-pages/man5/ovn-sb.5.html#ACL_TABLE (search `The following predicates are supported:`), I think we can simplify our default deny flow to just `(tcp || udp || sctp)`. For DHCP, we could use `udp.src == 67 && udp.dst == 68` for to-lport and the reverse for from-lport. This approach would eliminate the need for separate rules for ARP, ICMP, and different IP versions (IPv4/IPv6/Dualstack).

oilbeater (changes requested):

pkg/ovs/ovn-nb-acl.go
SkalaNetworks: Good idea, we'll still need DHCPv6 which uses different ports.

oilbeater: I didn't realize DHCP uses different ports for IPv4 and IPv6. We'll need to add separate rules for both, but we can simplify by including both protocols regardless of the Pod IP protocol. This shouldn't impact functionality and will make the code cleaner.

SkalaNetworks: Sure we can do that. Gotta point out that theoretically, the UDP ports could be used for other purposes than DHCP by bad actors. For example, to exfiltrate data. But at that point, I think using an enforcement called Lax falls on the responsibility of the user.

SkalaNetworks: @oilbeater here you go, latest commit only blocks tcp/udp/sctp in Lax enforcement, and excludes DHCPv4/v6

PR #5775, opened 2025-10-08
fix: delete all ips before deleting a subnet

tallquist10:
# Pull Request

## What type of this PR

- Bug fixes

Fixes #(issue-number)

## Additional details
When a Subnet CR is created with one or more IP CRs, and then the resources are eventually deleted, instances where the subnet is deleted before the finalizers are removed from the IPs cause a failure, and the IPs are left unable to be deleted due to the absence of the corresponding Subnet. With kube-ovn being heavily reliant on the ability for these resources to be able to be deleted simultaneously, this seems to be counterintuitive to that philosophy.


pkg/controller/subnet.go
```diff
@@ -1008,8 +1008,28 @@ func (c *Controller) handleDeleteSubnet(subnet *kubeovnv1.Subnet) error {
...
+		if ip.Spec.Subnet == subnet.Name {
+			// Delete with foreground propagation to wait for finalizer cleanup
+			deletePolicy := metav1.DeletePropagationForeground
+			if err := c.config.KubeOvnClient.KubeovnV1().IPs().Delete(
+				context.Background(),
+				ip.Name,
+				metav1.DeleteOptions{PropagationPolicy: &deletePolicy},
+			); err != nil && !k8serrors.IsNotFound(err) {
+				klog.Errorf("failed to delete ip %s: %v", ip.Name, err)
+				return err
```
zbb88888, line 1027: how about just logging error and continue?

tallquist10: Yeah, sorry about this confusion. I will put together a more thorough PR description today. Basically, the issue is that IPs aren't getting deleted before their subnets, so when the controller goes to remove the finalizer, and the subnet doesn't exist, it fails and doesn't remove the finalizer, and it never proceeds. I forgot I made a draft PR with my changes, and this will not be the final version.

tallquist10:
Closing, as this seems to have been handled with GC in 1.15.x

PR #5804, opened 2025-10-17
Add support for optional node selector labels for kube-ovn OVS/OVN pods

saeed-mcu:
This change introduces a new Helm value ovsNodesLabels that allows adding custom node selector labels to ovs-ovn pods. For example:
 ovsNodesLabels:
  kube-ovn/role: ovs
This is useful in environments such as OpenStack-Helm, where operators may want to schedule OVS/OVN pods only on specific nodes (e.g., compute nodes) instead of deploying them cluster-wide. When ovsNodesLabels is empty, only the default selector kubernetes.io/os: linux is applied.

### DIFF
--- a/charts/kube-ovn-v2/templates/agent/agent-daemonset.yaml
+++ b/charts/kube-ovn-v2/templates/agent/agent-daemonset.yaml
@@ -231,6 +231,9 @@ spec:
         {{- end }}
       nodeSelector:
         kubernetes.io/os: "linux"
+        {{- with .Values.ovsNodesLabels }}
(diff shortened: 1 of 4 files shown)


SkalaNetworks:
Shouldn't those labels be added to every daemon deployed and not just the ones you put? For example, the controller?

If not, there's an ovsOvn section in the chart where that value could probably fit better. Renaming it to "nodeSelector" could also help make its goal clearer.

SkalaNetworks:
@saeed-mcu are you still working on this?

PR #5813, opened 2025-10-21
Remove negative matching for ipblock except rules.

oilbeater:
# Pull Request

## What type of this PR

- Performance

### Problem

**Issue 1**: The original implementation used a single ACL rule to match both IPBlock CIDR and except ranges with `!= $except`, which caused OVS flow table bloat.

**Issue 2**: When a Pod IP matched by PodSelector also fell within an IPBlock except range, it would be incorrectly dropped, violating the NetworkPolicy "OR" semantics where any matching rule should allow traffic.

### Solution

Separate NetworkPolicy rules by type with different ACL priorities:

- **Priority 2003**: PodSelector/NamespaceSelector IPs → allow-related
- **Priority 2002**: IPBlock except IPs → drop  
- **Priority 2001**: IPBlock CIDR → allow-related

### Changes

1. Split `fetchPolicySelectedAddresses` to return three separate lists: `selectorIPs`, `blockCIDRs`, and `exceptCIDRs`
2. Create three address sets per rule: selector, ipblock, and except
3. Generate ACLs with proper priorities ensuring selector-matched IPs are always allowed
4. Add new priority constants: `IngressSelectorAllowPriority` and `EgressSelectorAllowPriority`

### Benefits

- ✅ Eliminates `!=` operator usage, reducing OVS flow table size
- ✅ Ensures correct NetworkPolicy semantics (OR relationship between rules)
- ✅ PodSelector-matched IPs always allowed regardless of IPBlock except ranges

---

### DIFF
--- a/mocks/pkg/ovs/interface.go
+++ b/mocks/pkg/ovs/interface.go
@@ -2094,33 +2094,33 @@ func (mr *MockACLMockRecorder) UpdateDefaultBlockExceptionsACLOps(npName, pgName
 }
 
 // UpdateEgressACLOps mocks base method.
-func (m *MockACL) UpdateEgressACLOps(pgName, asEgressName, asExceptName, protocol, aclName string, npp []v10.NetworkPolicyPort, logEnable bool, logACLActions []ovnnb.ACLAction, namedPortMap map[string]*util.NamedPortInfo) ([]ovsdb.Operation, error) {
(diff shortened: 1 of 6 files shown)


oilbeater:
If two networkpolicy overlapping the CIDR and Except field, we should allow the traffic and this method can not handle this situation.

PR #5834, opened 2025-10-28
OVN-Interconnect: Only set-up LRP/LSP between the transit switches and default router if we want to.

andrewlee1089:
OVN-Interconnect: Only set-up LRP/LSP between the transit switches and default router if we want to.

This allows people to create transit switches and only have non-default routers connected to them

# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)


andrewlee1089:
Hi all - this is small feature change I need for my work prototyping having OVN interconnect with kube-ovn - where I would like to control the interconnect config (I want transit switches per VPC and I don't want the transit switches connected to the ovn-default VPC)

oilbeater:
@andrewlee1089 I think if it's for prototyping, we can manually verify this scenario through commands first. This change isn’t a complete solution yet—it still requires many other manual commands to achieve the goal, so it’s not suitable for merging now.

PR #5868, opened 2025-11-05
feat(netpols): add enforcement level none

SkalaNetworks:
# Pull Request

## What type of this PR

- Features

This PR adds a new enforcement level for network policies: "None"

When this mode is selected, no blocking is done by the network policy. The rules/ACLs to authorize the traffic are still added to the ports of each pod/VM, but it is effectively as if nothing happened.

If another network policy with a different enforcement targets the same pods as a netpol with enforcement "none", it will take precedence.

## What is it useful for

Using this enforcement, users can create network policies that allow traffic without blocking what is not authorized in the definition of the policy.

This is useful to pre-authorize traffic to some pods in anticipation of a policy with a normal enforcement being created later.

Example:
- A user wishes to allow TCP port 80 on a pod, but doesn't want to block any other port yet.
- This is effectively as if nothing had changed, as every port is still alowed
- Another user adds a network policy that targets the same pod, but that only allows port 22
- Because there's a network policy that allows 80 and another that allows port 22, both are allowed and the rest is now blocked.

In this example, we can preemptively allow traffic on pods without disrupting other ports. This is useful if we anticipate that network policies might be applied to a set of pods and we want to implicitly allow traffic beforehand.

My use case for this:
- I'm deploying Kubernetes as a Service through Kamaji in the namespace of customers
- These pods need certain ports open (for example 6443)
- Customers in the namespace can create network policies that will target those pods, accidentally blocking essential ports and killing the traffic to the controlplane
- Adding a network policy that allows those ports on the controlplane also means blocking every other port, which we might not want to do.
- Adding a rule that doesn't add the default block solves this problem.

### DIFF
--- a/charts/kube-ovn-v2/values.yaml
+++ b/charts/kube-ovn-v2/values.yaml
@@ -324,10 +324,12 @@ natGw:
 # @section -- Network Policies
 # @default -- "{}"
 networkPolicies:
-  # -- Enforcement level of network policies when they get applied (can be: standard, lax).
+  # -- Enforcement level of network policies when they get applied (can be: standard, lax, none).
   # Enforcement "standard" blocks everything except what is allowed by the network policies.
   # Enforcement "lax" is similar to "standard" with the exception that ARP/DHCPv4/DHCPv6/ICMPv4/ICMPv6
   # is allowed by default. This mode is useful when using Kubevirt and VMs with IPs configured via Kube-OVN's DHCP.
+  # Enforcement "none" adds rules to authorize the traffic allowed within network policies, but doesn't block
+  # unauthorized traffic unless another network policy with a different enforcement also applies to the same pods.
   # @section -- Network Policies
   enforcement: "standard"
 
--- a/pkg/controller/network_policy.go
+++ b/pkg/controller/network_policy.go
@@ -29,6 +29,7 @@ import (
 const (
 	NetworkPolicyEnforcementStandard = "standard"
 	NetworkPolicyEnforcementLax      = "lax"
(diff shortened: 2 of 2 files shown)


oilbeater:
You might want to explore `AdminNetworkPolicy` and `BaselineAdminNetworkPolicy` from the latest [NetworkPolicy API](https://network-policy-api.sigs.k8s.io/), which add two additional ACL tiers beyond standard NetworkPolicy. These cluster-level controls would handle your scenario more cleanly and manageably than adding enforcement levels, which can increase complexity.

SkalaNetworks:
@oilbeater Looking at the API, it seems like AdminNetworkPolicy and BaselineAdminNetworkPolicy have been deprecated and replaced by a ClusterNetworkPolicy. I get that setting the tier to "Baseline" could kinda do what the job of my annotation here, but it's a CRD with different levels of targetting (cluster-wide, and not namespaced, unlike netpols).

I would want to give users the option to create "allow only" network policies, and I'm unsure giving them the ability to create ClusterNetworkPolicy would be secure. 

The code for ANPs need to be overhauled to take into account the new API version.

What do you think?

oilbeater:
In Kube-OVN, we've already implemented the core logic for ANP and BANP, so migrating to the new ClusterNetworkPolicy should be relatively straightforward.

Regarding the "allow only" policies, I actually think maintaining the original default behavior would be cleaner than mixing two different network policy modes. Having both approaches could create maintenance complexity and confusion.

SkalaNetworks:
Ok sure, I'll close this and see if I can fit my usecase with ANPs

PR #5928, opened 2025-11-21
fix: use the same cni dir as the mount

6ixfalls:
# Pull Request

## What type of this PR

Bug fixes

Sets the cni-conf-dir to `/etc/cni/net.d`, because that's the path mounted to the installer pod. configDirectory is already used properly in the mount's creation.

### DIFF
--- a/charts/kube-ovn-v2/templates/agent/agent-daemonset.yaml
+++ b/charts/kube-ovn-v2/templates/agent/agent-daemonset.yaml
@@ -77,7 +77,7 @@ spec:
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         command:
           - /kube-ovn/install-cni.sh
-          - --cni-conf-dir=/etc/cni/net.d
+          - --cni-conf-dir={{ .Values.cni.mountConfigDirectory }}
           - --cni-conf-file={{ .Values.cni.localConfigFile }}
           - --cni-conf-name={{- .Values.cni.configPriority -}}-kube-ovn.conflist
         securityContext:
@@ -86,7 +86,7 @@ spec:
         volumeMounts:
           - mountPath: /opt/cni/bin
             name: cni-bin
-          - mountPath: /etc/cni/net.d
+          - mountPath: {{ .Values.cni.mountConfigDirectory }}
             name: cni-conf
           {{- if .Values.cni.mountToolingDirectory }}
           - mountPath: /usr/local/bin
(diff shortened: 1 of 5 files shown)


charts/kube-ovn-v2/templates/agent/agent-daemonset.yaml
```diff
@@ -77,7 +77,7 @@ spec:
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         command:
           - /kube-ovn/install-cni.sh
```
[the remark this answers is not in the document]
zbb88888: I think the mount dir should be defined in the helm charts values too.
/etc/cni/net.d should use the privious {{ .Values.cni.configDirectory }}

zbb88888 (changes requested):

charts/kube-ovn-v2/templates/agent/agent-daemonset.yaml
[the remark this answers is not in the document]
6ixfalls: Should another value be added for the mount path in the container? I know that's how Cilium does it.

zbb88888:
the topic: use the same cni dir as the mount(what target ???)

Why not make this target mount dir as a values in helm charts? 

I think specifying the static /etc/cni/net.d is not better than keeping it as a helm chat values.

charts/kube-ovn-v2/values.yaml
```diff
@@ -240,6 +240,9 @@ cni:
   # -- Location of the CNI configuration on the node.
   # @section -- CNI configuration
   configDirectory: "/etc/cni/net.d"
+  # -- Location of the CNI configuration to be mounted inside the pod.
+  # @section -- CNI configuration
+  mountConfigDirectory: "/etc/cni/net.d"
```
zbb88888, line 245: mountConfigDirectory configDirectory is just the same, can we use just one?

6ixfalls: No, the existing value is so users can set what path on the host is mounted in the pod, and the new value is the mount location inside the pod. While they are typically the same, one may change the first without changing the second. The second could technically be any valid path, but it's a value so there's no hidden dependency.

zbb88888: OK

zbb88888:
@6ixfalls Thanks!

oilbeater (approved):

PR #5929, opened 2025-11-22
fix: handle missing NAD during pod deletion

DiMalovanyy:
Allow pod cleanup when NetworkAttachmentDefinition is deleted before pod.
  Extract subnet info from pod annotations to release IPs and prevent subnet
  finalizer deadlock.

# Pull Request

## What type of this PR

- [ ] Features
- [X] Bug fixes
- [ ] Docs
- [ ] Tests

  ## Description

  ### Problem
  When a NetworkAttachmentDefinition (NAD) is deleted before pods that use it, pod deletion fails with the following issues:
  - IP custom resources remain orphaned and cannot be cleaned up
  - Subnets get stuck in `Terminating` state with finalizers that can never be removed
  - This is a common scenario with Multus multi-NIC setups and KubeVirt VMs where NADs have lifecycles tied to VPCs

  ### Root Cause
  The `getPodAttachmentNet()` function in `pkg/controller/pod.go` returns an error immediately when NAD is not found, even during pod deletion. This differs from
  how subnet deletion is handled, which tolerates missing resources during cleanup.

  ### Solution
  This fix allows pod cleanup to proceed even when NAD is missing by:
  1. Detecting pod deletion state (`ignoreSubnetNotExist` flag)
  2. When NAD is not found during deletion, constructing the provider name: `{nad-name}.{nad-namespace}.ovn`
  3. Finding the subnet using either:
     - Pod annotation: `{provider}.kubernetes.io/logical_switch`
     - Matching `subnet.Spec.Provider` field
  4. Using the discovered subnet for normal IP CR deletion and IPAM release

  This approach follows existing patterns in the codebase for handling missing resources during deletion.

  ### Testing
  - Tested on production system with custom build (`ghcr.io/kube-nfv/kubeovn:v1.14.11`) - verified fix resolves the issue with KubeVirt VMs using Multus multi-NIC
  - E2E tests passed:
    ```bash
    make kind-init
    make kind-install
    make e2e
  - Code follows existing error handling patterns for deletion paths

No issue created.

### DIFF
--- a/pkg/controller/pod.go
+++ b/pkg/controller/pod.go
@@ -1784,6 +1784,43 @@ func (c *Controller) getPodAttachmentNet(pod *v1.Pod) ([]*kubeovnNet, error) {
 		network, err := c.netAttachLister.NetworkAttachmentDefinitions(attach.Namespace).Get(attach.Name)
 		if err != nil {
 			klog.Errorf("failed to get net-attach-def %s, %v", attach.Name, err)
+			if k8serrors.IsNotFound(err) && ignoreSubnetNotExist {
+				// NAD deleted before pod, find subnet for cleanup
+				providerName := fmt.Sprintf("%s.%s.%s", attach.Name, attach.Namespace, util.OvnProvider)
+				subnetName := pod.Annotations[fmt.Sprintf(util.LogicalSwitchAnnotationTemplate, providerName)]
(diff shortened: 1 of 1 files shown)


oilbeater:
Thanks! @DiMalovanyy

PR #5934, opened 2025-11-24
fix: correct error log message in getDBStatus function

hargrovee:
The error message in getDBStatus() was hardcoded to display 'ovn-northbound' regardless of which database actually failed. This could be misleading when OVN_Southbound database fails.


oilbeater:
Thanks! @hargrovee

PR #5990, opened 2025-12-04
deps: migrate github.com/osrg/gobgp/v3 to github.com/osrg/gobgp/v4

zhangzujian:
# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)

### DIFF
--- a/pkg/speaker/bgp.go
+++ b/pkg/speaker/bgp.go
@@ -20,14 +20,14 @@ import (
 // and to withdraw the ones that should not be announced anymore
 func (c *Controller) reconcileRoutes(expectedPrefixes prefixMap) error {
 	if c.config.ExtendedNexthop || len(c.config.NeighborAddresses) != 0 {
-		err := c.reconcileIPFamily(net.IPv4len*8, expectedPrefixes)
+		err := c.reconcileIPFamily(api.Family_AFI_IP, expectedPrefixes)
 		if err != nil {
 			return fmt.Errorf("failed to reconcile IPv4 routes: %w", err)
 		}
 	}
 
 	if c.config.ExtendedNexthop || len(c.config.NeighborIPv6Addresses) != 0 {
-		err := c.reconcileIPFamily(net.IPv6len*8, expectedPrefixes)
+		err := c.reconcileIPFamily(api.Family_AFI_IP6, expectedPrefixes)
 		if err != nil {
 			return fmt.Errorf("failed to reconcile IPv6 routes: %w", err)
 		}
@@ -38,13 +38,7 @@ func (c *Controller) reconcileRoutes(expectedPrefixes prefixMap) error {
 
 // reconcileIPFamily announces prefixes we are not currently announcing and withdraws prefixes we should
 // not be announcing for a given IP family (IPv4/IPv6)
-func (c *Controller) reconcileIPFamily(bitLen int, expectedPrefixes prefixMap) error {
-	// Get the address family associated with the Kube-OVN family
-	afi, err := bitLenToAFI(bitLen)
-	if err != nil {
-		return fmt.Errorf("couldn't convert family to afi: %w", err)
-	}
-
+func (c *Controller) reconcileIPFamily(afi api.Family_Afi, expectedPrefixes prefixMap) error {
 	// Craft a BGP path listing request for this AFI
 	listPathRequest := apiutil.ListPathRequest{
 		TableType: api.TableType_TABLE_TYPE_GLOBAL,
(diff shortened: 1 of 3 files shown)


pkg/speaker/utils.go
```diff
@@ -1,112 +1,90 @@
...
-	for e := range existsMap {
-		if !expectedMap[e] {
-			toDel = append(toDel, e)
-		}
-	}
-
-	return toAdd, toDel
-}
-
 // parseRoute returns the prefix and length of the prefix (in bits) by parsing the received route
```
SkalaNetworks: Comment is not accurate anymore, we only return the prefix, not the length.

Maybe rename to convertRouteToPrefix?

Linked to the comment above, maybe keep returning the bitlen, as it proves useful. It allows for refactoring of addExpectedPrefix

pkg/speaker/utils.go
```diff
@@ -1,112 +1,90 @@
...
-	prefix := fmt.Sprintf("%s/%d", ip, maskMap[ipFamily])
-	expectedPrefixes[ipFamily] = append(expectedPrefixes[ipFamily], prefix)
+	addr, err := netip.ParseAddr(ip)
+	if err != nil {
+		klog.Errorf("failed to parse IP address %q: %v", ip, err)
+		return
+	}
+
+	bitLen := addr.BitLen()
+	prefix := netip.PrefixFrom(addr, bitLen).String()
```
SkalaNetworks: For factoring, use parseRoute? it does exactly what you want here, but we should keep returning the bitLen

pkg/speaker/config.go
```diff
@@ -163,30 +169,24 @@ func ParseFlags() (*Configuration, error) {
...
 		}
 	}
-	if *argNeighborIPv6Address != "" {
-		config.NeighborIPv6Addresses = strings.Split(*argNeighborIPv6Address, ",")
-		for _, addr := range config.NeighborIPv6Addresses {
-			if ip := net.ParseIP(addr); ip == nil || ip.To16() == nil {
-				return nil, fmt.Errorf("invalid neighbor-ipv6-address format: %s", *argNeighborIPv6Address)
-			}
+	for _, addr := range config.NeighborIPv6Addresses {
+		if addr.To4() != nil {
```
SkalaNetworks, line 178: nit but .To16 == nil seems clearer to me, especially if the address is somehow neither ipv4 or ipv6 (dunno if that could happen somewhere along the path, but better safe than sorry)

pkg/speaker/bgp.go
```diff
@@ -1,36 +1,33 @@
...
 )
 
-var maskMap = map[string]int{kubeovnv1.ProtocolIPv4: 32, kubeovnv1.ProtocolIPv6: 128}
-
 // reconcileRoutes configures the BGP speaker to announce only the routes we are expected to announce
 // and to withdraw the ones that should not be announced anymore
 func (c *Controller) reconcileRoutes(expectedPrefixes prefixMap) error {
 	if c.config.ExtendedNexthop || len(c.config.NeighborAddresses) != 0 {
-		err := c.reconcileIPFamily(kubeovnv1.ProtocolIPv4, expectedPrefixes)
+		err := c.reconcileIPFamily(net.IPv4len*8, expectedPrefixes)
```
SkalaNetworks: why are we switching to a bitlen everywhere instead of an enum? i feel like it's a bit less clear, and it introduces  a function to convert a bitlen to an afi

pkg/speaker/bgp.go
```diff
@@ -41,60 +38,61 @@ func (c *Controller) reconcileRoutes(expectedPrefixes prefixMap) error {
 
 // reconcileIPFamily announces prefixes we are not currently announcing and withdraws prefixes we should
 // not be announcing for a given IP family (IPv4/IPv6)
-func (c *Controller) reconcileIPFamily(ipFamily string, expectedPrefixes prefixMap) error {
+func (c *Controller) reconcileIPFamily(bitLen int, expectedPrefixes prefixMap) error {
 	// Get the address family associated with the Kube-OVN family
```
SkalaNetworks: Comment is not accurate anymore, we're not using family, but bitlen, but I'm unsure why

pkg/speaker/config.go
zhangzujian: Method `To16()` always returns a non-nil value. Here is an example: https://go.dev/play/p/1rwLaHRojSb

SkalaNetworks: Oh, it returns the IPv4 mapped into the IPv6 range in  0:0:0:0:0:ffff:X.X.X.X

That's too bad...

zhangzujian:
@SkalaNetworks Thanks for your review! I have made some adjustment according to your suggestions, please make a review again.

SkalaNetworks (approved):
Looks good to me :)

Issue #5995, opened 2025-12-05
[BUG] Openstack neutron security group (OVN ACLs) removed after kube-ovn-controller restart (v1.13.15+)


ycy1766:
### Kube-OVN Version

v1.14.10 (Regression starts from v1.13.15)

### Kubernetes Version

v.1.32

### Operation-system/Kernel Version

Ubuntu 24.04 

### Description

We are running [Genestack](https://github.com/rackerlabs/genestack) (OpenStack on Kubernetes), where OpenStack Neutron and Kube-OVN share the same OVN Northbound Database.

We have observed a critical regression starting from v1.13.15 (also present in v1.14.10). When the kube-ovn-controller restarts, it deletes critical Neutron Security Group rules (OVN ACLs) from the acls list managed by Neutron 

### Steps To Reproduce

1. Set up an environment where Neutron and Kube-OVN share the OVN NB DB (e.g., Genestack with Shared Underlay OVN).

2. Create a Security Group in OpenStack and verify ACLs managed by Neutron exist using ovn-nbctl.

Command: ovn-nbctl list acl | grep external

3.  Restart the controller:

Command: kubectl rollout restart deployment/kube-ovn-controller -n kube-system

4. Check the ACLs again to see if Neutron-managed rules are preserved.

### Current Behavior

The ACLs indicate they are Neutron-managed:
```
$ ovn-nbctl list acl | grep external | more
external_ids        : {"neutron:security_group_rule_id"="99e5c663-98c5-47ea-9368-e77c11eef66c"}
external_ids        : {"neutron:security_group_rule_id"="b9d11705-fbee-4ed6-bf66-3d0791b2ed23"}
external_ids        : {"neutron:security_group_rule_id"="edc79494-8551-46e5-bd75-da79b14989a7"}
external_ids        : {"neutron:security_group_rule_id"="cc0514d0-8d1c-46e3-ac5e-50a242efd6a9"}
external_ids        : {"neutron:security_group_rule_id"="b8b07ce3-82ab-4448-a0c4-a9ffe4cb6e23"}
external_ids        : {"neutron:security_group_rule_id"="5ab1853f-8eb8-4564-bc52-b2a1b85d7ec1"}
external_ids        : {"neutron:security_group_rule_id"="c4450b66-1fd6-44b0-b08c-3444fe9aaa7e"}
external_ids        : {"neutron:security_group_rule_id"="0ce808eb-b376-496a-8e00-927f04320b1f"}
[...]

PR #6001, opened 2025-12-05
feat: add vendor information to objects

cloudnull:
This change will implement a vendor name on ACL objects which will allow the garbage collection system to key off the vendor name before attempting to cleanup and remove database entries.

In the course of this investigation, I've also discovered other resources that are not using the kube-ovn vendor object, so I've taken steps to add it where appropriate.

Closes-Bug: https://github.com/kubeovn/kube-ovn/issues/5995
Related: https://github.com/rackerlabs/genestack/issues/1342

### DIFF
--- a/pkg/controller/init.go
+++ b/pkg/controller/init.go
@@ -30,8 +30,24 @@ import (
 func (c *Controller) InitOVN() error {
 	var err error
 
-	if err = c.migrateACLForVersionCompat(); err != nil {
-		klog.Errorf("failed to sync the older acl : %v", err)
+	// migrate vendor externalIDs to kube-ovn resources created in versions prior to v1.15.0
+	// this must run before ACL cleanup to ensure existing resources are properly tagged
+	if err = c.OVNNbClient.MigrateVendorExternalIDs(); err != nil {
+		klog.Errorf("failed to migrate vendor externalIDs: %v", err)
+		return err
+	}
+
+	// migrate tier field of ACL rules created in versions prior to v1.13.0
+	// after upgrading, the tier field has a default value of zero, which is not the value used in versions >= v1.13.0
+	// we need to migrate the tier field to the correct value
+	if err = c.OVNNbClient.MigrateACLTier(); err != nil {
+		klog.Errorf("failed to migrate ACL tier: %v", err)
+		return err
+	}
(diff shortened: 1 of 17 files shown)


pkg/controller/admin_network_policy.go
oilbeater, line 1: I think we can move the `vendor` part into `CreatePortGroup` and `CreateAddressSet` to avoid forgetting to use it later. We also need to find a way to migrate existing data, so when users upgrade to v1.15.0 with this patch, the legacy resources are automatically updated with the `vendor` label.

cloudnull: any recommendations on how you'd like to handle migrations, is there an pattern I should follow?

oilbeater: We can use the `pre-upgrade` hook in Helm charts to upgrade all ACLs and PortGroups with the vendor label. The challenge is figuring out how to identify or filter the relevant resources from OpenStack.

cloudnull:
@gemini-code-assist  review

cloudnull:
@gemini-code-assist review

cloudnull:
@gemini-code-assist review

cloudnull:
@oilbeater  I think this should now be good to go. I added tests and a migration function, which should detect the existing kube-ovn resources and add the vendor tags accordingly. I'd love your feedback on the patch as it stands.

pkg/controller/init.go
```diff
@@ -71,6 +71,13 @@ func (c *Controller) InitOVN() error {
 }
 
 func (c *Controller) migrateACLForVersionCompat() error {
+	// migrate vendor externalIDs to kube-ovn resources created in versions prior to v1.15.0
+	// this must run before ACL cleanup to ensure existing resources are properly tagged
+	if err := c.OVNNbClient.MigrateVendorExternalIDs(); err != nil {
```
oilbeater: It would be clearer to move this function into `initOVN`.

oilbeater (changes requested):

pkg/controller/init.go
cloudnull: Thank you for the suggestion @oilbeater, done.

PR #6025, opened 2025-12-09
feat(charts/kube-ovn-v2): enable custom node affinity on deployments

tallquist10:
# Pull Request

- [x] Make sure you have followed [Kube-OVN Code Style](https://github.com/kubeovn/kube-ovn/blob/master/CODE_STYLE.md). (Not applicable, as this is only Helm)

## What type of this PR

- Features ✅ 
- Bug fixes
- Docs ✅ 
- Tests

Add the ability to provide custom match expressions for preferred and required node affinity in the four deployments (central, monitor, controller, and ic).

This does not address any previously defined issues

### DIFF
--- a/charts/kube-ovn-v2/values.yaml
+++ b/charts/kube-ovn-v2/values.yaml
@@ -575,11 +575,26 @@ monitor:
       memory: "200Mi"
 
   # -- Node affinity configuration for kube-ovn-monitor.
+  # -- More information on formatting nodeAffinity can be found at https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity
   # @section -- OVN monitoring daemon configuration
   nodeAffinity:
     # -- User-preferred node affinity matchExpressions. Each entry should be a series of match expressions to be AND'd together, and all entries will be OR'd together
+    # -- Each entry should look like the following:
+    # -- matchExpressions:
+    # --   - key: topology.kubernetes.io/zone
+    # --     operator: In
+    # --     values:
+    # --     - antarctica-east1
+    # --     - antarctica-west1
     preferredDuringSchedulingIgnoredDuringExecution: []
     # -- User-required node affinity matchExpressions. Each entry should be a series of match expressions to be AND'd together, and all entries will be OR'd together
+    # -- Each entry should look like the following:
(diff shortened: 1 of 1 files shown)


charts/kube-ovn-v2/templates/_helpers.tpl
```diff
@@ -151,3 +151,36 @@ Get IPs of master nodes from values
...
+    {{- end }}
+  {{- end }}
+  {{- if $mergedRequired }}
+  requiredDuringSchedulingIgnoredDuringExecution:
+    nodeSelectorTerms:
+      {{- range $mergedRequired }}
+      - matchExpressions:
+          {{- toYaml .matchExpressions | nindent 8 }}
+      {{- end }}
+  {{- end }}
```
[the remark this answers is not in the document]
tallquist10: ~This seems reasonable, I will address this~

Edit: upon further investigation, since each entry has its own set of match expressions, the way it is written actually does allow for the desired effect. I will instead improve the documentation to be clear about this.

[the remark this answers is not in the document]
tallquist10: Documentation updated in the values.yaml file

charts/kube-ovn-v2/values.yaml
```diff
@@ -574,6 +574,14 @@ monitor:
       cpu: "200m"
       memory: "200Mi"
 
+  # -- Node affinity configuration for kube-ovn-monitor.
+  # @section -- OVN monitoring daemon configuration
+  nodeAffinity:
+    # -- User-preferred node affinity matchExpressions. Each entry should be a series of match expressions to be AND'd together, and all entries will be OR'd together
+    preferredDuringSchedulingIgnoredDuringExecution: []
```
oilbeater, line 589: Could you give an example about how to write the expression here.

tallquist10: Certainly. I will also link the K8s docs for affinity to provide added context.

tallquist10: Updated

oilbeater (approved):

oilbeater:
Thanks! @tallquist10

PR #6057, opened 2025-12-15
fix(helm): template and podaffinity

SkalaNetworks:
# Pull Request

## What type of this PR

- Bug fixes

PR https://github.com/kubeovn/kube-ovn/commit/5262026646c460436c76bfc2c17c22ad4e9d6c95 and https://github.com/kubeovn/kube-ovn/commit/a6fcfbcf56bc6b9f7b3fb1c0728b016065523b8b have introduced bugs that break templating.

PodAffinity has a faulty indentation and the address in the pinger is suffixed with a dash that breaks the resolution.

### DIFF
--- a/charts/kube-ovn-v2/templates/monitor/monitor-deployment.yaml
+++ b/charts/kube-ovn-v2/templates/monitor/monitor-deployment.yaml
@@ -45,7 +45,7 @@ spec:
         - key: CriticalAddonsOnly
           operator: Exists
       affinity:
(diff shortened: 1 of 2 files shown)


oilbeater:
I think we need an E2E for the v2 charts

SkalaNetworks:
Ideally I think we should ditch the script used to install Kube-OVN in E2E tests and use the chart instead. Then we can deprecate the v1 of the chart in favour of this one

PR #6059, opened 2025-12-15
fix(vpcnatgateway): bgp speaker would crash if no ipv4 is available

SkalaNetworks:
GoBGP v4 seems to crash when serializing with an invalid RouterID, whereas on older version, it defaulted to 0.0.0.0. There's functionally no problem with multiple peers having the same RouterID when doing eBGP. If using route reflectors, it might be more of a problem. In that case, users can use the --router-id parameter. Note that this case should not occur anyway, the NAT GW doesn't support IPv6 only networks right now as it cannot do NATv6. This is purely a failover to prevent crashes.

# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)

### DIFF
--- a/pkg/speaker/config.go
+++ b/pkg/speaker/config.go
@@ -183,11 +183,11 @@ func ParseFlags() (*Configuration, error) {
 	if config.RouterID == nil {
 		if podIPv4 != "" {
 			config.RouterID = net.ParseIP(podIPv4)
-		} else {
-			config.RouterID = net.ParseIP(podIPv6)
 		}
+
 		if config.RouterID == nil {
-			return nil, errors.New("no router id or POD_IPS")
+			// RouterID must be an IPv4. If no IPv4 exists on the speaker, fallback to 0.0.0.0 to avoid GoBGP crashing.
+			config.RouterID = net.ParseIP("0.0.0.0")
 		}
 	}
 


pkg/speaker/config.go
```diff
@@ -183,11 +183,11 @@ func ParseFlags() (*Configuration, error) {
 	if config.RouterID == nil {
 		if podIPv4 != "" {
 			config.RouterID = net.ParseIP(podIPv4)
-		} else {
```
zbb88888, line 186: Hi @SkalaNetworks ,  not support v6 for now？

SkalaNetworks: @zbb88888 Hi, it does support IPv6, but in IPv6-only setup, we can't use the IPv4 (int32) as the routerID (int32). And the IPv6 (int128) won't fit in the routerID. 

As we need a routerID (or GoBGP seemingly crashes), we default to 0.0.0.0 if we can't find anything suitable. In IPv6-only setup, the users can provide a custom routerID through --router-id if it is necessary for it to be unique.

zbb88888: @SkalaNetworks Hi, Thanks for your reply!

PR #6079, opened 2025-12-19
crd: support new tolerations operators Gt and Lt introduced in k8s v1.35

zhangzujian:
# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)


oilbeater:
What would happen if the Kubernetes version is below 1.35?

zhangzujian:
> What would happen if the Kubernetes version is below 1.35?

If the field `operator` is set to `Gt` or `Lt`, the generated  StatefulSet/Deployment will not start/run as expected. Users should take care not to set an incorrect value.

PR #6108, opened 2025-12-27
allow add annotations to vpc nat pod

zbb88888:
# Pull Request

## What type of this PR

- Features

Fixes #(issue-number)

### DIFF
--- a/pkg/controller/vpc_nat_gateway.go
+++ b/pkg/controller/vpc_nat_gateway.go
@@ -152,6 +152,9 @@ func isVpcNatGwChanged(gw *kubeovnv1.VpcNatGateway) bool {
 	if !slices.Equal(gw.Spec.Selector, gw.Status.Selector) {
 		return true
 	}
+	if !maps.Equal(gw.Spec.Annotations, gw.Status.Annotations) {
+		return true
+	}
 	if !reflect.DeepEqual(gw.Spec.Tolerations, gw.Status.Tolerations) {
 		return true
 	}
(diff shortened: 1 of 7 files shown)


pkg/controller/vpc_nat_gateway.go
```diff
@@ -843,21 +846,20 @@ func (c *Controller) GetSubnetProvider(subnetName string) (string, error) {
 }
 
 func (c *Controller) genNatGwStatefulSet(gw *kubeovnv1.VpcNatGateway, oldSts *v1.StatefulSet, natGwPodContainerRestartCount int32) (*v1.StatefulSet, error) {
-	annotations := make(map[string]string, 7)
-	if oldSts != nil && len(oldSts.Annotations) != 0 {
-		annotations = maps.Clone(oldSts.Annotations)
-	}
```
zbb88888, line 849: this oldSts.Annotations is not template.annotations，so removed

only gen annotations and use it as template.annotations, which is always clean and right.

oilbeater:
In standard Kubernetes workloads, a mechanism is used where labels such as `controller-revision-hash` and `pod-template-generation` are added to pods to detect whether the pods align with the controller's expectations. We can also adopt a similar approach to avoid having the same configuration appear in both the spec and the status.

zbb88888:
> In standard Kubernetes workloads, a mechanism is used where labels such as `controller-revision-hash` and `pod-template-generation` are added to pods to detect whether the pods align with the controller's expectations. We can also adopt a similar approach to avoid having the same configuration appear in both the spec and the status.

@Thanks, will do later

zbb88888:
冲突太多了，重新提了

PR #6118, opened 2025-12-30
fix: preserve user-specified network attachments in non-primary CNI mode

mii443:
When VpcNatGateway is deployed with custom k8s.v1.cni.cncf.io/networks annotations in non-primary CNI mode, the controller was overwriting the user's network configuration by appending the external network using simple string concatenation.

This caused issues when users specified network attachments in JSON array format (as supported by Multus CNI), resulting in either:
1. Malformed annotations that get rejected
2. Loss of internal subnet attachments, leaving only the external network

Changes:
- If user specifies custom network attachments in VpcNatGateway annotations, use them as-is (preserving both JSON array and comma-separated formats)
- Only apply default external network if no custom attachments are specified
- Maintains backward compatibility with existing deployments

Fixes issue where internal subnet (e.g., subnet-test1) was not attached when using non-primary CNI mode with custom network attachments.

# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)

### DIFF
--- a/pkg/controller/vpc_nat_gateway.go
+++ b/pkg/controller/vpc_nat_gateway.go
@@ -865,21 +865,24 @@ func (c *Controller) genNatGwStatefulSet(gw *kubeovnv1.VpcNatGateway, oldSts *v1
 	}
 	if c.config.EnableNonPrimaryCNI {
 		// We specify NAD using annotations when Kube-OVN is running as a secondary CNI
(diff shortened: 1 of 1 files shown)


mii443:
I apologize. I mistakenly submitted the PR.

PR #6119, opened 2025-12-31
set transform on informers to remove objects' managed fields

zhangzujian:
# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)


oilbeater:
We have extensively used Update(object) directly like [this](https://github.com/kubeovn/kube-ovn/blob/master/pkg/controller/subnet.go#L154), and I am concerned that trimming it might lead to changes in the fields of the original object in the apiserver.

zhangzujian:
> We have extensively used Update(object) directly like [this](https://github.com/kubeovn/kube-ovn/blob/master/pkg/controller/subnet.go#L154), and I am concerned that trimming it might lead to changes in the fields of the original object in the apiserver.

According to [k8s official document](https://kubernetes.io/docs/reference/using-api/server-side-apply/#clearing-managedfields), if the client-side operation is *update* or *patch*, managedFields will be stripped only when the passed in value is a list containing a single empty entry.

oilbeater (approved):

PR #6129, opened 2026-01-06
chore(charts): update readme on v2 chart and soft-deprecate v1

SkalaNetworks:
# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)

### DIFF
--- a/charts/kube-ovn/README.md
+++ b/charts/kube-ovn/README.md
@@ -3,6 +3,7 @@
 > [!CAUTION]
 > This chart is maintained as best effort, the v2 chart is the recommended way of deploying Kube-OVN.  
 > Please head [to this Helm Chart](../kube-ovn-v2) instead.
+> There is not yet a guide to upgrade from one chart to another, be careful if converting from v1 to v2.
 
 ## Installing the Chart
 


SkalaNetworks:
@oilbeater this updates the fields of the chart doc which haven't been updated in a while, also puts a warning on the old chart to ask users to head to the v2 version

oilbeater:
I think we need to find a way to seamlessly migrate from v1 to v2. Previously @changluyi found that when upgrade to v2, helm will delete and recreate all CRDs which will break the cluster network.

SkalaNetworks:
@changluyi The behaviour you encountered is strange considering the Helm chart for v2 uses exactly the same manifest for the CRDs as the v1 chart, do you have a reproducer?

oujonny:
I recently migrated one of our cluster from chart v1 to chart v2.

The main challenge I encountered was around the labels and matchLabels (selectors). But I managed to seamlessly migrated everything by deleting daemonsets and deployments with --cascade=orphan. I will write a proper guide for this and could share it here.

SkalaNetworks:
@oilbeater  is it ok if I merge this? I updated the disclaimer to notify this should be used for new installations and that there's no guide yet to migrate.

@oujonny if you've got a procedure that'd be lovely

oujonny:
https://github.com/kubeovn/kube-ovn/pull/6423

I created this PR with my notes for the v1 to v2 migration.

PR #6160, opened 2026-01-13
fix(helm): rolling upgrade strategy for ovs is wrong

SkalaNetworks:
We really need to introduce Helm tests at some point. There's no way the PR that introduced this bug was tested.

The syntax was wrong and any new deployment of the chart instantly lead to errors in the templating of ovs-ovn

### DIFF
--- a/charts/kube-ovn-v2/templates/_helpers.tpl
+++ b/charts/kube-ovn-v2/templates/_helpers.tpl
@@ -117,7 +117,7 @@ Get IPs of master nodes from values
       {{- end -}}
     {{- end -}}
   {{- else -}}
-    {{- $.Values.ovsOvn.updateStrategy -}}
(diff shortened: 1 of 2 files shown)


oujonny:
hey @SkalaNetworks 
I'm the one to blame... I just tested it locally, but changed something just before pushing the changes upstream. I'm sorry! I was in too much of a rush. But hey, mistakes can happen, right? As you already mentioned, adding a test for the Helm chart is the important thing/learning from it. 

Therefore I started with this PR: https://github.com/kubeovn/kube-ovn/pull/6184

SkalaNetworks:
@oujonny You're not to blame, no tests are in place, so no one can be sure the PR works.

Thanks for the PR, I'll take a look :)

PR #6184, opened 2026-01-20
feat: add GitHub Actions workflow for linting and testing Helm charts

oujonny:
# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)

### DIFF
--- a/charts/kube-ovn-v2/values.yaml
+++ b/charts/kube-ovn-v2/values.yaml
@@ -942,6 +942,3 @@ extraObjects: []
 #           }
 #         ]
 #       }
-
-# TODO: remove only temorary
-trigger: true


oujonny:
I managed to get the first successful run. This means a successful `lint` and a `helm install`
see: https://github.com/kubeovn/kube-ovn/actions/runs/21176042613/job/60905034334#step:10:33

@SkalaNetworks wdyt?

charts/kube-ovn-v2/Chart.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: v2
 name: kube-ovn-v2
 description: Helm chart for Kube-OVN
 type: application
-version: v1.16.0
+version: v1.16.1
 appVersion: "1.16.0"
 kubeVersion: ">= 1.29.0-0"
+maintainers:
```
SkalaNetworks: I'm not sure we should use this field, the maintainers of the chart are the maintainers and contributors of this project

.github/workflows/helm-testing.yaml
```diff
@@ -0,0 +1,61 @@
+name: Lint and Test Charts
+
+on: pull_request
```
SkalaNetworks: Shouldn't we only trigger when something in charts/* is modified?

My frame of reference is https://github.com/kubeovn/kube-ovn/blob/master/.github/workflows/build-x86-image.yaml

.github/workflows/helm-testing.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+      - uses: actions/setup-python@v6.0.0
+        with:
+          python-version: '3.x'
+          check-latest: true
+
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@v2.8.0
+
+      - name: Run chart-testing (list-changed)
```
SkalaNetworks, line 37: This checks if a change was introduced? If yes, it seems to target only "master", though this executes on every PR no matter the branch

SkalaNetworks (commented):
Thanks for the PR, have you tested this CI to see its results? I'm in interested in how it fails if linting or deploying doesn't work because of a broken commit.

SkalaNetworks:
image


The tests executed during this PR seem to fail but without triggering any error.

oujonny:
@SkalaNetworks, this is nothing new and the same for all other workflows with kind involved. For example, here: https://github.com/kubeovn/kube-ovn/actions/runs/21705765236/job/62597476753 (Post Install kind)

But I just realized that we have a basic installation (no linting) for the v1 chart in a workflow here: https://github.com/kubeovn/kube-ovn/blob/master/.github/workflows/build-x86-image.yaml#L1696

I think haveing it in a dedicated workflow is maybe better, or should we update the build-x86 to also run the tests for v2 chart?

.github/workflows/helm-testing.yaml
oujonny: The check get's executed on every PR, which is what I would expect right? We want this helm-testing before we end up on master. 
This check compares the source against the default branch; that's what the command is about.

.github/workflows/helm-testing.yaml
oujonny: absolutly!
Should we only trigger it for v2 chart?

SkalaNetworks:
I think a separate workflow makes everything cleaner

charts/kube-ovn-v2/Chart.yaml
oujonny: fair point, I found a way to disable this maintainer's validation and removed it from Chart.yaml

.github/workflows/helm-testing.yaml
oujonny: I would say let's only run it for v2 chart as I committed it now.

oujonny:
@SkalaNetworks, I resolved your comments and would say this PR is ready now for a merge, or did I miss something?

charts/kube-ovn-v2/values.yaml
```diff
@@ -942,3 +942,6 @@ extraObjects: []
 #           }
 #         ]
 #       }
+
+# TODO: remove only temorary
+trigger: true
```
SkalaNetworks: As says the comment, needs to be removed

.github/workflows/helm-testing.yaml
```diff
@@ -0,0 +1,68 @@
...
+        run: |
+          sudo pip3 install jinjanator j2cli
+          sudo PATH=~/.local/bin:$PATH make kind-init-${{ matrix.ip-family }}
+          sudo cp -r /root/.kube/ ~/.kube/
+          sudo chown -R $(id -un). ~/.kube/
+
+      - name: Run chart-testing (install)
+        if: steps.list-changed.outputs.changed == 'true'
+        run: |
+          kubectl label node --overwrite -l node-role.kubernetes.io/control-plane kube-ovn/role=master
```
SkalaNetworks, line 67: Will this ultimately verify if the deployment is working? If not, do we really need to go the full length of deploying a kind cluster and installing the chart inside (maybe we do for the few helpers in the chart that need a live running API?)

Can a simple dry-run do the job? I'm open to both honestly

oujonny: I think a Helm template could also be enough... that's true. But from the upstream GitHub action I'm using it forsees the install... and I think it does not hurt.

SkalaNetworks:
LGTM, thank you, this will get merged shortly

PR #6201, opened 2026-01-22
fix(vpcnatgw): cannot DNAT same EIP and same external port on two different protocols

SkalaNetworks:
You should be able to DNAT 1.1.1.1:22 TCP and 1.1.1.1:22 UDP on the same EIP, but the check doesn't verify if it is two different protocols. 

On Windows, RDP asks to open both the UDP and TCP protocols on the same port. It cannot be done behind an EIP attached to a VPC NAT gateway because of this bug.


SkalaNetworks:
@oilbeater do you think we can get this backported to 1.15?

PR #6204, opened 2026-01-23
fix(vpc-nat-gw): use consistent provider for all pod annotations

### LINKED ISSUE #6202: [BUG] Custom VPC NAT gateway routes do not work anymore
### Kube-OVN Version

v1.15.0

### Kubernetes Version

/

/

### Description

VPC NAT gateways have an option to set custom routes inside. This is useful, for example, to add a static route to redirect traffic from the Internet to some SwitchLBRules.

The routes are injected through the CRD, and propagated through the "ovn.kubernetes.io/routes: xxxx" annotation on the pod. 

This should work fine, as the "ovn.kubernetes.io" prefix refers to the primary interface of the Pod, which is the interface connected to the Subnet on which the VPC NAT gateway is located.

But this recent commit refactored the logic a bit:

https://github.com/kubeovn/kube-ovn/commit/4e35b4687816c3343a7989136304a1094d95ef9e

And now, instead, the routes are added to the NAD

Image

While this sounds like it should be doing the same exact thing, as both "ovn.kubernetes.io" and the NAD point to the same subnet, it seems like Kube-OVN's logic is not ready to handle that.

Because the routes are on a "bad annotation", the routes are never added to the pod and are completely ignored.

@zbb88888 I think this is a regression, is there a specific reason for which you wish to use the NAD provider instead of the default provider?

### Steps To Reproduce

1. Provision a VPC NAT gateway with custom routes
2. Head in the pod of the NAT gateway
3. "ip route"
4. Witness the route doesn't exist

### Current Behavior

Routes are ignored in the NAT GW

### Expected Behavior

Routes are applied

zbb88888:
The VPC NAT gateway custom routes were not being applied because of inconsistent annotation key prefixes. The `GenNatGwPodAnnotations` function was hardcoding the "ovn" prefix for annotations like `ovn.kubernetes.io/logical_switch`, while `setPodRoutesAnnotation` was using the actual subnet provider (e.g., "subnet.namespace.ovn").

This caused the CNI daemon to look for routes annotation with a different key than what was set by the controller, resulting in routes being ignored.

Changes:
- Add `provider` parameter to `GenNatGwPodAnnotations` function
- Use annotation templates instead of hardcoded constants to generate annotation keys dynamically based on the provider
- Remove redundant annotation settings in `EnableNonPrimaryCNI` branch since `GenNatGwPodAnnotations` now handles all cases uniformly
- Update unit tests to cover both "ovn" and NAD provider scenarios

Fixes #6202

# Pull Request

## What type of this PR

- Features- Bug fixes

Fixes #(issue-number)

### DIFF
--- a/pkg/util/vpc_nat_gateway.go
+++ b/pkg/util/vpc_nat_gateway.go
@@ -61,15 +61,19 @@ func GenNatGwSelectors(selectors []string) map[string]string {
 // GenNatGwPodAnnotations returns the Pod annotations for a NAT gateway
 // additionalNetworks is optional, used when user specifies extra NADs in gw.Annotations
 func GenNatGwPodAnnotations(gw *kubeovnv1.VpcNatGateway, externalNadNamespace, externalNadName, provider, additionalNetworks string) map[string]string {
+	p := provider
+	if p == "" {
+		p = OvnProvider
+	}
 	attachedNetworks := fmt.Sprintf("%s/%s", externalNadNamespace, externalNadName)
 	if additionalNetworks != "" {
 		attachedNetworks = additionalNetworks + ", " + attachedNetworks
 	}
 	return map[string]string{
-		fmt.Sprintf(VpcNatGatewayAnnotationTemplate, provider): gw.Name,
-		nadv1.NetworkAttachmentAnnot:                           attachedNetworks,
-		fmt.Sprintf(LogicalSwitchAnnotationTemplate, provider): gw.Spec.Subnet,
-		fmt.Sprintf(IPAddressAnnotationTemplate, provider):     gw.Spec.LanIP,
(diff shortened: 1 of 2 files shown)


SkalaNetworks:
Thanks a lot @zbb88888 !

This is a breaking problem in my environments so I would be really glad if we can get this backported in v1.15

@oilbeater Do we have an idea of when we can expect the first patch version of 1.15? There's several fixes that could benefit from a new release. Especially this PR, considering it fixes a regression introduced recently.

oilbeater:
@SkalaNetworks when this get merged I will bump a new release.

zbb88888:
@oilbeater @SkalaNetworks Hi，this PR is already in release-1.15

SkalaNetworks:
Thanks, I'll test by targeting the branch and wait for a v1.15.1 to release for production environments :)

PR #6211, opened 2026-01-26
charts: add external-gateway-config-ns option for controller

dnugmanov:
Add configurable namespace for ovn-external-gw-config ConfigMap.

By default, the controller's ConfigMap informer only watches ConfigMaps in the PodNamespace (where the controller runs). However, the default value for --external-gateway-config-ns is "kube-system", which causes the ConfigMap to never be found if the controller runs in a different namespace (e.g., "kubeovn").

This option allows users to explicitly set the namespace to match where their ovn-external-gw-config ConfigMap is located, or leave it empty to use the controller's namespace (PodNamespace).

Added to both kube-ovn and kube-ovn-v2 charts.


dnugmanov:
Hi, @oilbeater, could you help me with this PR? what suggestions from gemini do i need to fix to move PR on?

oilbeater (approved):

oilbeater:
Sorry for the late. Gemini's suggestion is not necessary and I will merge it.

PR #6212, opened 2026-01-26
fix(vpcnatgw): nat gateway uses faulty default network

SkalaNetworks:
Fixes https://github.com/kubeovn/kube-ovn/issues/6202

The 1.15.1 release has broken VPC NAT gateways. If you're not doing a NAT gateway in the default subnet/VPC (which is unlikely), you'll end up with NAT gateways that do not exist within your .spec.subnet at all. Instead, they'll be in the default VPC/subnet of the cluster.

In that state, 1.15.1 cannot be used to create NAT gateways.


SkalaNetworks:
@zbb88888 tell me what you think. I tested that on my cluster and it seems like the subnet injected is now the correct one.

Routes are also working just fine thanks to your fix.

SkalaNetworks:
image

Probably unrelated problem but NAT gateways have another issue when it comes to executing commands in the script that adds IPs/DNAT/SNATs, I'm investigating..

SkalaNetworks:
Ok the controller expects an annotation of type ovn.kubernetes.io/vpc_nat_gw to know the pod is a NAT gateway...
Except we use the provider now instead of ovn.***, so this broke. I don't think we need the provider here, so I'll remove that.

SkalaNetworks:
@zbb88888 The name of the annotation that denotes that a pod is a VPC NAT gateway was templated using the name of the provider. I don't think that's needed, a pod is either a NAT gateway or it isn't, we don't really care what provider it is attached to. Please tell me if you think this is ok.

This was introduced back when the "secondary CNI" mode was added to the code. But that seems unnecessary. I reverted this annotation to default to ovn.kubernetes.io/vpc_nat_gw

The code that checks if a pod is a NAT gateway to launch the init script within wasn't even programmed to check the templated annotation anyway, so it was bound to fail. It only worked because the default case was hardcoded to use the "ovn" provider. And your PR removed that provider from the annotation altogether, so that part of the code ceased to work.

I tested a NAT gateway (announced with BGP), custom routes are working, new IPs are added to the external interface and DNAT/SNAT/FIPs are now available :)

This PR should fix the 1.15.1 release, but I advise we rollout a 1.15.2 immediatly to prevent people from trying to create NAT gateways that will not work.

zbb88888:
> The 1.15.1 release has broken VPC NAT gateways. If you're not doing a NAT gateway in the default subnet/VPC (which is unlikely), you'll end up with NAT gateways that do not exist within your .spec.subnet at all. Instead, they'll be in the default VPC/subnet of the cluster.
> 
> In that state, 1.15.1 cannot be used to create NAT gateways.

@SkalaNetworks Let's make it more clear

1. If you're not doing a NAT gateway in the default subnet/VPC, you'll end up with NAT gateways that do not exist within your .spec.subnet at all.
2. 1.15.1 cannot be used to create NAT gateways.


about the `doing a NAT gateway in the default subnet/VPC`, Do you mean vpc-nat-gw spec subnet must use default vpc (ovn-cluster) subnet or the subnet in the default namspace?


please show the gw pod yamls

zbb88888:
In my env:  about the base case(not the bgp inside gw case ),the nat gw pod eth0 and net1 is ok

image

zbb88888:
About the `default` subnet, there are two ways:



1.   ovn.kubernetes.io/logical_switch

```bash

root@jfcs1001:~# k get pod -n kube-system                     kube-ovn-pinger-rd5b7 -o yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    k8s.v1.cni.cncf.io/network-status: |-
      [{
          "name": "generic-veth",
          "interface": "eth0",
          "ips": [
              "10.222.0.11"
          ],
          "mac": "66:46:47:a6:66:8b",
          "default": true,
          "dns": {},
          "gateway": [
              "10.222.0.1"
          ]
      }]
    ovn.kubernetes.io/allocated: "true"
    ovn.kubernetes.io/cidr: 10.222.0.0/16
    ovn.kubernetes.io/gateway: 10.222.0.1
    ovn.kubernetes.io/ip_address: 10.222.0.11
    ovn.kubernetes.io/logical_router: ovn-cluster
    ovn.kubernetes.io/logical_switch: ovn-default
    ovn.kubernetes.io/mac_address: 66:46:47:a6:66:8b
    ovn.kubernetes.io/pod_nic_type: veth-pair
    ovn.kubernetes.io/routed: "true"
    ovn.kubernetes.io/tunnel_key: "3"
  creationTimestamp: "2026-01-26T07:01:16Z"
  generateName: kube-ovn-pinger-
  labels:
    app: kube-ovn-pinger
    component: network
    controller-revision-hash: 88f5fb4bc
    pod-template-generation: "2"
    type: infra
  name: kube-ovn-pinger-rd5b7
  namespace: kube-system

``` 

2.  v1.multus-cni.io/default-network: default/ovn-vm

```bash

metadata:
  annotations:
    descheduler.alpha.kubernetes.io/request-evict-only: ""
    k8s.v1.cni.cncf.io/network-status: |-
      [{
          "name": "default/ovn-vm",
          "interface": "eth0",
          "ips": [
              "10.198.10.8"
          ],
          "mac": "7a:f4:db:78:d2:b8",
          "default": true,
          "dns": {}
      }]
    k8s.v1.cni.cncf.io/networks-status: |-
      [{
          "name": "default/ovn-vm",
          "interface": "eth0",
          "ips": [
              "10.198.10.8"
          ],
          "mac": "7a:f4:db:78:d2:b8",
          "default": true,
          "dns": {}
      }]
    kubectl.kubernetes.io/default-container: compute
    kubevirt.io/migrationTransportUnix: "true"
    kubevirt.io/vm-generation: "1"
    ovn-vm.default.ovn.kubernetes.io/allocated: "true"
    ovn-vm.default.ovn.kubernetes.io/cidr: 10.198.0.0/16
    ovn-vm.default.ovn.kubernetes.io/gateway: 10.198.0.1
    ovn-vm.default.ovn.kubernetes.io/ip_address: 10.198.10.8
    ovn-vm.default.ovn.kubernetes.io/logical_router: ovn-cluster
    ovn-vm.default.ovn.kubernetes.io/logical_switch: ovn-vm
    ovn-vm.default.ovn.kubernetes.io/mac_address: 7a:f4:db:78:d2:b8
    ovn-vm.default.ovn.kubernetes.io/pod_nic_type: veth-pair
    ovn-vm.default.ovn.kubernetes.io/routed: "true"
    post.hook.backup.velero.io/container: compute
    pre.hook.backup.velero.io/container: compute
    v1.multus-cni.io/default-network: default/ovn-vm
  creationTimestamp: "2026-01-12T11:16:18Z"
  labels:
    kubevirt.io: virt-launcher
    kubevirt.io/domain: ubuntu
    kubevirt.io/size: normal
  name: virt-launcher-i-xxx
  namespace: xxx

```


I think the multus NAD and the kube-ovn subnet spec provider should be accurate in all cases.



about this `this annotation to default to ovn.kubernetes.io/vpc_nat_gw`,  I think we should always use subnet spec provider. not just the ovn.kubernetes.io. just like the last PR, which I post to change the `hard code` way.

 Because there are two ways to provide the `default` subnet in kube-ovn


About remove `c.config.EnableNonPrimaryCNI`, I considered this too in the last PR. and the AI told me it is ok.
but I was not remove it, just to make that PR more clear.

zbb88888:
I think:

the `v1.multus-cni.io/default-network` exist is just equal to `c.config.EnableNonPrimaryCNI`, and `v1.multus-cni.io/default-network`  value is the NAD, which is more flexible

SkalaNetworks:
@zbb88888 What I see in my env on the a build from main:

- NAT gateways start using the custom provider to refer to the internal network
- They do not specify a default-network, so Kube-OVN defaults to creating a default interface on the Pod, connected to the default VPC/Subnet of the kube-system namespace (which is the default VPC/Subnet of the cluster)

My PR forces the use of a default-network annotation to prevent Kube-OVN CNI daemon from injecting the default network. If we don't do that, our custom internal network is entirely ignored (because it is nor the default network or in the Multus additional network annotation)

For the `ovn.kubernetes.io/vpc_nat_gw` annotation, I also agree that we should always use the templated annotation with the name of the provider as a prefix (e.g provider.ovn.kubernetes.io/ip_address: 1.1.1.1). But my understanding is that those annotations exist to provide contextual information to the CNI daemon. We basically tell Kube-OVN we want this particular provider to receive this particular configuration. This is not what `ovn.kubernetes.io/vpc_nat_gw` is about. This annotation only says the pod as a whole is a NAT gateway, it isn't information tied to a particular provider.

We could put this annotation on any of the providers of the NAT gateway, the internal, the external or in the case of a BGP VPC NAT gateway, to the API provider. A lot of annotations in Kube-OVN are non-templated with the provider when it doesn't make sense to do so, and I believe this is one of those cases.

Concerning your example https://github.com/kubeovn/kube-ovn/issues/6202#issuecomment-3803200341

I notice the annotations on the pod are of type ovn.kubernetes.io, which indicates the provider of your custom subnet is "ovn" I think?

Can you try with a provider that has a custom provider with a NAD (providername.providernamespace)? That should break it. The annotations wil be templated thanks to your PR and make the config cleaner, but a side effect will also try to check if the pod is a gateway based on the ovn.kubernetes.io annotation, eventhough you use a templated one.

That means the controller will not consider the pod to be a NAT gateway and will never run the init script on it. You'll probably also see that your internal network is completely ignored, as Kube-OVN will inject its own default one.

zbb88888:
> because it is nor the default network or in the Multus additional network annotation

I think if you should specify `default network`, and must use one annotation of `ovn-default annotation or the Multus defualt network annotation`


About the `For the ovn.kubernetes.io/vpc_nat_gw annotation: it isn't information tied to a particular provider.`, I think you are right. 

> Can you try with a provider that has a custom provider with a NAD (providername.providernamespace)? That should break it. The annotations wil be templated thanks to your PR and make the config cleaner, but a side effect will also try to check if the pod is a gateway based on the ovn.kubernetes.io annotation, eventhough you use a templated one.

I think this part is right too!

zbb88888:
@SkalaNetworks This PR look good to me! except  just removing this part.

image


I really want to refactor this part too!


I think:

the v1.multus-cni.io/default-network exist is just equal to c.config.EnableNonPrimaryCNI, and v1.multus-cni.io/default-network value is the NAD, which is more flexible.


I think we can use `v1.multus-cni.io/default-network`  to replace the flag c.config.EnableNonPrimaryCNI?  

if user specify the v1.multus-cni.io/default-network on the pod, kube-ovn-controller should do nothing about default nic.

How do you think about this?

SkalaNetworks:
@zbb88888 if the user is running Kube-OVN as a secondary CNI, the v1.multus-cni.io/default-network annotation will probably not be present.

The primary CNI will handle its networking as it wishes, there's no expectation that it is specified through a "default-network" annotation. Considering the current implementation, it seems like the default network of a NAT gateway is expected to be a Kube-OVN Subnet, no matter if the CNI is running as primary or secondary.

The bit you highlighted is about templating the annotation that specifies a pod is a NAT GW, is that the code you wanted edited? If yes, can you make some edits  to the PR using edit suggestions as I'm not sure what you want modified. I don't understand the secondary CNI mode of Kube-OVN quite well so my PR might miss something.

pkg/controller/pod.go
```diff
@@ -2627,23 +2627,7 @@ func (c *Controller) checkIsPodVpcNatGw(pod *v1.Pod) (bool, string) {
 	if pod.Annotations == nil {
 		return false, ""
 	}
-	// default provider
-	providerName := util.OvnProvider
-	// In non-primary CNI mode, we get the providers from the pod annotations
-	// We get the vpc nat gw name using the provider name
-	if c.config.EnableNonPrimaryCNI {
```
zbb88888, line 2634: @SkalaNetworks Are you using EnableNonPrimaryCNI？ if not, should not remove this section

zbb88888 (approved):

pkg/controller/pod.go
SkalaNetworks: This particular function doesn't need to know if we're primary/secondary CNI to infer if  a pod is a NAT gateway, so no, I'm not using EnableNonPrimaryCNI anymore.

But why would we keep this?

SkalaNetworks:
@oilbeater if we're good on this, I advise a 1.15.2 to overwrite the 1.15.1 that has a breaking bug with NAT GWs

pkg/controller/pod.go
zbb88888: ok， I see, seems util.VpcNatGatewayAnnotation is enough to identify the nat gw pod

zbb88888 (approved):

pkg/controller/pod.go
zbb88888: @SkalaNetworks Thanks for the details reply! LGTM!

zbb88888: image
you are right, only one place to use this annotatition key

SkalaNetworks:
@oilbeater could you please take a look at this for a merge+patch release considering the bug introduced in the latest version?

oilbeater:
I have cherry-pick it to release-1.15.

SkalaNetworks:
@oilbeater thanks, should we do something about the 1.15.1 release? Deployments of NAT GWs in that versions are non-functional in most cases.

oilbeater:
> @oilbeater thanks, should we do something about the 1.15.1 release? Deployments of NAT GWs in that versions are non-functional in most cases.

@SkalaNetworks I will bump a new release today.

PR #6223, opened 2026-01-28
NetworkPolicy: Provider-scoped policies for multi-network pods

### LINKED ISSUE #6205: [Feature Request] NetworkPolicy: Select OVN Logical Switch Ports for Multi-NIC Pods
## Summary
Add a way to explicitly select which logical port(s) a NetworkPolicy should apply to, rather than always implicitly applying to all OVN ports for a selected pod. This is needed when a pod has multiple OVN interfaces (across default or custom VPCs) and policy should target only a specific interface/port.

## Motivation / Use Case
- Multi-interface pods need different policy scopes per interface.
- Today, NetworkPolicy effectively targets **all OVN ports** of selected pods, which makes least‑privilege segmentation difficult.
- We want explicit per‑port selection for policies across **all VPCs** (default and custom).

## Current Behavior (verified in v1.15.0 code)
- **All OVN ports of selected pods are included**: `fetchSelectedPorts()` enumerates all Kube‑OVN networks for a pod and appends a port per provider when allocated. (`pkg/controller/network_policy.go:534–569`) — this is the issue being addressed.

## Expected Behavior
- Provide a mechanism to select specific logical ports or interfaces for NetworkPolicy enforcement **across all VPCs** (default and custom).
- Backward compatibility: if no selection is provided, behavior remains unchanged (apply to all relevant ports in a pod).

## Proposed Solution (optional)
One possible approach is a new NetworkPolicy annotation (name `ovn.kubernetes.io/policy-for`) for port selection, for example:
- Only primary port:
  - `ovn.kubernetes.io/policy-for: primary`
- Primary and selected additional port (Multus + Kube‑OVN):
  - `ovn.kubernetes.io/policy-for: primary, <net-attach-def-ns1>/<net-attach-def-name1>,<net-attach-def-ns2>/<net-attach-def-name2>`
- Only additional port (Multus + Kube‑OVN):
  - `ovn.kubernetes.io/policy-for: <net-attach-def-ns1>/<net-attach-def-name1>,<net-attach-def-ns2>/<net-attach-def-name2>`
- For all ports: Don't use the annotation

## Real Use Case (KubeVirt VM)
A VM Pod has two Kube‑OVN NICs (via Multus):
- **NIC A**: `192.168.131.11`
- **NIC B**: `192.168.132.11`

Two external networks exist and can reach VM via any NIC:
- **Network X**: `192.168.129.0/24`
- **Network Y**: `192.168.130.0/24`

Organizational policy requires:
- Network X can reach the VM **only via NIC A**.
- Network Y can reach the VM **only via NIC B**.

Today, NetworkPolicy applies to **all** OVN ports on the Pod, so you cannot enforce the split without changing upstream network devices (which may be out of your control). A per‑port selector would allow applying policy to a specific NIC and enforce this rule at the VM boundary.

_No response_

### Anything else?

I'm inspired by the [MultiNetworkPolicy](https://github.com/k8snetworkplumbingwg/multi-networkpolicy) project, but for Kube-OVN an annotation-based extension to NetworkPolicy aligns better with the existing OVN/LSP-centric design and avoids introducing a parallel policy API.

akbarkn:
# Pull Request

## What type of this PR

- Features
- Tests

## Describe your changes here

- Add `ovn.kubernetes.io/policy-for` parsing and provider‑scoped NetworkPolicy application.
- Filter selected ports/peers by provider and gate Service ClusterIP inclusion to primary‑only policies.
- Add unit tests for `parsePolicyFor`.
- Add e2e tests for provider scoping and ClusterIP gating.

Fixes #6205

### DIFF
--- a/pkg/controller/network_policy.go
+++ b/pkg/controller/network_policy.go
@@ -121,6 +121,8 @@ func (c *Controller) handleUpdateNp(key string) error {
 	}
 	logRate := parseACLLogRate(np.Annotations)
 
+	providers := parsePolicyFor(np)
+
 	npName := np.Name
 	nameArray := []rune(np.Name)
 	if !unicode.IsLetter(nameArray[0]) {
@@ -142,7 +144,7 @@ func (c *Controller) handleUpdateNp(key string) error {
 	}
 
 	namedPortMap := c.namedPort.GetNamedPortByNs(np.Namespace)
-	ports, subnetNames, err := c.fetchSelectedPorts(np.Namespace, &np.Spec.PodSelector)
+	ports, subnetNames, err := c.fetchSelectedPorts(np.Namespace, &np.Spec.PodSelector, providers)
 	if err != nil {
 		klog.Errorf("fetch ports belongs to np %s: %v", key, err)
 		return err
@@ -212,7 +214,7 @@ func (c *Controller) handleUpdateNp(key string) error {
 				} else {
 					var allow, except []string
 					for _, npp := range npr.From {
-						if allow, except, err = c.fetchPolicySelectedAddresses(np.Namespace, protocol, npp); err != nil {
+						if allow, except, err = c.fetchPolicySelectedAddresses(np.Namespace, protocol, npp, providers); err != nil {
 							klog.Errorf("failed to fetch policy selected addresses, %v", err)
 							return err
 						}
@@ -359,7 +361,7 @@ func (c *Controller) handleUpdateNp(key string) error {
 				} else {
 					var allow, except []string
 					for _, npp := range npr.To {
-						if allow, except, err = c.fetchPolicySelectedAddresses(np.Namespace, protocol, npp); err != nil {
+						if allow, except, err = c.fetchPolicySelectedAddresses(np.Namespace, protocol, npp, providers); err != nil {
 							klog.Errorf("failed to fetch policy selected addresses, %v", err)
 							return err
 						}
@@ -531,7 +533,53 @@ func (c *Controller) handleDeleteNp(key string) error {
 	return nil
 }
 
-func (c *Controller) fetchSelectedPorts(namespace string, selector *metav1.LabelSelector) ([]string, []string, error) {
+func parsePolicyFor(np *netv1.NetworkPolicy) set.Set[string] {
+	raw := strings.TrimSpace(np.Annotations[util.NetworkPolicyForAnnotation])
+	if raw == "" {
+		return nil
+	}
+
+	providers := set.New[string]()
+	invalidMsg := `ignore invalid network_policy_for annotation %q for netpol %s/%s, expect "ovn" or "<namespace>/<net-attach-def>"`
+
+	for token := range strings.SplitSeq(raw, ",") {
+		t := strings.TrimSpace(token)
+		if t == "" {
+			continue
+		}
+
+		if strings.EqualFold(t, "ovn") {
+			providers.Insert(util.OvnProvider)
+			continue
+		}
+		if strings.Contains(t, "/") {
+			parts := strings.SplitN(t, "/", 2)
+			if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+				klog.Warningf(invalidMsg, t, np.Namespace, np.Name)
+				continue
+			}
+			provider := fmt.Sprintf("%s.%s.%s", parts[1], parts[0], util.OvnProvider)
+			providers.Insert(provider)
+			continue
+		}
+		klog.Warningf(invalidMsg, t, np.Namespace, np.Name)
+	}
+
+	if providers.Len() == 0 {
+		klog.Warningf("network_policy_for annotation has no valid entries; policy %s/%s selects no pods", np.Namespace, np.Name)
+		return providers
+	}
+	return providers
+}
+
+func netpolAppliesToProvider(provider string, providers set.Set[string]) bool {
+	if providers == nil {
+		return true
+	}
+	return providers.Has(provider)
+}
+
+func (c *Controller) fetchSelectedPorts(namespace string, selector *metav1.LabelSelector, providers set.Set[string]) ([]string, []string, error) {
 	var subnets []string
 	sel, err := metav1.LabelSelectorAsSelector(selector)
 	if err != nil {
@@ -553,16 +601,24 @@ func (c *Controller) fetchSelectedPorts(namespace string, selector *metav1.Label
 			return nil, nil, fmt.Errorf("failed to get pod networks, %w", err)
 		}
 
+		matchedProvider := false
 		for _, podNet := range podNets {
 			if !isOvnSubnet(podNet.Subnet) {
 				continue
 			}
+			if !netpolAppliesToProvider(podNet.ProviderName, providers) {
+				continue
+			}
+			matchedProvider = true
 
 			if pod.Annotations[fmt.Sprintf(util.AllocatedAnnotationTemplate, podNet.ProviderName)] == "true" {
 				ports = append(ports, ovs.PodNameToPortName(podName, pod.Namespace, podNet.ProviderName))
 				// Pod selected by networkpolicy has its own subnet which is not the default subnet
 				subnets = append(subnets, podNet.Subnet.Name)
 			}
+		}
+		if providers != nil && !matchedProvider {
+			klog.V(4).Infof("skip pod %s/%s: no network attachment matches network_policy_for", pod.Namespace, pod.Name)
 		}
 	}
 	subnets = slices.Compact(subnets)
@@ -587,7 +643,7 @@ func hasEgressRule(np *netv1.NetworkPolicy) bool {
 	return np.Spec.Egress != nil
 }
 
-func (c *Controller) fetchPolicySelectedAddresses(namespace, protocol string, npp netv1.NetworkPolicyPeer) ([]string, []string, error) {
+func (c *Controller) fetchPolicySelectedAddresses(namespace, protocol string, npp netv1.NetworkPolicyPeer, providers set.Set[string]) ([]string, []string, error) {
 	selectedAddresses := []string{}
 	exceptAddresses := []string{}
 
@@ -643,7 +699,14 @@ func (c *Controller) fetchPolicySelectedAddresses(namespace, protocol string, np
 				klog.Errorf("failed to get pod nets %v", err)
 				return nil, nil, err
 			}
+			matchedProvider := false
 			for _, podNet := range podNets {
+				provider := podNet.ProviderName
+				if !netpolAppliesToProvider(provider, providers) {
+					continue
+				}
+				matchedProvider = true
+
 				podIPAnnotation := pod.Annotations[fmt.Sprintf(util.IPAddressAnnotationTemplate, podNet.ProviderName)]
 				podIPs := strings.SplitSeq(podIPAnnotation, ",")
 				for podIP := range podIPs {
@@ -654,6 +717,9 @@ func (c *Controller) fetchPolicySelectedAddresses(namespace, protocol string, np
 				if len(svcs) == 0 {
 					continue
 				}
+				if !shouldIncludeServiceIPs(podNet) {
+					continue
+				}
 
 				svcIPs, err := svcMatchPods(svcs, pod, protocol)
 				if err != nil {
@@ -661,9 +727,16 @@ func (c *Controller) fetchPolicySelectedAddresses(namespace, protocol string, np
 				}
 				selectedAddresses = append(selectedAddresses, svcIPs...)
 			}
+			if providers != nil && !matchedProvider {
+				klog.V(4).Infof("skip pod %s/%s: no network attachment matches network_policy_for", pod.Namespace, pod.Name)
+			}
 		}
 	}
 	return selectedAddresses, exceptAddresses, nil
+}
+
+func shouldIncludeServiceIPs(podNet *kubeovnNet) bool {
+	return podNet != nil && podNet.Subnet != nil && podNet.Subnet.Spec.Vpc == util.DefaultVpc
 }
 
 func svcMatchPods(svcs []*corev1.Service, pod *corev1.Pod, protocol string) ([]string, error) {
--- a/pkg/controller/network_policy_test.go
+++ b/pkg/controller/network_policy_test.go
@@ -1,0 +1,122 @@
+package controller
+
+import (
+	"testing"
+
+	netv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/utils/set"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/kubeovn/kube-ovn/pkg/util"
+)
+
+func TestParsePolicyFor(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name          string
+		annotation    *string
+		wantProviders set.Set[string]
+	}{
+		{
+			name:          "annotation omitted",
+			annotation:    nil,
+			wantProviders: nil,
+		},
+		{
+			name:       "ovn only",
+			annotation: ptrString("ovn"),
+			wantProviders: set.New(
+				util.OvnProvider,
+			),
+		},
+		{
+			name:       "duplicate ovn",
+			annotation: ptrString("ovn, ovn"),
+			wantProviders: set.New(
+				util.OvnProvider,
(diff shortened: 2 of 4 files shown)


pkg/controller/network_policy.go
```diff
@@ -28,6 +29,7 @@ import (
 const (
 	NetworkPolicyEnforcementStandard = "standard"
 	NetworkPolicyEnforcementLax      = "lax"
+	policyForAnnotation              = "ovn.kubernetes.io/policy-for"
```
SkalaNetworks: This could be in here https://github.com/kubeovn/kube-ovn/blob/503af3dbecc941c5d433655d623e52916b1549b6/pkg/util/const.go#L21


also we seem to prefer _ to -

and this https://github.com/kubeovn/kube-ovn/blob/503af3dbecc941c5d433655d623e52916b1549b6/pkg/util/const.go#L128C2-L128C36 annotation is specific to network policies

it is prefixed with network_policy

For consistency, we should probably reuse that prefix

pkg/controller/network_policy.go
```diff
@@ -531,7 +538,50 @@ func (c *Controller) handleDeleteNp(key string) error {
 	return nil
 }
 
-func (c *Controller) fetchSelectedPorts(namespace string, selector *metav1.LabelSelector) ([]string, []string, error) {
+func parsePolicyFor(np *netv1.NetworkPolicy) (map[string]struct{}, bool, error) {
```
SkalaNetworks: Could we remove the map[string]struct{} in favor of something more fitting such as Sets? 

I think we use 	"k8s.io/utils/set"  in a few parts of the code to deduplicate stuff and use it to know if something is present within a set

pkg/controller/network_policy.go
```diff
@@ -531,7 +538,50 @@ func (c *Controller) handleDeleteNp(key string) error {
...
+	includeSvc := false
+
+	for _, token := range strings.Split(raw, ",") {
+		t := strings.TrimSpace(token)
+		if t == "" {
+			continue
+		}
+
+		switch strings.ToLower(t) {
+		case "primary":
```
SkalaNetworks: In some cases, the primary interface might actually be something else than the default provider.

There can be an annotation on the Pod: `v1.multus-cni.io/default-network: provider`

If it is present, I think we should interpret "primary" as this.

pkg/controller/network_policy.go
```diff
@@ -531,7 +538,50 @@ func (c *Controller) handleDeleteNp(key string) error {
...
+			providers[util.OvnProvider] = struct{}{}
+			includeSvc = true
```
SkalaNetworks: This is obviously only true if we're in the default VPC, so my previous comment adds a case. This can only be set to true if the VPC of the provider is also the default VPC of the cluster.

pkg/controller/network_policy.go
```diff
@@ -531,7 +538,50 @@ func (c *Controller) handleDeleteNp(key string) error {
...
+			continue
+		case "default":
```
SkalaNetworks: Do we need this?

pkg/controller/network_policy.go
```diff
@@ -531,7 +538,50 @@ func (c *Controller) handleDeleteNp(key string) error {
...
+			return nil, false, fmt.Errorf("invalid policy-for entry %q (use 'primary')", t)
+		case "all":
```
SkalaNetworks: I guess this could be omitted as well

pkg/controller/network_policy.go
```diff
@@ -531,7 +538,50 @@ func (c *Controller) handleDeleteNp(key string) error {
...
+			return nil, false, fmt.Errorf("invalid policy-for entry %q (omit annotation for all)", t)
+		}
+		if strings.Contains(t, "/") {
+			parts := strings.SplitN(t, "/", 2)
+			if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+				return nil, false, fmt.Errorf("invalid policy-for entry %q", t)
```
SkalaNetworks: maybe specify the expected syntax in the error message to help debug

pkg/controller/network_policy.go
```diff
@@ -531,7 +538,50 @@ func (c *Controller) handleDeleteNp(key string) error {
...
+			}
+			provider := fmt.Sprintf("%s.%s.%s", parts[1], parts[0], util.OvnProvider)
+			providers[provider] = struct{}{}
```
SkalaNetworks: usage of a set might help a bit here to make the code easier to understand

pkg/controller/network_policy.go
```diff
@@ -557,6 +607,12 @@ func (c *Controller) fetchSelectedPorts(namespace string, selector *metav1.Label
 			if !isOvnSubnet(podNet.Subnet) {
 				continue
 			}
+			provider := podNet.ProviderName
+			if providers != nil {
```
SkalaNetworks: Can we make this a little function like "netpolAppliesToProvider(provider, providers) bool) just to make it easier to understand we're filtering the providers

pkg/controller/network_policy.go
```diff
@@ -644,14 +700,21 @@ func (c *Controller) fetchPolicySelectedAddresses(namespace, protocol string, np
 				return nil, nil, err
 			}
 			for _, podNet := range podNets {
+				provider := podNet.ProviderName
+				if providers != nil {
```
SkalaNetworks: reuse function from comment above? (netpolAppliesToProvider or something)

SkalaNetworks:
@akbarkn Thanks for the great feature

I made a few comments. Just out of curiosity, if I create a netpol that forbids all traffic for a specific provider, will this also work?

pkg/controller/network_policy.go
akbarkn: I’ll move the annotation constant to `pkg/util/const.go` next to other NetworkPolicy annotations and rename the annotation key to `ovn.kubernetes.io/network_policy_for` for consistency.

pkg/controller/network_policy.go
akbarkn: I’ll switch `providers` to `set.Set[string]` from `k8s.io/utils/set` and update signatures/usages so membership checks are clearer.

pkg/controller/network_policy.go
akbarkn: I will drop the explicit `default` branch. Treat it as an invalid entry: log a warning and ignore it. If no valid entries remain, the policy selects no ports (annotation omitted is the way to target all).

pkg/controller/network_policy.go
akbarkn: Agree. We can omit explicit `default`/`all` errors. We’ll log and ignore invalid entries. `all` should be represented by omitting the annotation; if all entries are invalid, the policy selects no ports.

pkg/controller/network_policy.go
akbarkn: I’ll log and ignore invalid entries and include the expected forms in the message.

pkg/controller/network_policy.go
akbarkn: Agree. I will change the `providers` to use set.

pkg/controller/network_policy.go
akbarkn: I will add `netpolAppliesToProvider(provider, providers)` and use it where we filter providers.

pkg/controller/network_policy.go
akbarkn: I will reuse the function from comment above

akbarkn:
> @akbarkn Thanks for the great feature
> 
> I made a few comments. Just out of curiosity, if I create a netpol that forbids all traffic for a specific provider, will this also work?

Yes. A policy with `policyTypes: [Ingress,Egress]` and no rules + `ovn.kubernetes.io/network_policy_for=<ns>/<nad>` will default‑deny traffic for that provider only. Other providers remain unaffected. If you omit the annotation, it applies to all providers (current behavior).

pkg/controller/network_policy.go
akbarkn: I think `primary` is ambiguous with Multus default network. I’ll rename the token to `ovn` to mean the default OVN provider (no NAD).

`ovn` doesn't work if:
•  the pod has `v1.multus-cni.io/default-network` (Kube‑OVN default port is not added), or  
•  the pod only has NAD‑based attachments.
In those cases you must use:
```
ovn.kubernetes.io/network_policy_for: <nad-namespace>/<nad-name>

File: pkg/controller/network_policy.go (hunk 4 above)
Comment (akbarkn) [reply pkg/controller/network_policy.go]: I will only include `ClusterIP` for ports that belong to the cluster default VPC (i.e., `podNet.Subnet.Spec.Vpc == util.DefaultVpc`). This prevents `ClusterIP` leakage into custom VPCs and remains correct for the default OVN provider (`ovn`).

**Comment by akbarkn** [comment]
@SkalaNetworks Please review when you have time. Thank you.

File: pkg/controller/network_policy.go [hunk 11]
```diff
@@ -531,7 +536,53 @@ func (c *Controller) handleDeleteNp(key string) error {
...
+			}
+			provider := fmt.Sprintf("%s.%s.%s", parts[1], parts[0], util.OvnProvider)
+			providers.Insert(provider)
+			continue
+		}
+		klog.Warningf(invalidMsg, t)
+	}
+
+	if len(providers) == 0 {
+		klog.Warning("network_policy_for annotation has no valid entries; policy selects no pods")
```
SkalaNetworks:
```suggestion
		klog.Warningf("network_policy_for annotation has no valid entries; policy %s/%s selects no pods", np.Namespace, np.Name)
```

pkg/controller/network_policy.go
```diff
@@ -531,7 +536,53 @@ func (c *Controller) handleDeleteNp(key string) error {
...
 
-func (c *Controller) fetchSelectedPorts(namespace string, selector *metav1.LabelSelector) ([]string, []string, error) {
+func parsePolicyFor(np *netv1.NetworkPolicy) (set.Set[string], error) {
+	raw := strings.TrimSpace(np.Annotations[util.NetworkPolicyForAnnotation])
+	if raw == "" {
+		return nil, nil
+	}
+
+	providers := set.New[string]()
+	invalidMsg := `ignore invalid network_policy_for entry %q, expect "ovn" or "<namespace>/<net-attach-def>"`
```
SkalaNetworks:
```suggestion
	invalidMsg := `ignore invalid network_policy_for annotation %q for netpol %s/%s, expect "ovn" or "<namespace>/<net-attach-def>"`
```

pkg/controller/network_policy.go
```diff
@@ -531,7 +536,53 @@ func (c *Controller) handleDeleteNp(key string) error {
...
+		}
+
+		if strings.EqualFold(t, "ovn") {
+			providers.Insert(util.OvnProvider)
+			continue
+		}
+		if strings.Contains(t, "/") {
+			parts := strings.SplitN(t, "/", 2)
+			if len(parts) != 2 || parts[0] == "" || parts[1] == "" {
+				klog.Warningf(invalidMsg, t)
```
SkalaNetworks:
```suggestion
				klog.Warningf(invalidMsg, t, np.Namespace, np.Name)
```

pkg/controller/network_policy.go
```diff
@@ -531,7 +536,53 @@ func (c *Controller) handleDeleteNp(key string) error {
...
+				continue
+			}
+			provider := fmt.Sprintf("%s.%s.%s", parts[1], parts[0], util.OvnProvider)
+			providers.Insert(provider)
+			continue
+		}
+		klog.Warningf(invalidMsg, t)
```
SkalaNetworks:
```suggestion
		klog.Warningf(invalidMsg, t, np.Namespace, np.Name)
```

pkg/controller/network_policy.go
```diff
@@ -553,17 +604,26 @@ func (c *Controller) fetchSelectedPorts(namespace string, selector *metav1.Label
 			return nil, nil, fmt.Errorf("failed to get pod networks, %w", err)
 		}
 
+		matchedProvider := false
 		for _, podNet := range podNets {
 			if !isOvnSubnet(podNet.Subnet) {
 				continue
 			}
+			provider := podNet.ProviderName
+			if !netpolAppliesToProvider(provider, providers) {
```
SkalaNetworks:
```suggestion
			if !netpolAppliesToProvider(podNet.ProviderName, providers) {
```

nit to remove a line

pkg/controller/network_policy.go
```diff
@@ -553,17 +604,26 @@ func (c *Controller) fetchSelectedPorts(namespace string, selector *metav1.Label
 			return nil, nil, fmt.Errorf("failed to get pod networks, %w", err)
 		}
 
+		matchedProvider := false
 		for _, podNet := range podNets {
 			if !isOvnSubnet(podNet.Subnet) {
 				continue
 			}
+			provider := podNet.ProviderName
```
SkalaNetworks:
```suggestion

```

pkg/controller/network_policy.go
```diff
@@ -531,7 +536,53 @@ func (c *Controller) handleDeleteNp(key string) error {
 	return nil
 }
 
-func (c *Controller) fetchSelectedPorts(namespace string, selector *metav1.LabelSelector) ([]string, []string, error) {
+func parsePolicyFor(np *netv1.NetworkPolicy) (set.Set[string], error) {
```
SkalaNetworks: does this ever return an error? your tests also check this function and is equipped to check for errors, but you never have one

pkg/controller/network_policy_test.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+func TestParsePolicyFor(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name          string
+		annotation    *string
+		wantProviders set.Set[string]
+		wantErr       bool
```
SkalaNetworks: this is never checked for "true", as your function doesn't error out

SkalaNetworks (changes requested):
apart from a few nits, this all seems good to me :)

pkg/controller/network_policy.go
akbarkn: will do

pkg/controller/network_policy.go
akbarkn: will do

pkg/controller/network_policy.go
akbarkn: will do

pkg/controller/network_policy.go
akbarkn: will do

pkg/controller/network_policy.go
akbarkn: will do

pkg/controller/network_policy.go
akbarkn: will do

pkg/controller/network_policy.go
akbarkn: I will change the function to not return error

pkg/controller/network_policy_test.go
akbarkn: I will remove it. the test will only check for `wantProviders`

SkalaNetworks:
@akbarkn Thanks, this looks good to me.

If you can get some docs published on Kube-OVN userguide, it would be great!

/lgtm

@oilbeater this looks fine considering the initial proposal in issue https://github.com/kubeovn/kube-ovn/issues/6205

akbarkn:
> @akbarkn Thanks, this looks good to me.
> 
> If you can get some docs published on Kube-OVN userguide, it would be great!
> 
> /lgtm
> 
> @oilbeater this looks fine considering the initial proposal in issue #6205

@SkalaNetworks Thanks for reviewing. I will update the docs once this PR merged.

oilbeater:
LGTM, only some lint issues here https://github.com/kubeovn/kube-ovn/actions/runs/21829170600/job/63069753460?pr=6223

SkalaNetworks:
@oilbeater am I authorized to merge PRs when they reach the dedicated state or do you want to add the extra check and handle that yourself?

oilbeater:
> @oilbeater am I authorized to merge PRs when they reach the dedicated state or do you want to add the extra check and handle that yourself?

Of course, go ahead!

SkalaNetworks:
@akbarkn thank you so much for this feature :)

If the tests are a go, I'll merge this!

akbarkn:
@SkalaNetworks @oilbeater Thanks to you too 🚀
I'll update the docs to add this feature

PR #6224, opened 2026-01-28
feat: support BGP and EVPN for vpc-egress-gateway

oilbeater:
This is an initial attempt to integrate BGP and EVPN functionality into vpc-egress-gateway.

The main architecture involves running an FRR container within the egress-gateway Pod to handle the BGP and EVPN control and data planes. The gateway init script will manage Linux network devices such as VRF, bridge, and VXLAN devices.

VXLAN encapsulation only takes place in the egress-gateway pod; the network between Kubernetes nodes still uses Geneve.

I'm not familiar with FRR or EVPN, so I need your advice on this architecture and implementation. @zbb88888 @SkalaNetworks @abasitt @donghee-park-rsk 

You can try a local demo with:

```bash
make kind-init-bgp kind-install
cd frr
bash start.sh
```

Not yet implemented:
1. L2VPN – this PR only implements L3VPN. I'm still learning how to configure FRR for L2VPN.
2. FRR hot reload.
3. BFD.

### DIFF
--- a/test/e2e/framework/bgpconf.go
+++ b/test/e2e/framework/bgpconf.go
@@ -1,0 +1,96 @@
+package framework
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kubernetes/test/e2e/framework"
+
+	"github.com/onsi/gomega"
+
+	apiv1 "github.com/kubeovn/kube-ovn/pkg/apis/kubeovn/v1"
+	clientset "github.com/kubeovn/kube-ovn/pkg/client/clientset/versioned"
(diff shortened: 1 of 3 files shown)


abasitt:
@oilbeater I have some thoughts and tradeoff about this and a different design approach. I will try to put that together today or tomorrow. Thank you

SkalaNetworks:
Will take a look tomorrow!

abasitt:
@oilbeater @SkalaNetworks @donghee-park-rsk what do you folks think of this? specially the alternative design. 

BTW i will be happy to validate the route type 5 integration but I am traveling next week, also like I mentioned in the doc, IMO direct BGP makes more sense than route type5 integration in the per gateway model. 
I don't think so we ever need to provide layer2 extension of EVPN, i can't imagines hundreds of pods/vms being advertised with it's mac addresses to the TOR.

Thank you again for the great work to keep bringing new features, we really love BGP.
https://docs.google.com/document/d/1UPAEX7Nq-A_XhiElXoBLsuilA4-aB1wG44TDRdGz45g/edit?usp=sharing

PR #6239, opened 2026-01-31
Ensure Kubevirt optimizations do not use old migration state

jskrill:
# Pull Request

- [x] Make sure you have followed [Kube-OVN Code Style](https://github.com/kubeovn/kube-ovn/blob/master/CODE_STYLE.md).

## What type of this PR

- [x] Bug fixes

This fixes a timing related bug where the Kubevirt live migration optimizations get the vmi before it has been updated with actual migration information. There are already some provisions in the code for this such as checking whether `vmi.Status.MigrationState != nil` but this only works on the first VM migration. On later VM migrations, `vmi.Status.MigrationState` may be set but with information for the last completed migration (not the current migration). The consequence of this is that the current code gets the wrong sourceNode and may skip (if returning to a previous node) or incorrectly set logical switch port options.

This is fixed by checking to make sure the migration id recorded on the vmi migration status is the same as the vmi migration object that we're handling at the moment.

In addition, the early return on `vmiMigration.Status.MigrationState.Completed` sometimes caused cleanup to be skipped (this one was inconsistent in my environment). I've removed the check to be safe, especially since there should be no harm in continuing the later logic unconditionally.

## Which issue(s) this PR fixes

Fixes #6220

PR #6261, opened 2026-02-04
Fix KubeVirt VM double migration

sircthulhu:
# Pull Request

## What type of this PR

Bug fixes

- After migration, traffic no longer sticks to the old Completed pod and correctly reaches the active Running pod.
- LSP migration cleanup is safer and avoids destructive updates when migration state or chassis data is incomplete.

We hit a repeatable outage after double consecutive KubeVirt VM migrations. The VM pod is migrated twice in a row. The old launcher pod goes Completed, the new pod is Running, but traffic is still routed to the old Completed pod IP. The VM stays unreachable until we manually delete the Completed pod or restart kube-ovn.

### Symptoms
- VM becomes unreachable after the second migration.
- virt-launcher old pod is `Completed`, new pod is `Running`.
- Both pods have the same IP (as expected in KubeVirt migration).
- Traffic still targets the old `Completed` pod; new pod gets no traffic.
- Deleting the Completed pod or restarting kube-ovn immediately fixes it.

### Reproduction steps
1. Create a KubeVirt VM (with live migration enabled).
2. Migrate VM from node A → node B.
3. Immediately migrate again from node B → node C.
4. Observe the old launcher pod in Completed and a new launcher pod in Running with the same IP.
5. Try to access the VM — traffic goes to the Completed pod, VM unreachable.

### DIFF
--- a/pkg/controller/kubevirt.go
+++ b/pkg/controller/kubevirt.go
@@ -133,10 +133,14 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
 		} else {
 			klog.Infof("current vmiMigration %s status %s, vmi MigrationState is nil", key, vmiMigration.Status.Phase)
 		}
-		// If we're at an end state and the vmi migration state is stale or nil, we're probably looking at an old migration
-		// either way, we can't proceed since we don't have the source and target nodes for resetting the migrate options
-		if vmiMigration.Status.Phase == kubevirtv1.MigrationSucceeded || vmiMigration.Status.Phase == kubevirtv1.MigrationFailed {
-			klog.V(3).Infof("VirtualMachineInstanceMigration %s migration state is Succeeded/Failed but VMI migration state is stale or nil, skipping", key)
+		if vmiMigration.Status.MigrationState != nil {
+			srcNodeName = vmiMigration.Status.MigrationState.SourceNode
+			targetNodeName = vmiMigration.Status.MigrationState.TargetNode
+		}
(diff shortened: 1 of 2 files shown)


sircthulhu:
Hi, @oilbeater 
Could you please review changes and give your opinion whether this will fix given issue?

oilbeater:
Have you tried https://github.com/kubeovn/kube-ovn/pull/6239? I think it resolves a similar issue by checking whether the vmiMigration ID matches the one in the vmi status. I also think we should return early if the IDs don't match.

jskrill:
As @oilbeater said, I think my fix #6239 should handle this already.

@sircthulhu Can you please test the existing logic?

@oilbeater As for why we only early return when VMIM state is terminal (Success/Failure) and migration state from VMI status is nil/stale is because the migration state from VMI status isn't needed to determine the srcNode/dstNode when the VMIM is in MigrationScheduling (the only non-terminal VMIM migration state). This let's us set the lsp options at the earliest possible point. Otherwise we might not set it at all

Ex (if we unconditionally return when VMIM != VMI migration state id):
1. VMIM is Scheduling/migration status of VMI is not yet set by the time we fetch it from the API
2. We early return since migration ID of VMI doesn't match VMIM
3. VMI migration status is finally set but we don't get the event since we're only watching VMIM changes
4. Pods try to start without lsp options set

We could also requeue the reconcile if VMI migration state doesn't match but but the existing logic is faster/simpler IMO.

PR #6268, opened 2026-02-05
perf: only enqueue subnet when necessary

### LINKED ISSUE #6235: [BUG] Unstoppable AddSubnet after upgrade to v1.15.2
### Kube-OVN Version

v1.15.2

### Kubernetes Version

-

-

### Description

Previous issue: https://github.com/kubeovn/kube-ovn/issues/5945
Same environment, upgrade from v1.13.13 to v1.15.1

Image

For unknown reason, my AddSubnet cannot being done as the screenshot. The last 2 times I both created a new subnet here. 

Image

And `SyncSecurityGroupPorts` seems never be done.
Image

### Steps To Reproduce

IDK...

### Current Behavior

-

### Expected Behavior

-

oilbeater:
Only enqueue subnet when necessary to improve performance.

Also fix vpc reconcile static routes and policy may incorrectly delete rules created by u2o and egress-gateway. This can be auto recover before because vpc will enqueue all subnets when updated.

Fix #6235 

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/controller/vpc.go
+++ b/pkg/controller/vpc.go
@@ -459,6 +459,23 @@ func (c *Controller) handleAddOrUpdateVpc(key string) error {
 				}
 			}
 		}
+	} else {
+		subnets, err := c.subnetsLister.List(labels.Everything())
+		if err != nil {
+			klog.Error(err)
+			return err
+		}
+		// Add static routes created by addCustomVPCStaticRouteForSubnet
+		for _, subnet := range subnets {
+			if subnet.Spec.Vpc == key {
+				staticTargetRoutes = append(staticTargetRoutes, &kubeovnv1.StaticRoute{
+					Policy:     kubeovnv1.PolicySrc,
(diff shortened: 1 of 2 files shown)


oujonny:
@oilbeater, we are hitting the mentioned bug from the issue that we have constant reconcilations. We run large clusters (ca. 70 nodes) with kubevirt. Therefore, the subnet controller can not keep up with the heartbeat annotation changes.

Any chance we get this PR backported to v1.14.x?

oilbeater:
@oujonny I recommend upgrading to v1.15 or later. When fixing the enqueue issues, we discovered many hidden bugs that did not appear previously because the constant reconciliation automatically resolved them. These bugs surfaced once we removed this loop. Therefore, backporting only this fix could lead to other issues, and it is not easy to backport all the related bug fixes because the logical and structure has changed since v1.14.

zsxsoft:
@oilbeater @oujonny I upgraded to v1.16.1, all works fine.

PR #6287, opened 2026-02-09
fix(iptables-eip): look up subnet directly by name instead of resolving through NAD

### LINKED ISSUE #6286: [BUG] IptablesEIP controller resolves ExternalSubnet through NAD lookup instead of direct Subnet lookup
### Kube-OVN Version

v1.15.0

### Kubernetes Version

v1.34.3+rke2r1

```shell
$ awk -F '=' '/PRETTY_NAME/ { print $2 }' /etc/os-release
"Harvester v1.7.1-rc4"

$ uname -r
6.4.0-36-default
```

### Description

The IptablesEIP API defines `spec.externalSubnet` as a Subnet resource name. However, the controller functions `handleAddIptablesEip` and `handleUpdateIptablesEip` in `pkg/controller/vpc_nat_gw_eip.go` incorrectly treat this field as a NetworkAttachmentDefinition (NAD) name by resolving it through `findSubnetByNetworkAttachmentDefinition`.

This causes a mismatch: users provide a Subnet name in the `externalSubnet` field (as the API intends), but the controller tries to look up a NAD with that name and then resolve it back to a Subnet, which fails when the NAD name differs from the Subnet name.

Other parts of the codebase (IPAM initialization in `pkg/controller/init.go` and webhook validation in `pkg/webhook/vpc_nat_gateway.go`) already handle this field correctly as a direct Subnet name.

### Steps To Reproduce

1. Create a Subnet whose name differs from the corresponding NetworkAttachmentDefinition name
2. Create an IptablesEIP resource with `externalSubnet` set to the Subnet name
3. The controller fails to resolve the subnet because it passes the Subnet name to `findSubnetByNetworkAttachmentDefinition`, which expects a NAD name

### Current Behavior

`handleAddIptablesEip` and `handleUpdateIptablesEip` call:
```go
subnets, err := c.subnetsLister.List(labels.Everything())
nadName := util.GetExternalNetwork(cachedEip.Spec.ExternalSubnet)
subnet, err := c.findSubnetByNetworkAttachmentDefinition(c.config.PodNamespace, nadName, subnets)
```
This treats `ExternalSubnet` as a NAD name and resolves it indirectly, which is inconsistent with the API definition and breaks when the Subnet name and NAD name differ.

There is also an inconsistency between the webhook and the controller: the webhook (`ValidateIptablesEIP`) correctly treats `ExternalSubnet` as a Subnet name by looking it up directly via `v.cache.Get(ctx, cli.ObjectKey{Name: externalNetwork}, subnet)`, while the controller resolves it through `findSubnetByNetworkAttachmentDefinition` as if it were a NAD name. In practice, the webhook acts as a gatekeeper that rejects any `ExternalSubnet` value that doesn't match an existing Subnet name, so the incorrect NAD-based resolution in the controller happens to work because the Subnet name is passed through the NAD lookup path, which coincidentally resolves to the same Subnet when names align.

### Expected Behavior

The controller should look up the Subnet directly by name, consistent with the API definition and other parts of the codebase.

starbops:
# Pull Request

## What type of this PR

- Bug fixes

Fixes #6286

### DIFF
--- a/pkg/controller/vpc_nat_gw_eip.go
+++ b/pkg/controller/vpc_nat_gw_eip.go
@@ -85,16 +85,10 @@ func (c *Controller) handleAddIptablesEip(key string) error {
 		return nil
 	}
 
-	subnets, err := c.subnetsLister.List(labels.Everything())
+	subnetName := util.GetExternalNetwork(cachedEip.Spec.ExternalSubnet)
+	subnet, err := c.subnetsLister.Get(subnetName)
 	if err != nil {
-		klog.Errorf("failed to list subnets: %v", err)
-		return err
-	}
-
-	nadName := util.GetExternalNetwork(cachedEip.Spec.ExternalSubnet)
-	subnet, err := c.findSubnetByNetworkAttachmentDefinition(c.config.PodNamespace, nadName, subnets)
-	if err != nil {
-		klog.Error(err)
+		klog.Errorf("failed to get subnet %s: %v", subnetName, err)
 		return err
 	}
 
@@ -194,16 +188,10 @@ func (c *Controller) handleUpdateIptablesEip(key string) error {
 	defer func() { _ = c.vpcNatGwKeyMutex.UnlockKey(key) }()
 	klog.Infof("handle update iptables eip %s", key)
 
-	subnets, err := c.subnetsLister.List(labels.Everything())
+	subnetName := util.GetExternalNetwork(cachedEip.Spec.ExternalSubnet)
+	subnet, err := c.subnetsLister.Get(subnetName)
 	if err != nil {
-		klog.Errorf("failed to list subnets: %v", err)
-		return err
-	}
-
-	nadName := util.GetExternalNetwork(cachedEip.Spec.ExternalSubnet)
-	subnet, err := c.findSubnetByNetworkAttachmentDefinition(c.config.PodNamespace, nadName, subnets)
-	if err != nil {
-		klog.Error(err)
+		klog.Errorf("failed to get subnet %s: %v", subnetName, err)
 		return err
 	}
 


starbops:
I didn't thoroughly check whether this has already been handled elsewhere. If you find that this PR is a duplicate or irrelevant, feel free to close it. Appreciate your time.

cc @oilbeater

oilbeater:
@starbops You're correct. I will backport it to release-1.15.

starbops:
@oilbeater Thank you so much.

rrajendran17:
@oilbeater I believe this commit is not backported to v1.15.0 yet. Can you please let us know once its backported ?
I was testing with kubeovn v1.15.0  version bumped on Harvester and was still seeing these errors.
```
I0219 00:02:32.512344       7 vpc_nat_gw_eip.go:81] handle add iptables eip my-eip
E0219 00:02:32.512396       7 subnet.go:2823] failed to get net-attach-def kube-system/subnetexternal: networkattachmentdefinition.k8s.cni.cncf.io "subnetexternal" not found

```
So kubeovn webhook expects a subnet and does look up in subnet cache and controller looks up for the nad.
https://github.com/kubeovn/kube-ovn/blob/dedad474819f37903cf9fcc085a5ea2f70ab754b/pkg/webhook/vpc_nat_gateway.go#L402
I had to bypass the webhook check temporarily to verify the vpc nat gw functionality in v1.15.0.

When I originally verified v1.15.0 with install script from kubeovn, it did not had the kubeovn webhook running, so I missed this error so went with the workaround of using a nad name in externalSubnet field of EIP.
Thanks @starbops for the fix.

oilbeater:
@rrajendran17 v1.15.4 has included this commit.

PR #6301, opened 2026-02-12
Fix IP Port Mapping overwritten by new LBs in the same VPC

SkalaNetworks:
# Pull Request

## What type of this PR
- Bug fixes
- Tests

When adding a SwitchLBRule in a VPC, the ip_port_mapping is populated to help the healthchecks run. If that field is not generated correctly, the healthchecks will not run and the LB will not work.

In the following condition, the ip_port_mapping is wrong:
- Add a first LB in the VPC (mapping is correct and holds the ports to the endpoints)
- Add a second LB (mapping is overwritten for this LB, first LB has its ip_port_mapping erased)

The logic to handle ip_port_mapping has been faulty for a while, I tried fixing it a couple month back but made the issue worse.

This handles the fix differently:
- I'm not trying to have an ip_port_mapping that contains only the backends that are currently active, if a LoadBalancer is deleted while the controller is down, the ip_port_mapping WILL contain stale data.
- Because the ip_port_mapping is not trying to be updated in a smart way, we can't accidentally erase an entry we need.
- The problem is, LBs will eventually clog the field with stale data.
- For that, I added a GC on the OVN loadbalancers. It checks if the ip_port_mapping contains backends that are not targetted by any VIP.

With this fix:
- We fix a bug that prevents creating more than one SLR per VPC (if healthchecks are activated, which they are by default)
- We ensure loadbalancers are clean in OVN even if the ovn-controller is down while the user deletes LBs/Services

I still lack 2 things that I will  add later in the GC:
- Cleaning up VIPs if no service/SLR points to it
- Cleaning up healthchecks if no service/SLR points to it

### DIFF
--- a/pkg/controller/gc.go
+++ b/pkg/controller/gc.go
@@ -51,7 +51,7 @@ func (c *Controller) gc() error {
 		c.gcVip,
 		c.gcLbSvcPods,
 		c.gcVPCDNS,
-		c.gcOvnLb,
+		c.gcIpPortMapping,
 	}
 	for _, gcFunc := range gcFunctions {
 		if err := gcFunc(); err != nil {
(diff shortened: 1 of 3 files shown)


SkalaNetworks:
I truly was under the impression that the GC ran periodically, but it actually only works for LSPs.

My GC function runs only once when the controller is restarted. I still wrote tests, but obviously they don't work.
I'm skipping the test right now and keeping the test in the code. 

I think we'll have to reasses that point

SkalaNetworks:
The current GC could do  what I need, I'll clos this and address it elsewhere

PR #6302, opened 2026-02-12
fix(loadbalancers): ip_port_mapping is overwritten by new vips

SkalaNetworks:
# Pull Request

## What type of this PR
- Bug fixes

There's a problem with SwitchLBRules right now:
- The ip_port_mapping is overwritten for any new SLR
- That means new SLRs will break previous ones

I seem to be the one that introduced this regression a few month back. This fixes it by adding ip_port_mappings one after another.

The GC will take care of cleaning up ip_port_mappings if some become stale.


SkalaNetworks:
@oilbeater this bug is breaking SLRs in custom VPCs, the fix seems to work on my clusters.

PR #6308, opened 2026-02-18
feat(helm):Add affinity and nodeSelector support for ovs-ovn and ovs-ovn-dpdk DaemonSets using v2 chart

morete:
# Pull Request

## What type of this PR

- Features

> - Add `affinity` and `nodeSelector` support for ovs-ovn and ovs-ovn-dpdk DaemonSets using v2 chart.
> - Default hardcoded values are moved over to the Helm default values

N/A

I need to get this ported back to helm chart v2 v1.14.x and v1.15.x

### DIFF
--- a/charts/kube-ovn-v2/templates/ovs-ovn/ovs-ovn-daemonset.yaml
+++ b/charts/kube-ovn-v2/templates/ovs-ovn/ovs-ovn-daemonset.yaml
@@ -47,8 +47,10 @@ spec:
           operator: Exists
         - key: CriticalAddonsOnly
           operator: Exists
+      {{- with .Values.ovsOvn.affinity }}
       affinity:
-        {{- toYaml .Values.ovsOvn.affinity | nindent 8 }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       priorityClassName: system-node-critical
       serviceAccountName: ovn-ovs
       automountServiceAccountToken: true
@@ -190,8 +192,10 @@ spec:
           resources:
             {{- toYaml . | trim | nindent 12 }}
           {{- end }}
+      {{- with .Values.ovsOvn.nodeSelector }}
       nodeSelector:
-        {{- toYaml .Values.ovsOvn.nodeSelector | nindent 8 }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       volumes:
         - name: usr-local-sbin
           emptyDir: {}
--- a/charts/kube-ovn-v2/templates/ovs-ovn/ovs-ovn-dpdk-daemonset.yaml
+++ b/charts/kube-ovn-v2/templates/ovs-ovn/ovs-ovn-dpdk-daemonset.yaml
@@ -43,8 +43,10 @@ spec:
     spec:
       tolerations:
       - operator: Exists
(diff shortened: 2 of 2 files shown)


morete:
I have verified the rendering of both ovs-ovn and ovs-ovn-dpdk daemonsets under various scenarios:

   - **With custom values:** Both affinity and nodeSelector render correctly. Note that Helm merges these with the default values in values.yaml unless explicitly overridden or set to null.
   - **Passed as `null`:** Both fields are correctly removed from the rendered YAML, as the {{- with ... }} block skips them.
   - **Passed as empty `{}`:** They merge with defaults. For example, nodeSelector still renders kubernetes.io/os: linux because it's defined in the base values.yaml.

  Verification Results:

 | Scenario | Affinity | NodeSelector | Result |
|--------|--------|--------|--------|
| Custom Values  | Rendered (Custom + Defaults) | Rendered (Custom + Defaults) | Success |
| Passed as `null` | Removed | Removed | Success |
| Passed as `{}` | Skipped (if default is {})| Rendered (Defaults) | Success |
| Defaults only | Skipped | Rendered (Defaults) | Success |

SkalaNetworks:
LGTM, thank you

SkalaNetworks:
@oujonny your recently added "Lint and Test Charts" test doesn't seem to be running?

I also notice we have those tests 
https://github.com/kubeovn/kube-ovn/actions/runs/22141372312/job/64125159405?pr=6308
https://github.com/kubeovn/kube-ovn/actions/runs/22141372312/job/64125159378?pr=6308

For the v1 Helm Chart, it has the benefits of testing if the Pods are ok, and passes a bunch of parameters to make Kube-OVN work. That might be a good idea.

oujonny:
I would say my new test is not running because the source branch has not yet been updated/rebased. @morete, can you rebase your commits to the latest upstream master branch? :)

I think for the v2 chart, we do not have to provide that much of paramets to make the kube-ovn installation run. The additional checks for the pods are nice. Let me add them with another PR.

morete:
@oujonny , rebased this branch now and test should then succeed.
@SkalaNetworks  thanks for picking this one up. Could workflows be approved again please?

SkalaNetworks:
@oujonny unsure what the pipeline is complaining about, but it is failing for some reason

oujonny:
The helm chart linting is failing because the chart version got not updated.

I would argue is a best practice to bump the chart version for each change. Similar as for docker images is not recommended to push multiple times to the same tag/version. But is seems like this practice is not applied to this chart. So not sure what to do know.

oilbeater:
@oujonny I think we can try to disable version check from option https://github.com/helm/chart-testing/blob/main/doc/ct_lint.md

SkalaNetworks:
I recommend we address that in this PR https://github.com/kubeovn/kube-ovn/pull/6311

We can let this test fail here, it doesn't really matter.

morete:
Hi @oilbeater , Would you please release this one? I would need to have it available on helm chart v2 1.14.x and 1.15.x would that be possible? Thanks a lot!

oilbeater:
@morete I have cherry-pick it and will release on next version.

morete:
Thanks a lot @oilbeater ! 🙏

PR #6311, opened 2026-02-19
feat: check for pod restarts after installation

oujonny:
# Pull Request

## What type of this PR

- Features

This PR improves the previously introduced chart testing for the kube-ovn-v2 helm chart. It features now a proper ct.yaml config file and performs additional checks after the installation to validate if pods are not crashing-looping.

Fixes #(issue-number)

### DIFF
--- a/.github/workflows/helm-testing.yaml
+++ b/.github/workflows/helm-testing.yaml
@@ -34,16 +34,7 @@ jobs:
       - name: Set up chart-testing
         uses: helm/chart-testing-action@v2.8.0
 
-      - name: Run chart-testing (list-changed)
-        id: list-changed
-        run: |
-          changed=$(ct list-changed --target-branch ${{ github.event.repository.default_branch }})
-          if [[ -n "$changed" ]]; then
-            echo "changed=true" >> "$GITHUB_OUTPUT"
-          fi
-
       - name: Run chart-testing (lint)
-        if: steps.list-changed.outputs.changed == 'true'
         run: |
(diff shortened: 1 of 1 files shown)


oujonny:
@SkalaNetworks, I improved the chart testing here and added the extra validation with the pod restarts check.
Here is an example of an successfull run: https://github.com/kubeovn/kube-ovn/actions/runs/22182595012/job/64147553679 

The current run is failing, as I added files the chart dir (ct.yaml and .helmignore), therefore it requires a chart bump, but I would say that's not needed here particularly.

SkalaNetworks:
https://github.com/kubeovn/kube-ovn/pull/6308

As highlited in this PR, can we disable the chart bump check?

```
 --check-version-increment              Activates a check for chart version increments (default true)
```

oujonny:
Hey
Yes aggree, for this particular chart make sense to disable this check. I implemented that via the ct.yaml config file.
Addtionalliy I revered the version bump introduced in https://github.com/kubeovn/kube-ovn/pull/6184 (back to v1.16.0) for now.

SkalaNetworks:
@oujonny Question: do we need to check the "ct list-changed" ? 

This CI/CD workflow should trigger when the Helm chart is changed, so why do we check again if it changes using "ct"?

Some steps run all the time (for example, the kind cluster). Only the lint runs if "ct list-changed" returned something, the rest runs all the time.

oujonny:
@SkalaNetworks, I just copied that from the example GitHub action. It has its roots in the fact that the ct is built to support multiple charts in the same repo, I would say.

As we run this only against the v2 chart, I removed it.

PR #6330, opened 2026-02-24
feat: Extend SG API to have tiers, larger priority range, localAddress and port matches

abhishek-pandey-1:
Enhances securityGroup API to allow users to fully utilize multiple tiers and do source IP and port filtering if needed.

# Pull Request

## What type of this PR

This PR contributes a enhancements to the SecurityGroup feature.

- Users can now use a larger priority range for the SecurityGroup: 1 – 16384. This expansion can be done safely without impacting any existing support for Kubernetes Network policies. 
- Users can now optionally define the tier to which a securityGroups will contribute ACL rules and set policy to Pass. This enables stacking of SecurityGroups to do multi-tier ACL processing. 
- Users can now also optionally specify localAddress and localPortRanges to do source IP/port based filtering.

### Review of how OVN ACL and kue-ovn 
OVN implements a multi-tier ACL: 

Total 4 tiers: 0,1,2,3 

Each tier can have rules with a priority in the range 0 – 32k. 
ACLs are processed starting at tier 0, with tier progressing until a decision is reached. 
Tier progression criteria: 
- If a match has action “pass”, the ACL processing moves to the next tier. 
- If no decision has been made after evaluating all rules in a tier, the tier progresses. 
If all tiers are exhausted, then ACL option set on port-group/switch is checked to see if the packet should be default deny or default allow. 

### current KubeOVN ACL Usage 

Kubeovn has the following interfaces to create/contribute ACL rules to a the ACL evaluated for a port. These are: 

- Kubernetes AdminNetworkPolicy 
- Kubernetes BaselineNetworkPolicy 
- Kubernetes NetworkPolicy 
- ACL field on a kubeovn/Subnet resource 
- Rules in a kubeovn/SecurityGroup resource 

#### Kubernetes policies 

1, 2, and 3 are applied on the default kubernetes network (eth0) of pods selected using namespace or pod selectors. 

The behaviour defined by Kubernetes is that: 

Admin denies traffic using AdminNetworkPolicies -  take precedence over all other network policies and cannot be overridden by network policy. 

Admin allows traffic using BaselineAdminNetworkPolicy. Developers can still further restrict this using NetworkPolicies. 

#### KubeOVN current tiered usage 
 
image

- Tier 1 used for admin policies which range from priority 30000 – 20000 
- Tier 2 used for: 
    - SecurityGroups 2300 – 2003 
    - K8s NetworkPolicy – 2001/2000 (can never override K8s AdminNetworkPolicy) 
    - Subnets 0-32000 

- Tier 3 used for BaselinNetworkPolicy which range from priority 1800-1700 

This means tier 2 and tier 3 have enough space to expand the priority range and tiers used by SecurityGroups. 

### Behaviour with the changes in this PR

KubeOVN Tiered ACLs Review

### DIFF
--- a/charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
+++ b/charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
@@ -2429,8 +2429,8 @@ spec:
                       priority:
                         type: integer
                         description: Priority of the policy route (0-32767)
-                        min: 0
-                        max: 32767
+                        minimum: 0
+                        maximum: 32767
                       action:
                         type: string
                       match:
@@ -3555,8 +3555,8 @@ spec:
                       priority:
                         type: integer
                         description: Rule priority (1-16384)
-                        min: 1
-                        max: 16384
+                        minimum: 1
+                        maximum: 16384
                       remoteType:
                         type: string
                         description: Type of remote (address, cidr, or securityGroup)
@@ -3569,13 +3569,13 @@ spec:
                       portRangeMin:
                         type: integer
                         description: Start of port range (1-65535)
-                        min: 1
-                        max: 65535
+                        minimum: 1
+                        maximum: 65535
                       portRangeMax:
                         type: integer
                         description: End of port range (1-65535)
-                        min: 1
-                        max: 65535
+                        minimum: 1
+                        maximum: 65535
                       policy:
                         type: string
                         description: Policy action (allow, pass or deny)
@@ -3585,13 +3585,13 @@ spec:
                       sourcePortRangeMin:
                         type: integer
                         description: Start of source port range (1-65535)
-                        min: 1
-                        max: 65535
+                        minimum: 1
+                        maximum: 65535
                       sourcePortRangeMax:
                         type: integer
                         description: End of source port range (1-65535)
-                        min: 1
-                        max: 65535
+                        minimum: 1
+                        maximum: 65535
                 egressRules:
                   type: array
                   description: Egress traffic rules for the security group
@@ -3607,8 +3607,8 @@ spec:
                       priority:
                         type: integer
                         description: Rule priority (1-16384)
-                        min: 1
-                        max: 16384
+                        minimum: 1
+                        maximum: 16384
                       remoteType:
                         type: string
                         description: Type of remote (address, cidr, or securityGroup)
@@ -3621,13 +3621,13 @@ spec:
                       portRangeMin:
                         type: integer
                         description: Start of port range (1-65535)
-                        min: 1
-                        max: 65535
+                        minimum: 1
+                        maximum: 65535
                       portRangeMax:
                         type: integer
                         description: End of port range (1-65535)
-                        min: 1
-                        max: 65535
+                        minimum: 1
+                        maximum: 65535
                       policy:
                         type: string
                         description: Policy action (allow, pass or deny)
@@ -3637,21 +3637,21 @@ spec:
                       sourcePortRangeMin:
                         type: integer
                         description: Start of source port range (1-65535)
-                        min: 1
-                        max: 65535
+                        minimum: 1
+                        maximum: 65535
                       sourcePortRangeMax:
                         type: integer
                         description: End of source port range (1-65535)
-                        min: 1
-                        max: 65535
+                        minimum: 1
+                        maximum: 65535
                 allowSameGroupTraffic:
                   type: boolean
                   description: Allow traffic between pods in the same security group
                 tier:
                   type: integer
                   description: ACL tier to which the rules are added
-                  min: 0
-                  max: 1
+                  minimum: 0
+                  maximum: 1
             status:
               type: object
               properties:
--- a/dist/images/install.sh
+++ b/dist/images/install.sh
@@ -2679,8 +2679,8 @@ spec:
                       priority:
                         type: integer
                         description: Priority of the policy route (0-32767)
-                        min: 0
-                        max: 32767
+                        minimum: 0
+                        maximum: 32767
                       action:
                         type: string
                       match:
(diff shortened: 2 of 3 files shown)


abhishek-pandey-1:
@oilbeater these are the tiered ACL changes we discussed a while back. These are fully back-compatible and only provide optional additive features to allow users to stack securityGroups to create a multi-tier ACL filtering and optionally do source ip/port based filtering as well.

I could not think of a sensible e2e test to add for this so I have skipped that and only added UTs. Let me know if there are any concerns.

(Apologies for the delay!)

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
```diff
@@ -3504,10 +3510,17 @@ spec:
...
+                        description: Policy action (allow, pass or deny)
+                      localAddress:
+                        type: string
+                      sourcePortRangeMin:
+                        type: integer
+                      sourcePortRangeMax:
+                        type: integer
                 allowSameGroupTraffic:
                   type: boolean
-                  description: Allow traffic between pods in the same security group
```
oilbeater, line 3623: The description for `allowSameGroupTraffic` is deleted by accident.

pkg/ovs/ovn-nb-acl.go
```diff
@@ -1415,7 +1452,7 @@ func (c *OVNNbClient) sgRuleNoACL(sgName, direction string, rule kubeovnv1.Secur
 
 	securityGroupHighestPriority, _ := strconv.Atoi(util.SecurityGroupHighestPriority)
 	priority := securityGroupHighestPriority - rule.Priority
-	exists, err := c.ACLExists(pgName, direction, strconv.Itoa(priority), match.String())
+	exists, err := c.ACLExists(pgName, direction, strconv.Itoa(priority), match.String(), tier)
```
oilbeater, line 1628: The `match` doesn't include `localAddress` and `sourcePortRange`

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
oilbeater, line 1: We can use crd openapi schema to limit the range of these fields.

pkg/controller/security_group.go
```diff
@@ -295,13 +303,33 @@ func (c *Controller) validateSgRule(sg *kubeovnv1.SecurityGroup) error {
 			return fmt.Errorf("not support sgRemoteType '%s'", rule.RemoteType)
 		}
 
+		if rule.LocalAddress != "" {
+			if strings.Contains(rule.LocalAddress, "/") {
+				if _, _, err := net.ParseCIDR(rule.LocalAddress); err != nil {
```
oilbeater: Can use `util.CheckCidr` and `util.IsValidIP` to simplify code.

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
abhishek-pandey-1: Good catch, will revert

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
abhishek-pandey-1: Thanks for calling this out. I will update the limits and also update the restriction put on priority

pkg/controller/security_group.go
abhishek-pandey-1: I have updated to use these functions - thanks!

pkg/ovs/ovn-nb-acl.go
abhishek-pandey-1: Ah great find - fixed now - let me know if any more concerns

dist/images/install.sh
oilbeater, line 1: The `install.sh` should also be updated

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
```diff
@@ -3504,10 +3517,25 @@ spec:
...
+                        max: 65535
+                      sourcePortRangeMax:
+                        type: integer
+                        description: End of source port range (1-65535)
+                        min: 1
+                        max: 65535
                 allowSameGroupTraffic:
                   type: boolean
                   description: Allow traffic between pods in the same security group
+                tier:
```
oilbeater, line 3650: Should also add limit for tier

abhishek-pandey-1: Thanks - added

dist/images/install.sh
abhishek-pandey-1: Thanks - I forgot about that one. Do you know if there is a plan to converge to a single CRD? Duplicate CRDs tend to go out of sync. Happy to contribute something here if you think that is a good change?

oilbeater: I usually just tell AI to sync them.

abhishek-pandey-1: Fair enough haha. Think I broke the tests, let me fix and ping you back for approvals. Feel free to suggest any more changes in the meanwhile.

pkg/controller/security_group.go
```diff
@@ -295,13 +308,33 @@ func (c *Controller) validateSgRule(sg *kubeovnv1.SecurityGroup) error {
...
 				return errors.New("portRange err, range Minimum value greater than maximum value")
 			}
+			if rule.LocalAddress != "" {
+				if rule.LocalPortRangeMin < 1 || rule.LocalPortRangeMin > 65535 || rule.LocalPortRangeMax < 1 || rule.LocalPortRangeMax > 65535 {
+					return errors.New("portRange is out of range")
+				}
+				if rule.LocalPortRangeMin > rule.LocalPortRangeMax {
+					return errors.New("portRange err, range Minimum value greater than maximum value")
+				}
+			}
```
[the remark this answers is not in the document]
abhishek-pandey-1: I think it is fine as it is

dist/images/install.sh
abhishek-pandey-1: Hi - I am a bit perplexed as to why min and max was being rejected for the tier but not for the the other fields like priority, portRange etc. given that min and max are not valid and the correct fields to use is minimum and maximum.

I wanted to look at this more but haven't had time to do so - for now fixed all instances of min and max and think everything is happy now.

oilbeater:
Thanks! @abhishek-pandey-1 can you also update the doc in https://github.com/kubeovn/docs/blob/master/docs/vpc/security-group.en.md

PR #6378, opened 2026-03-02
fix: skip redundant VPC annotation patch in non-primary CNI mode

### LINKED ISSUE #6377: Non-primary CNI: unnecessary VPC annotation patched on all services
In non-primary CNI mode, `handleUpdateEndpointSlice` patches `ovn.kubernetes.io/vpc: ovn-cluster` onto **every** service in the cluster — including services that have no kube-ovn endpoints (e.g. `kubernetes`, `kube-dns`).

This happens because VPC discovery falls back to the default VPC when no kube-ovn provider is found on endpoint pods, and the annotation is always written regardless.

These patches are redundant — all consumers of this annotation (`handleDeleteService`, `handleUpdateService`, `getDefaultVpcAndSubnet`) already fall back to `ClusterRouter` when it is absent. The unnecessary API writes cause extra reconcile cycles and annotation noise on unrelated services.

dnugmanov:
Fixes #6377

In non-primary CNI mode, `handleUpdateEndpointSlice` patches `ovn.kubernetes.io/vpc: ovn-cluster` onto every service — including those unrelated to kube-ovn. This is redundant because all consumers (`handleDeleteService`, `handleUpdateService`, `getDefaultVpcAndSubnet`) fall back to `ClusterRouter` when the annotation is absent.

**Change:** skip the annotation patch when `EnableNonPrimaryCNI` is true and the discovered VPC equals the default `ClusterRouter`. Custom VPC annotations are still written.

No behavior change for primary CNI mode.

### DIFF
--- a/pkg/controller/endpoint_slice.go
+++ b/pkg/controller/endpoint_slice.go
@@ -263,10 +263,14 @@ func (c *Controller) handleUpdateEndpointSlice(key string) error {
 	}
 
 	if svcVpc = svc.Annotations[util.VpcAnnotation]; svcVpc != vpcName {
-		patch := util.KVPatch{util.VpcAnnotation: vpcName}
-		if err = util.PatchAnnotations(c.config.KubeClient.CoreV1().Services(namespace), svc.Name, patch); err != nil {
-			klog.Errorf("failed to patch service %s: %v", key, err)
-			return err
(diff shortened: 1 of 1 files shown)


oilbeater:
@dnugmanov please fix the lint issue

Issue #5919, opened 2025-11-17
[feature] Do we need keep the subnet and ippool status type just the same?


zbb88888:
### Kube-OVN Version

master

### Kubernetes Version

no

### Operation-system/Kernel Version

no

### Description

```bash


ype SubnetStatus struct {
	// Conditions represents the latest state of the object
	// +optional
	// +patchMergeKey=type
	// +patchStrategy=merge
	Conditions []Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`

	V4AvailableIPs         float64                       `json:"v4availableIPs"`
	V4AvailableIPRange     string                        `json:"v4availableIPrange"`
	V4UsingIPs             float64                       `json:"v4usingIPs"`
	V4UsingIPRange         string                        `json:"v4usingIPrange"`
	V6AvailableIPs         float64                       `json:"v6availableIPs"`
	V6AvailableIPRange     string                        `json:"v6availableIPrange"`
	V6UsingIPs             float64                       `json:"v6usingIPs"`
	V6UsingIPRange         string                        `json:"v6usingIPrange"`
	ActivateGateway        string                        `json:"activateGateway"`
	DHCPv4OptionsUUID      string                        `json:"dhcpV4OptionsUUID"`
	DHCPv6OptionsUUID      string                        `json:"dhcpV6OptionsUUID"`
	U2OInterconnectionIP   string                        `json:"u2oInterconnectionIP"`
	U2OInterconnectionMAC  string                        `json:"u2oInterconnectionMAC"`
	U2OInterconnectionVPC  string                        `json:"u2oInterconnectionVPC"`
	NatOutgoingPolicyRules []NatOutgoingPolicyRuleStatus `json:"natOutgoingPolicyRules"`
	McastQuerierIP         string                        `json:"mcastQuerierIP"`
	McastQuerierMAC        string                        `json:"mcastQuerierMAC"`
}


type IPPoolStatus struct {
	V4AvailableIPs     internal.BigInt `json:"v4AvailableIPs"`
	V4AvailableIPRange string          `json:"v4AvailableIPRange"`
	V4UsingIPs         internal.BigInt `json:"v4UsingIPs"`
	V4UsingIPRange     string          `json:"v4UsingIPRange"`
	V6AvailableIPs     internal.BigInt `json:"v6AvailableIPs"`
	V6AvailableIPRange string          `json:"v6AvailableIPRange"`
[...]

PR #6403, opened 2026-03-07
subnet status use big int

zbb88888:
# Pull Request

## What type of this PR

- Features
https://github.com/kubeovn/kube-ovn/issues/5919

Fixes #(issue-number)


charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
```diff
@@ -2861,13 +2861,13 @@ spec:
               type: object
               properties:
                 v4availableIPs:
-                  type: number
```
oilbeater, line 2864: `number` should be fine. `IPPool` has already used number for bigInt.

oilbeater:
We also need to test if there are upgrade issues.

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
zbb88888: done

zbb88888:
> We also need to test if there are upgrade issues.

upgrade from 1.14 and 1.15 already covered in the e2e.

PR #6405, opened 2026-03-07
Split ECMP next-hops in custom VPC policy routes

### LINKED ISSUE #6401: [BUG] VPC spec.policyRoutes: ECMP next-hops silently dropped for custom VPCs
### Kube-OVN Version

v1.14.28 (also confirmed on v1.15.x)

### Kubernetes Version

v1.32.x

Ubuntu 24.04, Kernel 6.x

### Description

When using `spec.policyRoutes` on a **custom VPC** (non-default), comma-separated ECMP next-hops in `nextHopIP` are silently dropped — the policy route is created in OVN with **empty next-hops**.

**Root cause in source code:**

In `pkg/controller/vpc.go`, the VPC reconciliation for custom VPCs uses:

```go
// add new policies
for _, item := range policyRouteNeedAdd {
    if err = c.OVNNbClient.AddLogicalRouterPolicy(vpc.Name, item.Priority, item.Match,
        string(item.Action), []string{item.NextHopIP}, nil, externalIDs); err != nil {
```

This wraps the **raw comma-separated string** as a single-element slice `[]string{"172.26.31.51,172.26.31.52"}` instead of splitting it.

Compare with `addPolicyRouteToVpc()` in the same file, which correctly handles ECMP:

```go
func (c *Controller) addPolicyRouteToVpc(...) error {
    if policy.NextHopIP != "" {
        nextHops = strings.Split(policy.NextHopIP, ",")  // ← correct
    }
    c.OVNNbClient.AddLogicalRouterPolicy(vpcName, ..., nextHops, ...)
```

**Additionally**, duplicate `policyRoutes` entries with the same priority+match but different `nextHopIP` (an alternative approach to achieve ECMP) results in only the last entry being applied — the VPC controller overwrites the previous one during reconciliation via `diffPolicyRouteWithLogical`.

### Steps To Reproduce

1. Create a custom VPC with ECMP policyRoutes:

```yaml
apiVersion: kubeovn.io/v1
kind: Vpc
metadata:
  name: my-vpc
spec:
  policyRoutes:
    - action: reroute
      match: 'ip4.src == 172.26.31.0/24 && ip4.dst == 10.30.31.0/24'
      nextHopIP: '172.26.31.51,172.26.31.52'
      priority: 31100
```

2. Check OVN policy routes:

```
$ ovn-nbctl lr-policy-list my-vpc
     31100 ip4.src == 172.26.31.0/24 && ip4.dst == 10.30.31.0/24         reroute
```

The reroute has **no next-hop IPs** — they were silently dropped.

3. With a single next-hop it works correctly:

```yaml
  policyRoutes:
    - action: reroute
      match: 'ip4.src == 172.26.31.0/24 && ip4.dst == 10.30.31.0/24'
      nextHopIP: '172.26.31.51'
      priority: 31100
```

```
$ ovn-nbctl lr-policy-list my-vpc
     31100 ip4.src == 172.26.31.0/24 && ip4.dst == 10.30.31.0/24         reroute              172.26.31.51
```

### Current Behavior

- Comma-separated `nextHopIP` values are passed as a single string to OVN, resulting in empty/invalid next-hops
- No error or warning is logged — the failure is completely silent
- Attempting ECMP via two separate policyRoute entries with the same match results in only the last one being applied

### Expected Behavior

- `nextHopIP: '172.26.31.51,172.26.31.52'` should create an ECMP reroute policy with both next-hops (matching the behavior of `addPolicyRouteToVpc()`)
- Alternatively, multiple policyRoute entries with the same priority+match but different nextHopIP should create ECMP entries in OVN

### Suggested Fix

In `pkg/controller/vpc.go`, replace:

```go
[]string{item.NextHopIP}
```

with:

```go
strings.Split(item.NextHopIP, ",")
```

to match the existing behavior in `addPolicyRouteToVpc()`.

### Additional Context

**Related issue:** This bug is particularly impactful when combined with VpcEgressGateway. Using `spec.policyRoutes` on a custom VPC triggers full reconciliation via `diffPolicyRouteWithLogical`, which deletes all non-spec policy routes — including VpcEgressGateway-managed routes at priority 29100. After VPC reconciliation, the EG routes must be manually restored via annotation trigger. This is tracked separately in #6100.

utafrali:
Policy routes with multiple next-hop addresses (ECMP) were getting silently dropped for custom VPCs. The old code was passing the comma-separated NextHopIP as a single string in an array, when it should've been splitting them into individual addresses.

Fixed by using strings.Split to parse the addresses before passing to AddLogicalRouterPolicy. Added a test case to cover this scenario.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] I have read the **CODE_STYLE** document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- Fixes #6401

### DIFF
--- a/pkg/controller/vpc.go
+++ b/pkg/controller/vpc.go
@@ -563,7 +563,7 @@ func (c *Controller) handleAddOrUpdateVpc(key string) error {
 	// add new policies
 	for _, item := range policyRouteNeedAdd {
 		klog.Infof("add policy route for router: %s, match %s, action %s, nexthop %s, externalID %v", c.config.ClusterRouter, item.Match, string(item.Action), item.NextHopIP, externalIDs)
-		if err = c.OVNNbClient.AddLogicalRouterPolicy(vpc.Name, item.Priority, item.Match, string(item.Action), []string{item.NextHopIP}, nil, externalIDs); err != nil {
+		if err = c.OVNNbClient.AddLogicalRouterPolicy(vpc.Name, item.Priority, item.Match, string(item.Action), strings.Split(item.NextHopIP, ","), nil, externalIDs); err != nil {
 			klog.Errorf("add policy route to vpc %s failed, %v", vpc.Name, err)
 			return err
 		}
--- a/pkg/controller/vpc_test.go
+++ b/pkg/controller/vpc_test.go
@@ -294,3 +294,74 @@ func Test_handleAddOrUpdateVpc_staticRoutes(t *testing.T) {
 		require.NoError(t, err)
 	})
 }
+
+func Test_handleAddOrUpdateVpc_policyRoutes_ecmpNextHops(t *testing.T) {
+	t.Parallel()
+
+	vpcName := "test-vpc-policy"
+
+	t.Run("ECMP next-hops are split correctly for custom VPC policy routes", func(t *testing.T) {
+		fakeController := newFakeController(t)
+		ctrl := fakeController.fakeController
+		fakeinformers := fakeController.fakeInformers
+		mockOvnClient := fakeController.mockOvnClient
+
+		ctrl.vpcKeyMutex = keymutex.NewHashed(500)
+
+		vpc := &kubeovnv1.Vpc{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: vpcName,
+			},
+			Spec: kubeovnv1.VpcSpec{
+				StaticRoutes:   []*kubeovnv1.StaticRoute{},
+				EnableExternal: false,
+				PolicyRoutes: []*kubeovnv1.PolicyRoute{
+					{
+						Priority:  100,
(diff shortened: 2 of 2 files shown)


oilbeater:
@utafrali The DCO check failed, can you sign off the commit?

utafrali:
Done, pushed the fix.

oilbeater:
@utafrali I guess you forgot to push it 🤣

utafrali:
Sorry about that! Force-pushed with the DCO sign-off now.

oilbeater:
Thanks! @utafrali

utafrali:
Cheers

PR #6410, opened 2026-03-09
fix(helm): make v2 helm chart run upgrade logic within ovn

### LINKED ISSUE #6248: [BUG] OVS/OVN upgrade logic doesn't run on ArgoCD/v2 Helm Chart
### Kube-OVN Version

Any

### Kubernetes Version

Any

Any

### Description

The OVS/OVN upgrade hook is only rendered if this template function evaluates:

https://github.com/kubeovn/kube-ovn/blob/master/charts/kube-ovn-v2/templates/_helpers.tpl#L124

Two problems with this function:
- It uses the "lookup" Helm function. This is a problem as ArgoCD will not be able to use it. The repo-servers can't query the API of the target cluster (by design, as a security guarantee).
- It assumes the Daemonset for OVS/OVN has an annotation to track the version of the Helm Chart. This annotation was dropped on the v2 Helm Chart.

### Steps To Reproduce

The upgrade logic will not run in the following conditions:
- The chart (v1/v2) is installed using ArgoCD
- The v2 chart is upgraded (even manually)

### Current Behavior

OVS/OVN upgrades do not occur correctly

### Expected Behavior

OVS/OVN is upgraded when necessary

SkalaNetworks:
# Pull Request

## What type of this PR

- Bug fixes

The upgrade logic with the value injection in ovn-central didn't run correctly in ArgoCD deployed charts.

This PR ditches the model where we evaluate the value to inject in OVS using a templated switch. 

Considering the Helm Charts version are pinned to Kube-OVN (and OVN) versions, it makes no sense to use the Helm Chart in version v1.15 to deploy Kube-OVN in version v1.14.

That's why the PR introduces a value that needs to be updated each time the compatibility version in OVN is changed to something else. When a user upgrades their Kube-OVN version, the value will change and will run the upgrade logic.

The upgrade hook is also set to run on each sync now, avoiding complex logic to determine if it should run or not. If the script has nothing to do, will it auto-abort anyway.

Fixes #6248


SkalaNetworks:
@zhangzujian @zbb88888 @oilbeater 

What do you think of this? It follows my train of thought from the Slack.
There was a major flaw where ArgoCD deployments didn't run the upgrade logic, due to how the chart is created (it assumes the "lookup" function exists, but that's only for manual deployments of the Helm Chart).

This PR introduces a change, when a new OVN version is rolled out, instead of updating the template and setting a new value in the switch, a simple value has to be edited in the "values.yaml" to pin the Helm chart version with the new compatibility.

I need your opinion on this to ensure we don't break anything. 

But considering anyone who ran Kube-OVN's charts using ArgoCD never got the `version_compatibility` variable set in OVS NB, I think some users like me were probably missing on some specific patches that were added on OVS and that depend on this value...

PR #6425, opened 2026-03-11
Add option to flush stale conntrack flows after SG update

andriishestakov:
- Features

Add option to flush stale conntrack flows after SG update


pkg/daemon/sg_ct_reconcile_linux.go
```diff
@@ -0,0 +1,456 @@
...
+	}
+	for _, pod := range pods {
+		for _, ip := range getPodAllIPs(pod) {
+			if ip.Equal(podIP) {
+				return slices.Contains(getPodAllSGNames(pod), sgName)
+			}
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
andriishestakov: fixed

pkg/daemon/sg_ct_reconcile_linux.go
```diff
@@ -0,0 +1,467 @@
...
+			})
+		}
+	}
+	return podsByZone
+}
+
+// buildIPToSGsCache builds a map from pod IP to SG names for all pods in the
+// cluster. This is used to avoid listing all pods repeatedly for every CT entry.
+func (c *Controller) buildIPToSGsCache() (map[string][]string, error) {
+	allPods, err := c.podsLister.Pods(v1.NamespaceAll).List(labels.Everything())
```
oilbeater, line 152: The podInformer is built by 
```
  podInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(config.KubeClient, 0,
      kubeinformers.WithTweakListOptions(func(listOption *v1.ListOptions) {
          listOption.FieldSelector = "spec.nodeName=" + config.NodeName + ",spec.hostNetwork=false"
      }))
```

So it only has a node local pod cache not the cluster pod cache.

oilbeater (changes requested):

oilbeater:
I found ACL has an option `persist-established` in https://man7.org/linux/man-pages/man5/ovn-nb.5.html

If set to false, when the ACL’s action is changed to something other than allow-related, or if the ACL is destroyed it will automatically flush the ct. Can we use this to simplify the implementation?

andriishestakov:
> I found ACL has an option `persist-established` in https://man7.org/linux/man-pages/man5/ovn-nb.5.html
> 
> If set to false, when the ACL’s action is changed to something other than allow-related, or if the ACL is destroyed it will automatically flush the ct. Can we use this to simplify the implementation?

I guess that option if false by default, but i just checked with minimal code change to explicitly set to false in user defined sq + deny-all acl:
```
_uuid               : 068de9d7-cd25-4704-a89d-6f052424c50e
action              : drop
direction           : to-lport
external_ids        : {parent=ovn.sg.sg1, vendor=kube-ovn}
label               : 0
log                 : false
match               : "outport == @ovn.sg.sg1 && ip4 && ip4.src == 0.0.0.0/0 && icmp4"
meter               : []
name                : []
options             : {persist-established="false"}
priority            : 2210
sample_est          : []
sample_new          : []
severity            : []
tier                : 2
```
and it has no effect.

Initially there is strange behavior when change allow->deny. When run ping with allow rule, then change to deny - it drops half of packets. (applied after ~seq=12, so 13,15,17 are missing for some reason):
```
64 bytes from 10.0.1.9: seq=6 ttl=64 time=0.120 ms
64 bytes from 10.0.1.9: seq=7 ttl=64 time=0.123 ms
64 bytes from 10.0.1.9: seq=8 ttl=64 time=0.168 ms
64 bytes from 10.0.1.9: seq=9 ttl=64 time=0.130 ms
64 bytes from 10.0.1.9: seq=10 ttl=64 time=0.174 ms
64 bytes from 10.0.1.9: seq=11 ttl=64 time=0.119 ms
64 bytes from 10.0.1.9: seq=12 ttl=64 time=0.142 ms
64 bytes from 10.0.1.9: seq=14 ttl=64 time=0.169 ms
64 bytes from 10.0.1.9: seq=16 ttl=64 time=0.165 ms
64 bytes from 10.0.1.9: seq=18 ttl=64 time=0.156 ms
64 bytes from 10.0.1.9: seq=20 ttl=64 time=0.175 ms
64 bytes from 10.0.1.9: seq=22 ttl=64 time=0.167 ms
64 bytes from 10.0.1.9: seq=24 ttl=64 time=0.132 ms
```

pkg/daemon/sg_ct_reconcile_linux.go
andriishestakov: It is expected, each daemon should flush local CT flows which are related to local pods. Correct me if i wrong.

oilbeater: However the comment says that it's a map for all pods in the cluster.

oilbeater:
> minimal code change to explicitly set to false in user defined sq + deny-all acl

I guess this `persist-established` option only works for `allow-related` ACL.

PR #6445, opened 2026-03-13
Improve hairpin logic to support FIP, SNAT from any CIDR within the VPC and LoadBalancers

SkalaNetworks:
# Pull Request

## What type of this PR
- Features

This feature makes it so that hairpin works in any of the following conditions:
- Intra-Subnet traffic tries to reach a Pod/VM in the subnet through its EIP (whether FIP or DNAT)
- Intra-VPC traffic (if 2 subnets use the same NAT gateway) tries to reach a Pod/VM in the subnet through its EIP (whether FIP or DNAT)
- Traffic reaching a LoadBalancer within the subnet

This an improvement of the recent PR that handled hairpin for SNAT rules added to EIPs

### DIFF
--- a/test/e2e/iptables-vpc-nat-gw/e2e_test.go
+++ b/test/e2e/iptables-vpc-nat-gw/e2e_test.go
@@ -283,7 +283,7 @@ func hairpinSnatRuleExists(natGwPodName, eip string) bool {
 	// Use regex with \b word boundaries for precise matching, and allow
 	// optional trailing args like --random-fully after the EIP.
 	hairpinRulePattern := fmt.Sprintf(
-		`-A HAIRPIN_SNAT -o eth0 -m mark --mark 0x1/0x1 -m conntrack --ctstate DNAT --ctorigdst \b%s\b -j SNAT --to-source \b%s\b`,
+		`-A HAIRPIN_SNAT -o \b.*\b -m mark --mark 0x1/0x1 -m conntrack --ctstate DNAT --ctorigdst \b%s\b -j SNAT --to-source \b%s\b`,
 		regexp.QuoteMeta(eip), regexp.QuoteMeta(eip),
 	)
 	re := regexp.MustCompile(hairpinRulePattern)


SkalaNetworks:
@zbb88888 I'm gonna refine this PR, but testing it on my env, it addresses every hairpin case possible.

Basically, if the traffic enters the NAT gateway from its "VPC interface" (the interface facing the VPC, and not the external network), packets are firewall marked.

Packets are SNATed normally to go out of the NAT gateway through SNATs/FIPs without being affected.
If the packets are aimed at an EIP on the NAT gateway, they are DNATed to the internal destination IP (which would normally cause asymetric routing and problems).

But later, when the packets are processed to go out of the NAT gateway, we check if the packet was DNATed, with the initial target being an EIP. We also check if it has the firewall mark indicating it's traffic coming from the VPC (we don't want to SNAT traffic coming from the external subnet). If those conditions are met, we SNAT using the EIP to solve hairpin problems.

test/e2e/iptables-vpc-nat-gw/e2e_test.go
```diff
@@ -285,8 +283,8 @@ func hairpinSnatRuleExists(natGwPodName, cidr, eip string) bool {
 	// Use regex with \b word boundaries for precise matching, and allow
 	// optional trailing args like --random-fully after the EIP.
 	hairpinRulePattern := fmt.Sprintf(
-		`-A HAIRPIN_SNAT -s \b%s\b -d \b%s\b -m conntrack --ctstate DNAT -j SNAT --to-source \b%s\b`,
-		regexp.QuoteMeta(cidr), regexp.QuoteMeta(cidr), regexp.QuoteMeta(eip),
+		`-A HAIRPIN_SNAT -o eth0 -m mark --mark 0x1/0x1 -m conntrack --ctstate DNAT --ctorigdst \b%s\b -j SNAT --to-source \b%s\b`,
```
oilbeater: eth0 -> $VPC_INTERFACE

oilbeater (changes requested):

test/e2e/iptables-vpc-nat-gw/e2e_test.go
SkalaNetworks: Done, I used ".*" to match any possible interface, considering the name could be anything in scenarios where Kube-OVN is a secondary CNI

zbb88888 (approved):

zbb88888:
@SkalaNetworks Hi, Refining the harpin-snat to the dst eip is a better way, Thanks!

PR #6463, opened 2026-03-17
dist: add 25.03 ACL compatibility for OVN patch

changluyi:
## Summary
- add `compatible_25_03` handling to `dist/images/patches/e76880e792af56b2a3836098105079f5f8f1ff26.patch`
- skip generating ACL flows that match on `ct_mark.allow_established` in compatibility mode
- keep `build_stateful()` and `build_acls()` consistent during mixed-version rolling upgrades

## Why
OVN 25.03 introduces `ct_mark.allow_established` for ACL persistence. `ovn-northd` can generate flows that write and later match on this field so established connections can continue to pass without re-evaluating the full ACL path.

That field is not understood by older `ovn-controller` versions such as 24.03. During a rolling upgrade, if `northd` starts emitting flows with `ct_mark.allow_established` before all controllers are upgraded, old controllers can fail to parse those flows.

This change adds `compatible_25_03` so compatibility mode disables both sides of that feature:
- the `ct_commit` action that writes `ct_mark.allow_established`
- the ACL flow that matches `ct_mark.allow_established == 1`

## What `ct_mark.allow_established` does
`ct_mark.allow_established` is an OVN conntrack mark used by ACL persistence.

It is written when the first packet of a connection is committed, and then read by later ACL flows for established/reply traffic. The purpose is to let an already-established connection continue to pass without re-running the full ACL decision path.

In practice:
- `build_stateful()` writes it into conntrack state
- `build_acls()` matches it for established traffic
- older `ovn-controller` versions cannot parse this field

## Validation
- `make lint`

### DIFF
--- a/dist/images/patches/e76880e792af56b2a3836098105079f5f8f1ff26.patch
+++ b/dist/images/patches/e76880e792af56b2a3836098105079f5f8f1ff26.patch
@@ -1,19 +1,19 @@
 From e76880e792af56b2a3836098105079f5f8f1ff26 Mon Sep 17 00:00:00 2001
 From: clyi <clyi@alauda.io>
 Date: Fri, 8 Aug 2025 17:27:38 +0800
-Subject: [PATCH] [PATCH] northd: add nb option version_compatibility
+Subject: [PATCH] northd: add nb option version_compatibility
 
 Signed-off-by: clyi <clyi@alauda.io>
 ---
- northd/en-global-config.c |   5 ++
- northd/northd.c           | 137 +++++++++++++++++++++++++++-----------
- 2 files changed, 103 insertions(+), 39 deletions(-)
+ northd/en-global-config.c |   5 +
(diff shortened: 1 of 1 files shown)


dist/images/Dockerfile.base
```diff
@@ -125,7 +125,7 @@ RUN cd /usr/src/ && git clone -b branch-25.03 --depth=1 https://github.com/ovn-o
     git apply $SRC_DIR/e5916eb53abc3b7d28c407c3c47566c46116090a.patch && \
     # support dedicated BFD LRP
     git apply $SRC_DIR/e4e6ea9c5f4ba080b719924e470daa8094ff38a7.patch && \	
-    # northd: add nb option version_compatibility
+    # northd: add nb option version_compatibility
```
oilbeater:
```suggestion
    # northd: add nb option version_compatibility
```

PR #6529, opened 2026-03-26
Fix VpcNatGateway connection issues

andriishestakov:
- Bug fixes

### DIFF
--- a/dist/images/vpcnatgateway/nat-gateway.sh
+++ b/dist/images/vpcnatgateway/nat-gateway.sh
@@ -194,6 +194,15 @@ function init() {
         echo "IFB kernel module loaded successfully"
     fi
 
+    # MASQUERADE for DNAT loop-back: when a DNAT rule rewrites the destination to
+    # an internal VIP (e.g. EIP:port -> VIP:port), the kernel re-routes the packet
+    # back out $VPC_INTERFACE.  Without MASQUERADE the original source IP is
+    # preserved, so the backend responds directly to the original source, bypassing
+    # the NAT GW's conntrack and breaking reverse-NAT unwind.  Using ctstate NEW
+    # rewrites the source to the VPC interface IP only for new connections, so the
+    # backend always responds through the NAT GW for correct DNAT unwind.
+    $iptables_cmd -t nat -A POSTROUTING -o "$VPC_INTERFACE" -m conntrack --ctstate NEW -j MASQUERADE
+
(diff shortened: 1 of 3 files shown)


pkg/controller/vpc_nat_gw_eip.go
```diff
@@ -937,3 +962,46 @@ func (c *Controller) patchEipLabel(eipName string) error {
...
+	if err != nil {
+		return fmt.Errorf("get vpc nat gw %s: %w", natGwDp, err)
+	}
+	if gw.Spec.Vpc == "" || gw.Spec.LanIP == "" {
+		return nil
+	}
+	klog.Infof("add VPC static route %s/32 -> %s on vpc %s for eip %s", eipIP, gw.Spec.LanIP, gw.Spec.Vpc, eipIP)
+	return c.OVNNbClient.AddLogicalRouterStaticRoute(
+		gw.Spec.Vpc, util.MainRouteTable, "dst-ip", eipIP+"/32", nil, nil, gw.Spec.LanIP,
+	)
```
zbb88888, line 985: This looks like all the eips routes are exist, not jsut the dnat eip.

Is default route should be enough to cover these eip routes? https://kubeovn.github.io/docs/v1.15.x/en/vpc/vpc/#:~:text=flexible%20policy%20routes.-,Static%20Routes,-%C2%B6

zbb88888:
@andriishestakov Hi, Thanks for the PR, please fix the DCO lint. 

ps: The GW Pod has supported the harpin mode. is it enough to cover the same function too?

dist/images/vpcnatgateway/nat-gateway.sh
```diff
@@ -194,6 +194,15 @@ function init() {
         echo "IFB kernel module loaded successfully"
     fi
 
+    # MASQUERADE for DNAT loop-back: when a DNAT rule rewrites the destination to
```
SkalaNetworks, line 197: I'm pretty sure the HAIRPIN feature handles that ?

SkalaNetworks: Doesn't this also NAT everything going towards the VPC? No matter if it comes from the Internet or from the VPC itself? I feel like this is gonna NAT the source IP even if the connection comes from the internet

SkalaNetworks:
This has obviously been AI generated but the PR description doesn't actually say what bug it's trying to fix. I recommend not merging.

PR #6553, opened 2026-03-30
fix(dnat): support DNAT rules targeting switch_lb_vip VIPs

andriishestakov:
OVN's ls_in_pre_lb rule (priority 110) bypasses switch LBs for router→switch traffic, so a DNAT pointing at a switch_lb_vip VIP fails to reach pod backends — the chained LB resolution never happens.

Root cause: When OvnDnatRule.spec.ipName references a VIP of type switch_lb_vip, the existing code creates a standard router DNAT (EIP:port → VIP), but OVN silently skips the switch LB that maps the VIP to actual pods.

Fix: For switch_lb_vip DNATs, instead of a router DNAT rule, create two dedicated LBs that resolve pod backends directly:
- Router LB (<dnat-name>): EIP:externalPort → pod:targetPort — handles external/node traffic.
- Switch LB (<dnat-name>-sw) with hairpin_snat_ip: same mapping on the internal subnet handles same-subnet hairpin traffic, with SNAT ensuring replies traverse the router for correct conntrack.

Backends are kept current by re-enqueuing affected OvnDnatRules whenever their EndpointSlice or SwitchLBRule changes.

### DIFF
--- a/mocks/pkg/ovs/interface.go
+++ b/mocks/pkg/ovs/interface.go
@@ -1758,6 +1758,20 @@ func (mr *MockLoadBalancerMockRecorder) SetLoadBalancerPreferLocalBackend(lbName
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "SetLoadBalancerPreferLocalBackend", reflect.TypeOf((*MockLoadBalancer)(nil).SetLoadBalancerPreferLocalBackend), lbName, preferLocalBackend)
 }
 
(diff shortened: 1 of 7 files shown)


andriishestakov:
PR #6563 implements separate crd for router lb

zbb88888:
@andriishestakov Hi, Thanks for the PR! ovn dnat + snat + fip are based on router. so ovn dnat should not using in switch lb things!

PR #6563, opened 2026-04-01
feat(controller): add RouterLBRule CRD for L3/L4 load balancing on VPC router

andriishestakov:
RouterLBRule is a cluster-scoped CRD that exposes multiple pod backends behind a single OvnEip:port entry point on the VPC logical router — the "multi-backend" counterpart of OvnDnatRule.

How it works:
- Controller creates a headless Service annotated with ovn.kubernetes.io/router_lb_vip=<eip_ip> and ovn.kubernetes.io/logical_router=<vpc>
- The existing endpoint_slice controller detects the annotation and inserts the VIP into the VPC's shared router LBs (TCPLoadBalancer etc.)
- On add/update the controller also calls LogicalRouterUpdateLoadBalancers so the VPC LBs are attached to the logical router, enabling OVN ARP proxy for the VIP on the external switch (required for traffic from outside the VPC)
- Pod backends are discovered dynamically via label selector; static endpoints are also supported
- OVN LBHC health checks are wired up for IPv4 backends (same as SwitchLBRule)

Port conflict detection:
- isOvnDnatDuplicated is replaced by the shared checkEipPortConflict which validates against both RouterLBRules and OvnDnatRules, preventing two rules from claiming the same EIP:port on the same router


oujonny:
@copilot can you explain the main difference between this new RouterLBRule and the existing SwitchLBRule?

pkg/controller/router_lb_rule.go
```diff
@@ -0,0 +1,549 @@
...
+
+	"github.com/ovn-kubernetes/libovsdb/ovsdb"
+
+	kubeovnv1 "github.com/kubeovn/kube-ovn/pkg/apis/kubeovn/v1"
+
+	"github.com/kubeovn/kube-ovn/pkg/ovsdb/ovnnb"
+	"github.com/kubeovn/kube-ovn/pkg/util"
+)
+
+type RlrInfo struct {
```
zbb88888: how about using RouterLBRuleInfo ？

zbb88888:
Thanks for the PR. I reviewed it against current master. The feature looks useful, but I think there are still a few important issues to address before merging:

 1. RBAC/manifests are incomplete for the new RouterLBRule CRD
 The CRD is added, but I do not see router-lb-rules / router-lb-rules/status added to the RBAC/manifests consistently (charts/kube-ovn-v2/templates/rbac/ovn-CR.yaml,
 charts/kube-ovn/templates/ovn-CR.yaml, kube-ovn-sa.yaml, yamls/audit-policy.yaml).
 Since the controller now starts an informer for this resource, missing RBAC can cause informer/watch/cache-sync failures at runtime.
 2. Stale VIPs may remain after RouterLBRule delete/recreate
 RouterLBRule reuses the Service/EndpointSlice pipeline, but the service cleanup path still only handles SwitchLBRuleVipsAnnotation, not the new RouterLBRuleVipsAnnotation.
 As a result, when a RouterLBRule-backed headless Service is deleted or recreated (for example after changing spec.ovnEip, spec.vpc, or spec.namespace), old VIPs may not be removed from the
 shared VPC load balancers.
 Also, handleDelRouterLBRule() only cleans LBHC/IP-port mapping, but does not explicitly remove VIPs from the shared LBs.
 3. Dual-stack EIP handling looks incorrect
 The code joins v4/v6 EIP addresses into a single comma-separated string and stores it in the Service annotation, but handleUpdateEndpointSlice() treats that annotation as a single VIP
 instead of splitting it.
 This likely breaks dual-stack RouterLBRule behavior.
 4. Generic informer support seems incomplete
 I do not see router-lb-rules added in pkg/client/informers/externalversions/generic.go, so generic informer / ForResource() usage would not work for this new resource.
 5. Missing tests
 This PR adds a new CRD and controller flow, but I do not see corresponding unit/e2e coverage for:
     - create/delete cleanup
     - update/recreate cases
     - dual-stack behavior
     - conflict checks with OvnDnatRule
 6. Minor issue: in generateRlrEndpoints(), TargetRef.Namespace uses rlr.Namespace, but RouterLBRule is cluster-scoped, so that field is empty.

 The main blockers for me are:
 - incomplete RBAC/manifests
 - stale VIP cleanup on delete/recreate
 - incorrect dual-stack handling

andriishestakov:
Fixed mentioned issues and also switchlbrule had same issue with namespace.

pkg/controller/router_lb_rule.go
andriishestakov: RlrInfo is similar as SlrInfo, but possible to rename.

zbb88888: ok, I think RouterLBRuleInfo is better, because it is base struct, RouterLBRuleInfo is not too long.  


SlrInfo should be SwitchLBInfo too. if you have time, could you please fix it in a new PR? Thanks!

andriishestakov:
@zbb88888 Hey, any suggestions what could be done to get this PR accepted?

zbb88888:
> @zbb88888 Hey, any suggestions what could be done to get this PR accepted?

@andriishestakov Hi, I rebased your PR, so many e2e failed. 


@oilbeater could you help review this? Thanks!

andriishestakov:
@zbb88888 fixed and added new e2e tests for rlr

zbb88888:
> @zbb88888 fixed and added new e2e tests for rlr

Thanks

@oilbeater could you please reiveiw this? thanks!

andriishestakov:
Added fixes for ipv6 labels. Related to #6643 
But might be dome by separate PR.

andriishestakov:
@zbb88888 @oilbeater could you review this PR please?

andriishestakov:
Hey @zbb88888 could this change be merged?

zbb88888:
> Hey @zbb88888 could this change be merged?

Hi @andriishestakov Please fix the conflicts, I will review again!

andriishestakov:
@zbb88888 rebased, but few tests failing, not sure because of this change

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
```diff
@@ -2170,6 +2170,144 @@ spec:
...
+              service:
+                type: string
+            type: object
+        type: object
+    served: true
+    storage: true
+    subresources:
+      status: {}
+---
+---
```
zbb88888, line 2308: redundant line？

charts/kube-ovn/templates/kube-ovn-crd.yaml
```diff
@@ -2186,6 +2186,145 @@ spec:
...
+              service:
+                type: string
+            type: object
+        type: object
+    served: true
+    storage: true
+    subresources:
+      status: {}
+---
+---
```
zbb88888, line 2325: redundant line？

dist/images/install.sh
```diff
@@ -2555,6 +2555,144 @@ spec:
...
+                type: string
+              service:
+                type: string
+            type: object
+        type: object
+    served: true
+    storage: true
+    subresources:
+      status: {}
+---
```
zbb88888, line 2692: redundant line？

pkg/controller/router_lb_rule.go
```diff
@@ -0,0 +1,645 @@
...
+}
+
+func (c *Controller) handleDelRouterLBRule(info *RouterLBRuleInfo) error {
+	klog.V(3).Infof("handleDelRouterLBRule %s", info.Name)
+
+	svcName := generateRlrSvcName(info.Name)
+
+	var vips []string
+	vpcForRlr := ""
+	if svc, e := c.servicesLister.Services(info.Namespace).Get(svcName); e == nil {
```
zbb88888, line 361: should we log and return the error and trigger reconcile？

pkg/controller/router_lb_rule.go
```diff
@@ -0,0 +1,645 @@
...
+	if err := c.config.KubeClient.CoreV1().Services(info.Namespace).Delete(context.Background(), svcName, metav1.DeleteOptions{}); err != nil {
+		if !k8serrors.IsNotFound(err) {
+			klog.Errorf("failed to delete service %s: %v", svcName, err)
+			return err
+		}
+	}
+
+	var vpcLBNames set.Set[string]
+	if vpcForRlr != "" {
+		vpc, e := c.vpcsLister.Get(vpcForRlr)
```
zbb88888, line 383: why not just use err，to make the error values consistent

zbb88888 (approved):

zbb88888:
@oilbeater could you please reiveiw this? thanks!

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
andriishestakov: looks like issue is in hack/gen-crd.sh script. other crds has same duplication

zbb88888: ok， ignore it

pkg/controller/controller.go
```diff
@@ -1389,6 +1426,10 @@ func (c *Controller) startWorkers(ctx context.Context) {
 		// run in a single worker to avoid delete the last vip, which will lead ovn to delete the loadbalancer
 		go wait.Until(runWorker("delete service", c.deleteServiceQueue, c.handleDeleteService), time.Second, ctx.Done())
 
+		go wait.Until(runWorker("add/update router lb rule", c.addRouterLBRuleQueue, c.handleAddOrUpdateRouterLBRule), time.Second, ctx.Done())
```
oilbeater, line 1429: This should also be wrapped by `if c.config.EnableLb`

test/e2e/kube-ovn/router_lb_rule/router_lb_rule.go
```diff
@@ -0,0 +1,515 @@
...
+		lrpEipName = vpcName + "-" + externalSubnetName
+		stsName = "rlr-sts-" + suffix
+		stsSvcName = stsName
+		clientPodName = "rlr-client-" + suffix
+		selRlrName = "sel-rlr-" + suffix
+
+		if skip {
+			ginkgo.Skip("RouterLBRule e2e only runs on Kind clusters")
+		}
+		f.SkipVersionPriorTo(1, 16, "RouterLBRule was introduced in v1.16")
```
oilbeater: This should be 1.17

pkg/controller/ovn_dnat.go
oilbeater: This should also use `util.IPv6ToLabelValue`

pkg/controller/router_lb_rule.go
```diff
@@ -0,0 +1,643 @@
...
+			}
+		}
+		if vpcForRlr = svc.Annotations[util.VpcAnnotation]; vpcForRlr == "" {
+			vpcForRlr = svc.Annotations[util.LogicalRouterAnnotation]
+		}
+	} else if !k8serrors.IsNotFound(e) {
+		klog.Warningf("failed to get service %s for cleanup enrichment: %v", svcName, e)
+	}
+
+	if err := c.config.KubeClient.CoreV1().Services(info.Namespace).Delete(context.Background(), svcName, metav1.DeleteOptions{}); err != nil {
```
oilbeater: We should cleanup ovn resources first before cleanup k8s resources to avoid resource leak.

oilbeater: And I think we should also add gc to prevernt resource leak if the kube-ovn-controller is down and in the meantime user manually deletes RouterLB.

oilbeater (changes requested):

pkg/controller/ovn_eip.go
```diff
@@ -21,9 +21,11 @@ import (
 )
 
 func (c *Controller) enqueueAddOvnEip(obj any) {
-	key := cache.MetaObjectToName(obj.(*kubeovnv1.OvnEip)).String()
+	eip := obj.(*kubeovnv1.OvnEip)
+	key := cache.MetaObjectToName(eip).String()
 	klog.Infof("enqueue add ovn eip %s", key)
 	c.addOvnEipQueue.Add(key)
+	c.requeueRouterLBRulesForEip(eip.Name, false)
```
oilbeater, line 28: **Bug: panic crashloop when `--enable-lb=false`**

`requeueRouterLBRulesForEip` accesses `routerLBRuleLister` and the three RLR queues, but these are only initialized inside `if config.EnableLb` (controller.go:1293-1309). OvnEip event handlers (controller.go:944-950) are registered unconditionally, so at startup every existing OvnEip triggers AddFunc → nil pointer access → panic → crashloop.

This is the exact same pattern that was fixed in PR #6845 (dns-resolver + nil ANP queue). The fix: add `if !c.config.EnableLb { return }` as the first line of `requeueRouterLBRulesForEip`.

pkg/controller/ovn_dnat.go
```diff
@@ -527,13 +503,13 @@ func (c *Controller) patchOvnDnatStatus(key, vpcName, v4Eip, v6Eip, internalV4Ip
 		needUpdateLabel = true
 		dnat.Labels = map[string]string{
 			util.EipV4IpLabel: v4Eip,
-			util.EipV6IpLabel: v6Eip,
+			util.EipV6IpLabel: util.IPv6ToLabelValue(v6Eip),
 		}
 	} else if dnat.Labels[util.EipV4IpLabel] != v4Eip || dnat.Labels[util.EipV6IpLabel] != v6Eip {
```
oilbeater: **Comparison still against raw IPv6 address**

The `else if` condition compares against the raw IPv6 address (e.g. `"fd00::1"`), but the label stores the sanitized value (`"fd00--1"` via `IPv6ToLabelValue`). This means the condition is *always* true and the replace branch fires on every reconcile. The write below was correctly changed to `IPv6ToLabelValue`, so the net effect is a no-op patch each cycle — not a crash, but an unnecessary API call every time any DNAT rule reconciles.

Should be: `dnat.Labels[util.EipV6IpLabel] != util.IPv6ToLabelValue(v6Eip)`

pkg/controller/ovn_snat.go
```diff
@@ -388,13 +388,13 @@ func (c *Controller) patchOvnSnatStatus(key, vpc, v4Eip, v6Eip, v4IpCidr, v6IpCi
 		needUpdateLabel = true
 		snat.Labels = map[string]string{
 			util.EipV4IpLabel: v4Eip,
-			util.EipV6IpLabel: v6Eip,
+			util.EipV6IpLabel: util.IPv6ToLabelValue(v6Eip),
 		}
 	} else if snat.Labels[util.EipV4IpLabel] != v4Eip || snat.Labels[util.EipV6IpLabel] != v6Eip {
```
oilbeater: **Same comparison issue as ovn_dnat.go:508**

The label stores the value from `IPv6ToLabelValue(v6eip)` but the comparison checks the raw v6 IP. This triggers the replace branch on every reconcile.

Should be: `snat.Labels[util.EipV6IpLabel] != util.IPv6ToLabelValue(v6Eip)`

oilbeater (commented):
Follow-up review of the force-push (commit 5816f1e975c84a6dba67d826ee968f2b6eee2d2a). Three items remain unfixed — one of which is a real crashloop bug.

oilbeater (approved):

PR #6629, opened 2026-04-13
Allow set external-gateway-switch in chart

andriishestakov:
Allow set external gateway switch name in chart


SkalaNetworks:
Thanks!

PR #6631, opened 2026-04-14
refactor(api): make pkg/apis independently importable as a Go module

abhishek-pandey-1:
# Pull Request

## What type of this PR

- Refactor

Today, consuming API structs for kubeovn result in the consumer inheriting all dependencies of kubeovn. This creates unnecessary coupling between the consumer's and kubeovn's dependencies. The goal here is to breakout api as a separate module that can be consumed with minimal dependencies.

Extract pkg/apis into its own Go module so external consumers can import API types without pulling in the entire kube-ovn dependency tree.

Changes:
- Create pkg/apis/go.mod with module github.com/kubeovn/kube-ovn/pkg/apis
- Move BigInt type to pkg/apis/types/ (from pkg/internal/) so it lives within the apis module boundary
- Inline OVN constants in vpc.go and security-group.go to remove the dependency on pkg/ovsdb/ovnnb
- Update subnet.go to use pkg/apis/types for BigInt
- Re-export BigInt as a type alias in pkg/internal/ for backward compatibility with existing consumers
- Add replace directive in root go.mod to use the local apis module

Fixes #(issue-number)


zbb88888 (approved):

oilbeater:
@abhishek-pandey-1  I think we need to update the ut, lint, and build-go make targets to cover the new module. We also need a mechanism to keep the two go.mod files aligned between this module and the main kube-ovn module. Any ideas on this?

abhishek-pandey-1:
> @abhishek-pandey-1 I think we need to update the ut, lint, and build-go make targets to cover the new module. We also need a mechanism to keep the two go.mod files aligned between this module and the main kube-ovn module. Any ideas on this?

Hey - yes, I agree - I am working on more changes trying to address this. The replace statement will always keep the modules in sync from kubeovn's POV. For projects that depend on kubeovn, I was thinking about updating the release target to bump the api module version to the release tag so that anyone depending on kubeovn x.x.x would pull in api x.x.x as well.

I will update the PR next week to bring in these changes.

PR #6645, opened 2026-04-17
Fix Helm chart CNI config directory override

### LINKED ISSUE #6644: [BUG] Helm chart: CNI_CONF_DIR env override not reflected in install-cni volumeMount
### Kube-OVN Version

v1.13.0

### Kubernetes Version

v1.34.6+rke2r3 (RKE2)

Ubuntu 22.04.5 LTS, kernel 5.15.0-173-generic, amd64

### Description

When the Helm chart is installed with `cni_conf.CNI_CONF_DIR` set to a non-default path, the `install-cni` init container's `cni-conf` volumeMount is **not** updated to match. The binary tries to write the conflist to a path that doesn't exist inside the container.

This breaks RKE2 installations following the documented pattern of using `/var/lib/rancher/rke2/agent/etc/cni/net.d` as the CNI config dir.

### Reproduction

On RKE2 v1.31+ with `cni: none`:

1. Pre-create the target directory on every node (via cloud-init or DaemonSet):
   ```
   mkdir -p /var/lib/rancher/rke2/agent/etc/cni/net.d
   ```
2. Install Kube-OVN with the RKE2-specific paths override:
   ```bash
   helm upgrade --install kube-ovn kubeovn/kube-ovn \
     --namespace kube-system \
     --version 1.13.0 \
     --set MASTER_NODES="<master-ip>" \
     --set kubelet_conf.KUBELET_DIR=/var/lib/rancher/rke2/agent/kubelet \
     --set cni_conf.CNI_CONF_DIR=/var/lib/rancher/rke2/agent/etc/cni/net.d \
     --set cni_conf.CNI_BIN_DIR=/opt/cni/bin
   ```
3. Observe that `kube-ovn-cni` pods CrashLoopBackOff in their `install-cni` init container:
   ```
   Installing cni config file "/kube-ovn/01-kube-ovn.conflist"
     to "/var/lib/rancher/rke2/agent/etc/cni/net.d/01-kube-ovn.conflist"
   E failed to mv cni config file
     err="open /var/lib/rancher/rke2/agent/etc/cni/net.d/01-kube-ovn.conflist:
          no such file or directory"
   ```

### Root cause

The `install-cni` container's binary targets the path from the `CNI_CONF_DIR` env var, which is correctly set from Helm values. However the volumeMount for the `cni-conf` volume is hardcoded at `/etc/cni/net.d`:

```yaml
# from the rendered DaemonSet
initContainers:
- name: install-cni
  volumeMounts:
  - name: cni-conf
    mountPath: /etc/cni/net.d   # ← hardcoded
volumes:
- name: cni-conf
  hostPath:
    path: /var/lib/rancher/rke2/agent/etc/cni/net.d  # ← host path matches the env
```

So the binary tries to write to `/var/lib/rancher/rke2/agent/etc/cni/net.d/01-kube-ovn.conflist` **inside the container**, but that directory doesn't exist — only `/etc/cni/net.d/` is mounted.

### Expected behavior

The chart template should render `mountPath` using the same `CNI_CONF_DIR` value that is set as the env var. Effectively, the container path should follow the configured target path.

### Workaround (validated)

```bash
kubectl patch ds -n kube-system kube-ovn-cni --type='json' -p='[
  {"op": "replace",
   "path": "/spec/template/spec/initContainers/1/volumeMounts/1/mountPath",
   "value": "/var/lib/rancher/rke2/agent/etc/cni/net.d"}
]'
kubectl delete pods -n kube-system -l app=kube-ovn-cni
```

After this patch the install succeeds and `01-kube-ovn.conflist` appears on the host at the configured path.

### Alternative workaround (cleaner, no kubectl patch)

On the hosts, create a symlink so both container paths resolve to the same directory:

```bash
mkdir -p /etc/cni
[ -d /etc/cni/net.d ] && [ -z "$(ls -A /etc/cni/net.d 2>/dev/null)" ] && rmdir /etc/cni/net.d
[ ! -e /etc/cni/net.d ] && ln -sfn /var/lib/rancher/rke2/agent/etc/cni/net.d /etc/cni/net.d
```

Then the chart can be installed without the `cni_conf.CNI_CONF_DIR` override. But that masks the underlying chart bug.

### Proposed fix

In the Kube-OVN Helm chart DaemonSet template for `kube-ovn-cni`, change:

```yaml
volumeMounts:
- name: cni-conf
  mountPath: /etc/cni/net.d
```

to:

```yaml
volumeMounts:
- name: cni-conf
  mountPath: {{ .Values.cni_conf.CNI_CONF_DIR | default "/etc/cni/net.d" }}
```

This matches the same value that's already used:
- In the `CNI_CONF_DIR` env var of the `install-cni` container
- In the `hostPath` spec of the `cni-conf` volume

### Impact

Every RKE2 production deployment of Kube-OVN follows the documented pattern and hits this issue. Reported by multiple teams as "Nodes stay NotReady despite Kube-OVN components all Running".

Happy to submit a PR if maintainers agree on this approach.

utafrali:

## What type of this PR

- Bug fixes

Fixes #6644

---

The charts had two separate values for the CNI config directory - one for the command argument, one for the volumeMount. If someone tried to override it via environment variables, only one would actually change.

Consolidated to a single value so overrides work consistently across both the command and the mount. Removed the duplicate `mountConfigDirectory` and `MOUNT_CNI_CONF_DIR` fields while we're at it.

### DIFF
--- a/charts/kube-ovn-v2/templates/agent/agent-daemonset.yaml
+++ b/charts/kube-ovn-v2/templates/agent/agent-daemonset.yaml
@@ -82,7 +82,7 @@ spec:
         imagePullPolicy: {{ .Values.image.pullPolicy }}
         command:
           - /kube-ovn/install-cni.sh
-          - --cni-conf-dir={{ .Values.cni.mountConfigDirectory }}
+          - --cni-conf-dir={{ .Values.cni.configDirectory }}
           - --cni-conf-file={{ .Values.cni.localConfigFile }}
           - --cni-conf-name={{- .Values.cni.configPriority -}}-kube-ovn.conflist
         env:
@@ -96,7 +96,7 @@ spec:
         volumeMounts:
           - mountPath: /opt/cni/bin
             name: cni-bin
-          - mountPath: {{ .Values.cni.mountConfigDirectory }}
+          - mountPath: {{ .Values.cni.configDirectory }}
             name: cni-conf
           {{- if .Values.cni.mountToolingDirectory }}
           - mountPath: /usr/local/bin
--- a/charts/kube-ovn-v2/values.yaml
+++ b/charts/kube-ovn-v2/values.yaml
@@ -241,9 +241,15 @@ features:
 # @section -- CNI configuration
 # @default -- "{}"
 cni:
-  # -- Location of the CNI configuration on the node.
+  # -- Location of the CNI configuration on the node. This value controls both
+  # the hostPath source and the pod mountPath for the CNI configuration.
   # @section -- CNI configuration
   configDirectory: "/etc/cni/net.d"
+  # -- Deprecated: compatibility alias retained for chart consumers upgrading
+  # from releases that still set this value. Use `configDirectory` instead.
+  # Keep for at least one release so upgrades do not silently ignore existing configuration.
+  # @section -- CNI configuration
+  mountConfigDirectory: false
   # -- Location on the node where the agent will inject the Kube-OVN binary.
   # @section -- CNI configuration
   binaryDirectory: "/opt/cni/bin"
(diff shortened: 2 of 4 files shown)


utafrali:
Done, pushed the fix.

PR #6649, opened 2026-04-19
perf(controller): deep-trim pod informer cache

oilbeater:
## Summary

- Extend the controller's pod informer `WithTransform` with `TrimPodForController`, a deep trim that drops pod fields the controller never reads (Volumes, Tolerations, container Command/Env/Probes/SecurityContext, InitContainerStatuses, etc.).
- Preserve every field still consumed by the controller: `Spec.NodeName`, `HostNetwork`, `RestartPolicy`, container `Name`+`Ports`, init-container `Name`+`Ports`+`RestartPolicy`, `Status.Phase`/`PodIP(s)`/`HostIP`/`Reason`, `ContainerStatuses[].Name`/`RestartCount`/`State` (NAT gateway redo reads `State.Running.StartedAt`), and `Conditions[].Type`/`Status`/`LastTransitionTime`. Non-pod objects fall back to `TrimManagedFields`.
- Benchmarked savings: per-pod cached footprint drops from ~6471 B → ~3004 B (~54%). Projected cluster-level savings: ~30 MB for 10k pods, ~150 MB for 50k pods. Daemon informer is intentionally untouched (still reads `Spec.Volumes` and container probes).

## Test plan

- [x] `go test ./pkg/util/ ./pkg/controller/ ./pkg/daemon/`
- [x] `make lint` (0 issues)
- [x] `BenchmarkPodInformerTrim` reports stable bytes/pod delta across 5 runs
- [x] Audit controller reads of every trimmed field (Codex review caught `ContainerStatuses.State.Running.StartedAt`, now preserved)
- [ ] CI green

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -348,7 +348,7 @@ func Run(ctx context.Context, config *Configuration) {
 	}
 
 	informerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(config.KubeFactoryClient, 0,
-		kubeinformers.WithTransform(util.TrimManagedFields),
+		kubeinformers.WithTransform(util.TrimPodForController),
 		kubeinformers.WithTweakListOptions(func(listOption *metav1.ListOptions) {
 			listOption.AllowWatchBookmarks = true
 		}))
--- a/pkg/util/k8s.go
+++ b/pkg/util/k8s.go
@@ -247,3 +247,130 @@ func TrimManagedFields(obj any) (any, error) {
 	}
 	return obj, nil
 }
+
+// TrimPodForController is an informer transform that strips pod fields the
+// controller never reads, in addition to clearing managed fields. It shrinks
+// the pod informer cache footprint on large clusters.
+//
+// Preserved (required by controller code paths):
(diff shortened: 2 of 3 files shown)


oilbeater:
Thanks for the review! Addressing the three comments:

1. **`new(true)` on `pkg/util/k8s_test.go:920/883/924`** — this is actually valid Go 1.26 syntax. The Go 1.26 spec extended the built-in `new` to accept a value expression (`new(v)` returns `*T` with value `v`, equivalent to `ptr.To(v)`). The repo's go.mod declares `go 1.26.2`, and these calls were auto-inserted by `gopls modernize` during `make lint`. Tests and `make lint` pass; reverting to `ptr.To(...)` would immediately be rewritten by the modernize linter on the next run.

2. **Benchmark ignoring `b.N`** (valid) — fixed in 62791cd6a by wrapping the work in `for b.Loop() { ... }` so the runner controls iteration count under the default time-based benchtime.

3. **`HeapAlloc` underflow** (valid) — fixed in 62791cd6a by casting the delta through `int64` and clamping negative values to 0.

PR #6659, opened 2026-04-20
ovn_eip: rewrite IPv6 addresses for label-value safety when creating OvnEip CRs

### LINKED ISSUE #6643: [BUG] VPC enableExternal with IPv6 external subnet crashes
### Kube-OVN Version

1.16.0

### Kubernetes Version

v1.34.6

Debian 13, kernel 6.17

### Description

When a VPC sets `spec.enableExternal: true` and the external subnet is IPv6 (or dual-stack), the controller crashes in a loop trying to create the `OvnEip` CR for the VPC's Logical Router Port (LRP).

 `createOrUpdateOvnEipCR()` in `pkg/controller/ovn_eip.go` stores the allocated IPv6 address as a Kubernetes label value(`ovn.kubernetes.io/eip_v6_ip`). IPv6 addresses contain `:` which is an illegal character in label values — Kubernetes labels only allow `[a-zA-Z0-9._-]`.

### Steps To Reproduce

```
apiVersion: kubeovn.io/v1
kind: Subnet
metadata:
  name: external
spec:
  protocol: IPv6
  cidrBlock: xxx
  gateway: xxx
  excludeIps:
    - xxx
  vlan: vlan-external
  natOutgoing: false
  disableGatewayCheck: true
---
apiVersion: kubeovn.io/v1
kind: Vpc
metadata:
  name: tenant-vpc
spec:
  enableExternal: true
  namespaces:
    - tenant
```

### Current Behavior

in controller (ip replaced by placeholder)
```
OvnEip.kubeovn.io "tenant-vpc-external" is invalid:
  metadata.labels: Invalid value: "1001:11d0:303:21c1:2000::100":
    a valid label must be an empty string or consist of alphanumeric characters,
    '-', '_' or '.', and must start and end with an alphanumeric character
    (e.g. 'MyValue', or 'my_value', or '12345',
    regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?') 
```
### Expected Behavior

no issues in controller

SAY-5:
## Summary

`createOrUpdateOvnEipCR` stores the allocated V6 IP straight into the `EipV6IpLabel` label value. Kubernetes label values only permit `[a-zA-Z0-9._-]`, so any IPv6 address (which contains `:`) makes the OvnEip Create fail with:

```
OvnEip.kubeovn.io "tenant-vpc-external" is invalid:
  metadata.labels: Invalid value: "1001:11d0:303:21c1:2000::100":
    a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.',
    and must start and end with an alphanumeric character
```

The controller then CrashLoops whenever a VPC sets `enableExternal` with an IPv6 or dual-stack external subnet (#6643).

Fixes #6643.

## Change

Add a small `util.IPForLabel` helper that:

- Returns empty input and IPv4 addresses unchanged.
- Replaces `:` with `.` for IPv6 so the string only contains `[a-zA-Z0-9.]`.
- Pads any leading/trailing `.` (which comes from `::` zero-compression) with `v6` so the result still starts and ends with alphanumerics.

Use it at both the Create and Update call sites in `ovn_eip.go`. The V4 selector at `ovn_eip.go:~610` only uses `EipV4IpLabel`, so it's unaffected.

Examples:

| IP | IPForLabel |
| --- | --- |
| `10.0.0.1` | `10.0.0.1` |
| `2001:db8::1` | `2001.db8..1` |
| `::1` | `v6..1` |
| `2001::` | `2001..v6` |

## Testing

- `make test` passes.
- Added a local regression: applying the reporter's VPC + IPv6 subnet manifest now creates the `OvnEip` CR successfully; previously the Create was rejected by the API server and the controller crashed.

### DIFF
--- a/pkg/controller/ovn_eip.go
+++ b/pkg/controller/ovn_eip.go
@@ -308,7 +308,10 @@ func (c *Controller) createOrUpdateOvnEipCR(key, subnet, v4ip, v6ip, mac, usageT
 						util.SubnetNameLabel: subnet,
 						util.OvnEipTypeLabel: usageType,
 						util.EipV4IpLabel:    v4ip,
-						util.EipV6IpLabel:    v6ip,
+						// IPv6 addresses contain ':' which is not a valid
+						// label-value character; rewrite via util.IPForLabel
+						// so the CR Create doesn't fail (#6643).
+						util.EipV6IpLabel: util.IPForLabel(v6ip),
 					},
 				},
 				Spec: kubeovnv1.OvnEipSpec{
@@ -348,7 +351,7 @@ func (c *Controller) createOrUpdateOvnEipCR(key, subnet, v4ip, v6ip, mac, usageT
 		ovnEip.Labels[util.SubnetNameLabel] = subnet
 		ovnEip.Labels[util.OvnEipTypeLabel] = usageType
 		ovnEip.Labels[util.EipV4IpLabel] = v4ip
-		ovnEip.Labels[util.EipV6IpLabel] = v6ip
+		ovnEip.Labels[util.EipV6IpLabel] = util.IPForLabel(v6ip)
 
 		needUpdate := false
 		if mac != "" && ovnEip.Spec.MacAddress != mac {
--- a/pkg/util/label.go
+++ b/pkg/util/label.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"fmt"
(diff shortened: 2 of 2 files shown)


oilbeater:
AI slop and a wastes of time.

PR #6661, opened 2026-04-21
feat(vpcnatgateway): add high-availability for nat gateways

SkalaNetworks:
This PR adds the first feature necessary to make VPC NAT gateways HA:
- It adds a `.replicas` field to NAT Gateways to provision more than one
- It adds a `.bfd` field to NAT Gateways to provision enable BFD sidecars
- It adds new fields in `.status` to reflect the NAT gateway HA state

This is very much like what the VPC Egress Gateway does. You can setup your NAT gateways to run with more than one instance (through a deployment), and the routes/BFDs are automatically injected in the VPC to route the traffic to the gateway.

One problem not yet addressed is that the conntrack is not shared. Which means that if the return traffic doesn't get sent to the same gateway that initially did the NATing, the asymmetry will trigger TCP RSTs.

Next steps will be:
- include conntrackd to sync-up NAT gateways
- add a passive/active feature so that only one gateway is working at a time, with a small delay and RSTs in case of one gateway dying

Example of a VPC NAT gateway with HA + BGP:
```yaml
---
apiVersion: kubeovn.io/v1
kind: VpcNatGateway
metadata:
  name: example-nat-gw
spec:
  vpc: example-vpc
  subnet: example-subnet

  # New field, any subnet included here will have a route injected in its VPC to forward traffic to the NAT GW
  internalSubnets:
    - example-subnet
  # New field, any CIDR included here will have a route injected in its VPC to forward traffic to the NAT GW
  # Cumulative with internalSubnets
  internalCIDRs:
    - 10.0.0.0/24   
  # New field, if unspecified or set to 1, the legacy system is used with a statefulset, otherwise, we use a deployment
  replicas: 2

  # LanIP doesn't need to be specified in HA, they are picked automatically using the subnet IPAM
  # lanIp: ""

  # BFD to enable fast failover, must be configured on the VPC aswell
  bfd:
    enabled: false
    minRX: 1000
    minTX: 1000
    multiplier: 3

  
  externalSubnets:
    - spx-internal-bgp
  bgpSpeaker:
    asn: 65500
    enableGracefulRestart: false
    enabled: true
    extraArgs:
    - -v5
    - --graceful-restart
    holdTime: 5s
    neighbors:
    - 172.17.0.1
    - fd00:0:0:ffff::1
    remoteAsn: 65000
```


SkalaNetworks:
@oilbeater  Hi, this is the first PR in the road to making NAT gateways HA.

Nex on the line:
- adding conntrackd
- making an optional switch to render gateways active/passive or active/active

SkalaNetworks:
@oilbeater have you had any time to look at this PR? Thanks

pkg/controller/controller.go
```diff
@@ -363,12 +359,11 @@ func Run(ctx context.Context, config *Configuration) {
 		kubeinformers.WithTweakListOptions(func(listOption *metav1.ListOptions) {
 			listOption.AllowWatchBookmarks = true
 		}))
-	// deployment informer used to list/watch vpc egress gateway workloads
+	// deployment informer used to list/watch vpc egress gateway and nat gateway workloads
 	deployInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(config.KubeFactoryClient, 0,
 		kubeinformers.WithTransform(util.TrimManagedFields),
 		kubeinformers.WithTweakListOptions(func(listOption *metav1.ListOptions) {
 			listOption.AllowWatchBookmarks = true
```
oilbeater, line 373: We should prevent watch all deployment in the cluster, it may consume lots of memory in large scale cluster.

pkg/controller/vpc_nat_gateway.go
```diff
@@ -1161,29 +1315,234 @@ func (c *Controller) genNatGwStatefulSet(gw *kubeovnv1.VpcNatGateway, oldSts *v1
...
 	return sts, nil
 }
 
-// getExternalSubnetNad returns the namespace and name of the NetworkAttachmentDefinition associated with
-// an external network attached to a NAT gateway
-func (c *Controller) getExternalSubnetNad(gw *kubeovnv1.VpcNatGateway) (string, string, error) {
-	externalNadNamespace := c.config.PodNamespace
-	// GetNatGwExternalNetwork returns the subnet name from ExternalSubnets, or "ovn-vpc-external-network" if empty
-	externalSubnetName := util.GetNatGwExternalNetwork(gw.Spec.ExternalSubnets)
+// genNatGwDeployment generates a Deployment for HA mode NAT gateway (replicas > 1).
```
oilbeater, line 1326: Why use deployment instead statefulset for ha gw?

oilbeater (changes requested):

pkg/controller/controller.go
SkalaNetworks: I will see how to minimize the amount of deployments we're watching

pkg/controller/vpc_nat_gateway.go
SkalaNetworks: I did the same thing as the Egress Gateway did, switch from statefulsets to deployments.

The rationale is pretty  simple, sts have a major flaw, if the node on which they are crashes without a proper shutdown, statefulsets are never re-scheduled.

So if you're going with a 2-replicas gateway (NAT or egress), and a node crashes, you do get a second gateway rescheduled on failure to ensure n+1 redundancy.

Statefulsets are hard to work with and to update, which is why deployments seem like the preferred way of doing this.

oilbeater: That's reasonable. I think we should use deployment even for single replicas gw to avoid the scheduler issue.

pkg/controller/controller.go
SkalaNetworks: @oilbeater I added a filter to ensure only the VPC Egress/NAT gateways are cached in the informer

SkalaNetworks:
@oilbeater I'd say this is ready for merge and for me to go to phase 2 where I implement the conntrack sharing

oilbeater:
we should run hack/update-codegen.sh to generate new deepcopy.

SkalaNetworks:
@oilbeater done and pushed!

oilbeater (approved):

rrajendran17:
@oilbeater Will there be a new minor or patch release with this PR changes ? (Like 1.16.x or v1.17.x)
The latest we see is as of v1.16.2 and we would like have this PR either in v1.16.x or v1.17.x. Thanks

rrajendran17:
> @oilbeater Will there be a new minor or patch release with this PR changes ? (Like 1.16.x or v1.17.x) The latest we see is as of v1.16.2 and we would like have this PR either in v1.16.x or v1.17.x. Thanks

@zbb88888 do you know about this ? Thanks

zbb88888:
> > @oilbeater Will there be a new minor or patch release with this PR changes ? (Like 1.16.x or v1.17.x) The latest we see is as of v1.16.2 and we would like have this PR either in v1.16.x or v1.17.x. Thanks
> 
> @zbb88888 do you know about this ? Thanks

@rrajendran17 Hi, Big feature will release in latest version as always.

rrajendran17:
> > > @oilbeater Will there be a new minor or patch release with this PR changes ? (Like 1.16.x or v1.17.x) The latest we see is as of v1.16.2 and we would like have this PR either in v1.16.x or v1.17.x. Thanks
> > 
> > 
> > @zbb88888 do you know about this ? Thanks
> 
> @rrajendran17 Hi, Big feature will release in latest version as always.

@zbb88888 so you mean this will be available in v1.17.x ? when can we expect this release ? Or will this fix be backported to v1.16.x ?

zbb88888:
@SkalaNetworks @rrajendran17 Hi, The PR looks like a feature, and it is huge.I think not just a bug fix. 

So, we will release it in next new main version v1.17.0.

rrajendran17:
> @SkalaNetworks @rrajendran17 Hi, The PR looks like a feature, and it is huge.I think not just a bug fix.
> 
> So, we will release it in next new main version v1.17.0.

Thanks @zbb88888 , so considering the release cadence for minor feature releases, can I expect this(v1.17.0) around  mid October ?

zbb88888:
> > @SkalaNetworks @rrajendran17 Hi, The PR looks like a feature, and it is huge.I think not just a bug fix.
> > So, we will release it in next new main version v1.17.0.
> 
> Thanks @zbb88888 , so considering the release cadence for minor feature releases, can I expect this(v1.17.0) around mid October ?

@zhangzujian Hi, Could you please decide when will release next new main version v1.17.0?

PR #6666, opened 2026-04-21
fix(controller): keep VM LSP port-group memberships when a sibling pod is alive

### LINKED ISSUE #6665: [BUG] VM loses port group memberships when a completed virt-launcher pod is GC'd (keepVMIP)
### Kube-OVN Version

v1.15.3

### Kubernetes Version

v1.32.x

Any (issue is in controller logic).

### Description

When `EnableKeepVMIP=true`, the single LSP per VM has `ExternalIDs["pod"]` set to `namespace/VM_NAME` (not the virt-launcher pod name), because `getNameByPod` returns the VM name for VM pods and that value is passed to `buildLogicalSwitchPort` via `CreateLogicalSwitchPort(..., podName, namespace, ...)` (see `pkg/ovs/ovn-nb-logical_switch_port.go:72` and `pkg/controller/pod.go:2420`).

`handleDeletePod` (`pkg/controller/pod.go:1002`) computes `podKey = namespace/VM_NAME` (line 1038), then the VM branch looks up the LSP with `ListNormalLogicalSwitchPorts(true, {"pod": podKey})` (line 1066). Because that external ID is shared across every virt-launcher pod of the VM, the lookup matches the VM's LSP for the deletion of **any** virt-launcher pod — including a completed migration source that kubernetes GCs after a successful live migration.

When the VM is still alive, `isVMToDel` returns false, `keepIPCR=true`, and the handler calls `RemovePortFromPortGroups(port.Name)` with no port-group names (line 1112), which removes the LSP from **all** port groups (`pkg/ovs/ovn-nb-port_group.go:336` falls back to `ListPortGroups(nil)` when no names are passed).

The LSP is still in use by the currently-running destination pod, so the active VM pod loses its memberships in node port groups, subnet port groups, security groups, and network policy port groups. The VM loses connectivity. The active pod does not re-reconcile on its own (no spec change), so the state persists until the kube-ovn-controller is restarted and all pods are re-synced.

### Steps To Reproduce

1. Deploy Kube-OVN with `ENABLE_KEEP_VM_IP=true`.
2. Create a `VirtualMachine` with KubeVirt and start it.
3. Live-migrate the VM at least once (`virtctl migrate ...`) and wait for the migration to complete.
4. Wait for kubernetes to GC the completed source `virt-launcher-*` pod (or delete it manually).
5. Observe the VM's networking break while the VMI is still Running.

### Current Behavior

Controller log excerpt on completed source pod deletion:

```
handle delete pod <ns>/virt-launcher-<vm>-<oldsuffix>
handle deletion of vm pod <ns>/<vm>
try keep ip for vm pod <ns>/<vm>
remove lsp <vm>.<ns> from all port groups
```

The LSP is stripped from every port group, including those owned by the destination pod's node. The VM's security groups, NP rules, and node-scoped routing/ACLs are gone until kube-ovn-controller is restarted (which triggers a full resync and re-adds the active pod to its port groups).

### Expected Behavior

Deletion of a non-active virt-launcher pod (completed migration source, or any GC'd sibling) must not affect the VM's LSP port-group memberships while another alive virt-launcher pod for the same VM exists.

### Proposed fix

In the VM branch of `handleDeletePod`, before calling `RemovePortFromPortGroups`, check whether another alive virt-launcher pod exists for the same VMI. If yes, skip the port-group cleanup — the memberships belong to the active pod, not the one being deleted.

Happy to open a PR with this change.

kvaps:
# Pull Request

## What type of this PR

- Bug fixes
- Tests

For VM pods with `EnableKeepVMIP=true` a single LSP is shared across every virt-launcher pod of the VM (`ExternalIDs["pod"]` is keyed by VM name, not by pod name). When kubernetes GCs a completed virt-launcher pod — most commonly the source of a successful live migration — `handleDeletePod` finds that LSP, and because `isVMToDel` is false it calls `RemovePortFromPortGroups(port.Name)` with no port-group names, which strips the LSP from every port group (node, subnet, security group, network policy).

The LSP is still in use by the currently running destination pod, so the active VM pod loses its memberships and connectivity until `kube-ovn-controller` is restarted and all pods are re-synced.

Before stripping port-group memberships in the `keepIPCR` branch, check whether another alive virt-launcher pod exists for the same VMI. If so, the memberships belong to that sibling and must be preserved. Added a unit test covering the sibling-detection logic.

Fixes #6665

### DIFF
--- a/pkg/controller/pod.go
+++ b/pkg/controller/pod.go
@@ -1222,7 +1222,8 @@ func (c *Controller) handleDeletePod(key string) (err error) {
 	}
 	if keepIPCR {
 		for _, port := range ports {
-			if vmOrphanedPorts[port.Name] {
+			switch {
+			case vmOrphanedPorts[port.Name]:
 				// Orphaned attachment LSP from NAD hotplug: delete and release its IP.
 				klog.Infof("delete orphaned vm attachment lsp %s", port.Name)
 				if err := c.OVNNbClient.DeleteLogicalSwitchPort(port.Name); err != nil {
@@ -1249,9 +1250,9 @@ func (c *Controller) handleDeletePod(key string) (err error) {
 						c.updateSubnetStatusQueue.Add(subnetName)
 					}
 				}
-			} else if hasAliveVMSibling {
+			case hasAliveVMSibling:
 				klog.Infof("skip removing lsp %s from port groups: another alive virt-launcher pod exists for vm %s/%s", port.Name, pod.Namespace, vmName)
-			} else {
+			default:
 				klog.Infof("remove lsp %s from all port groups", port.Name)
 				if err = c.OVNNbClient.RemovePortFromPortGroups(port.Name); err != nil {
 					klog.Errorf("failed to remove lsp %s from all port groups: %v", port.Name, err)
--- a/pkg/controller/pod.go
+++ b/pkg/controller/pod.go
@@ -1166,6 +1166,7 @@ func (c *Controller) handleDeletePod(key string) (err error) {
 		return err
 	}
 
+	var hasAliveVMSibling bool
 	isVMPod, vmName := isVMPod(pod)
 	if isVMPod && c.config.EnableKeepVMIP {
 		for _, port := range ports {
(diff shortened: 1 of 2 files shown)


oilbeater:
@kvaps LGTM can you also fix the lint issue?

zbb88888 (approved):

zbb88888:
好像有点问题，不清理，源旧 vm 所在节点会残留脏数据（由于旧节点没有 vm 不会触发网络问题），好像应该实现为从旧节点的 port-group 中清理。

PR #6716, opened 2026-05-11
feat(ovn-central): add ovn-central-controller for dynamic-peer lifecycle management

andriishestakov:
Summary

Introduces a new ovn-central-controller binary that owns the full OVN central DB lifecycle (preflight, bring-up, runtime watchdog) when DYNAMIC_PEERS=true. It replaces the dynamic-peers branch of start-db.sh and the legacy kube-ovn-leader-checker for that mode, and recovers in-process to avoid kubelet restart cycles.

What's included

- New package pkg/ovn_central_controller/ (controller + kube + ovsdb helpers) and cmd/ovn_central_controller entrypoint wired into cmd/cmdmain.go.
- Lifecycle: preflight → bringUp → runtimeLoop, with recoverCluster() choosing between wipe-rejoin / reconvert / bootstrap based on peer state. Periodic leader labeling, stale-member kicking, raft header backup, and status publishing.
- start-db.sh short-circuit: execs the new binary when DYNAMIC_PEERS=true; static NODE_IPS flow unchanged.
- New readiness probe ovn-central-controller-readiness.sh: passes only when both NB and SB report cluster member with a known leader and committed log range — prevents stub pods from being treated as authoritative during AddServer.
- Helm: new DYNAMIC_PEERS / DYNAMIC_JOIN_TIMEOUT values, RBAC additions, Dockerfile copies the probe script.

### DIFF
--- a/charts/kube-ovn-v2/templates/_helpers.tpl
+++ b/charts/kube-ovn-v2/templates/_helpers.tpl
@@ -103,7 +103,11 @@ Uses Exists operator for empty/nil-value labels and In for specific values.
 Number of master nodes
 */}}
 {{- define "kubeovn.nodeCount" -}}
-  {{- len (split "," ((join "," .Values.masterNodes) | default (include "kubeovn.nodeIPs" .))) }}
+  {{- if .Values.nodeCount -}}
+    {{- .Values.nodeCount }}
+  {{- else -}}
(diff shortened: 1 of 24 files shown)


andriishestakov:
@zbb88888 @oilbeater could you review this change please

PR #6718, opened 2026-05-11
Block vpc deletion while SNAT, DNAT or FIP rules are using it

### LINKED ISSUE #6717: [BUG] VPC deletion with active OvnDnatRule/OvnSnatRule leaves NAT resources permanently stuc
### Kube-OVN Version

1.14.20

### Kubernetes Version

v1.33.7

Microsoft Azure Linux 3.0.20260401

### Description

When a VPC is deleted while OvnDnatRule or OvnSnatRule resources still reference it, those NAT resources become permanently stuck with their finalizers never removed. The webhook that validates VPC deletion (VpcDeleteHook) does not check for live DNAT/SNAT rules referencing the VPC, so the deletion is permitted. Once the VPC's logical router is removed from the OVN Northbound database, the NAT rule delete handlers can no longer clean up their OVN state and loop forever with errors like:

```
error syncing delete ovn snat rule <name>: not found logical router <vpc-name> nat 'type snat ...', requeuing
```

### Steps To Reproduce

1. Create a VPC with an OvnEip, OvnSnatRule and/or OvnDnatRule attached.
2. Delete all subnets in the VPC so the VPC itself becomes deleteable.
3. Delete the VPC without first deleting the OvnDnatRule/OvnSnatRule.
4. Observe that the VPC is deleted successfully.
5. Attempt to delete the remaining OvnDnatRule/OvnSnatRule resources.

### Current Behavior

- The VPC deletion is allowed even though OvnDnatRule/OvnSnatRule resources reference it.
- After the VPC is gone, the controller tries to remove NAT rules from the now-deleted logical router and receives "not found logical router <name>" from the OVN NB client.
- This error is returned from the delete handler, causing the workqueue to requeue the item indefinitely.
- The DNAT/SNAT resources remain with their finalizers attached and can never be deleted.

### Expected Behavior

Deleting a VPC that still has OvnDnatRule or OvnSnatRule resources referencing it should be rejected with a clear error message, consistent with how the webhook already blocks VPC deletion when subnets still exist.
Users should delete all OvnDnatRule and OvnSnatRule resources first, then delete the VPC.

09mulcs:
# Pull Request

## What type of this PR

- Bug fixes

Add webhook checks that prevent vpc deletion if there are DNAT or SNAT rules or FIPs using that VPC.

Fixes #6717

### DIFF
--- a/pkg/webhook/vpc.go
+++ b/pkg/webhook/vpc.go
@@ -62,8 +62,13 @@ func (v *ValidatingHook) VpcDeleteHook(ctx context.Context, req admission.Reques
 	if err := v.cache.List(ctx, dnatList); err != nil {
 		return ctrlwebhook.Errored(http.StatusInternalServerError, err)
 	}
+	// Use Status.Vpc first as vpc may be inferred from spec.ipName if not in spec.
 	for _, item := range dnatList.Items {
-		if item.Spec.Vpc == vpc.Name {
+		referencedVpc := item.Status.Vpc
+		if referencedVpc == "" {
+			referencedVpc = item.Spec.Vpc
+		}
+		if referencedVpc == vpc.Name {
 			return ctrlwebhook.Denied("can't delete vpc when OvnDnatRules still reference it, delete them first")
 		}
 	}
@@ -72,8 +77,13 @@ func (v *ValidatingHook) VpcDeleteHook(ctx context.Context, req admission.Reques
 	if err := v.cache.List(ctx, snatList); err != nil {
 		return ctrlwebhook.Errored(http.StatusInternalServerError, err)
 	}
+	// Use Status.Vpc first as vpc may be inferred from spec.ipName if not in spec.
 	for _, item := range snatList.Items {
-		if item.Spec.Vpc == vpc.Name {
+		referencedVpc := item.Status.Vpc
+		if referencedVpc == "" {
+			referencedVpc = item.Spec.Vpc
+		}
+		if referencedVpc == vpc.Name {
 			return ctrlwebhook.Denied("can't delete vpc when OvnSnatRules still reference it, delete them first")
 		}
 	}
--- a/pkg/webhook/vpc.go
+++ b/pkg/webhook/vpc.go
@@ -49,7 +49,7 @@ func (v *ValidatingHook) VpcUpdateHook(_ context.Context, req admission.Request)
 	return ctrlwebhook.Allowed("bypass")
 }
 
-func (v *ValidatingHook) VpcDeleteHook(_ context.Context, req admission.Request) admission.Response {
+func (v *ValidatingHook) VpcDeleteHook(ctx context.Context, req admission.Request) admission.Response {
 	vpc := ovnv1.Vpc{}
 	if err := v.decoder.DecodeRaw(req.OldObject, &vpc); err != nil {
 		return ctrlwebhook.Errored(http.StatusBadRequest, err)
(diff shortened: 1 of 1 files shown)


pkg/webhook/vpc.go
```diff
@@ -49,13 +49,34 @@ func (v *ValidatingHook) VpcUpdateHook(_ context.Context, req admission.Request)
...
 	if len(vpc.Status.Subnets) != 0 {
 		return ctrlwebhook.Denied("can't delete vpc when any subnet in the vpc")
 	}
+
+	dnatList := &ovnv1.OvnDnatRuleList{}
+	if err := v.cache.List(ctx, dnatList); err != nil {
+		return ctrlwebhook.Errored(http.StatusInternalServerError, err)
+	}
+	for _, item := range dnatList.Items {
+		if item.Spec.Vpc == vpc.Name {
```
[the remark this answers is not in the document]
09mulcs: Suggestion applied

pkg/webhook/vpc.go
```diff
@@ -49,13 +49,44 @@ func (v *ValidatingHook) VpcUpdateHook(_ context.Context, req admission.Request)
...
+		return ctrlwebhook.Errored(http.StatusInternalServerError, err)
+	}
+	// Use Status.Vpc first as vpc may be inferred from spec.ipName if not in spec.
+	for _, item := range dnatList.Items {
+		referencedVpc := item.Status.Vpc
+		if referencedVpc == "" {
+			referencedVpc = item.Spec.Vpc
+		}
+		if referencedVpc == vpc.Name {
+			return ctrlwebhook.Denied("can't delete vpc when OvnDnatRules still reference it, delete them first")
```
oilbeater:
```suggestion
			return ctrlwebhook.Denied(fmt.Sprintf("can't delete vpc %q: still referenced by OvnDnatRule %q", vpc.Name, item.Name))
```

This can give more information to user.

pkg/webhook/vpc.go
```diff
@@ -49,13 +49,44 @@ func (v *ValidatingHook) VpcUpdateHook(_ context.Context, req admission.Request)
...
+		return ctrlwebhook.Errored(http.StatusInternalServerError, err)
+	}
+	// Use Status.Vpc first as vpc may be inferred from spec.ipName if not in spec.
+	for _, item := range snatList.Items {
+		referencedVpc := item.Status.Vpc
+		if referencedVpc == "" {
+			referencedVpc = item.Spec.Vpc
+		}
+		if referencedVpc == vpc.Name {
+			return ctrlwebhook.Denied("can't delete vpc when OvnSnatRules still reference it, delete them first")
```
oilbeater:
```suggestion
			return ctrlwebhook.Denied(fmt.Sprintf("can't delete vpc %q: still referenced by OvnSnatRule %q", vpc.Name, item.Name))
```

pkg/webhook/vpc.go
oilbeater, line 1: Should we also check OvnFip here?

oilbeater (changes requested):

pkg/webhook/vpc.go
09mulcs: Fixed

pkg/webhook/vpc.go
09mulcs: Fixed

pkg/webhook/vpc.go
09mulcs: Good point, we aren't actively using the OvnFip so didn't spot that but added now.

oilbeater (approved):

oilbeater:
Thanks @09mulcs

PR #6736, opened 2026-05-15
feat: add u2o overlay only routing

changluyi:
# Pull Request

## What type of this PR

- Features
- Tests

## What this PR does

This PR adds an optional U2O feature for underlay subnets:

```yaml
spec:
  u2oFeatures:
    overlayOnlyRouting: true
```

When enabled, underlay pods still use the U2O gateway, so traffic enters the OVN logical router. The logical router policies only keep U2O for overlay/join destinations and overlay-to-underlay return traffic. Other underlay-originated traffic is sent to the physical gateway.

## Design

The controller maintains one shared overlay CIDR address set per VPC and IP family:

- `<vpc>.u2o_overlay_cidrs.ip4`
- `<vpc>.u2o_overlay_cidrs.ip6`

The address set contains CIDRs of subnets that:

- belong to the same VPC
- are overlay subnets (`vlan` is empty)

Service CIDRs are not added separately because ClusterIP traffic is DNATed before route policy matching and then follows the backend pod IP destination.

When `overlayOnlyRouting` is enabled on a U2O underlay subnet, this PR now adds these policies per IP family:

| Priority | Match | Action |
| ---: | --- | --- |
| `31000` | `ipX.dst == $<subnet>.u2o_exclude_ip.ipX && ipX.src == <underlay-cidr>` | `reroute <physical-gw>` |
| `30060` | `ipX.src == <underlay-cidr> && ipX.dst == <underlay-cidr>` | `allow` |
| `30050` | `ipX.src == <underlay-cidr>` | `reroute <physical-gw>` |
| `29400` | `ipX.src == $<vpc>.u2o_overlay_cidrs.ipX && ipX.dst == <underlay-cidr>` | `allow` |

It still does not add an explicit underlay-to-overlay allow policy. Underlay-to-overlay traffic is already covered by the existing baseline subnet policy:

| Priority | Match | Action |
| ---: | --- | --- |
| `31000` | `ipX.dst == <overlay-cidr>` | `allow` |
| `31000` | `ipX.dst == <join-cidr>` | `allow` |
| `30000` | `ipX.dst == <node-ip>` | `reroute <join-ip>` |
| `29000` | `ipX.src == $<node-overlay-address-set>` | `reroute <join-ip>` |

The `29400` overlay-to-underlay allow policy is higher than the baseline `29000` source-based node reroute policy, so overlay-to-underlay traffic is not captured by the existing join-IP reroute path. The `31000` excluded-IP reroute keeps underlay-to-node-IP traffic on the physical gateway path even when node/provider IPs belong to the same underlay CIDR. The `30050` physical gateway policy remains lower than the existing overlay/join allow policies, so underlay-to-overlay and underlay-to-join-IP keep the baseline OVN/U2O behavior.

### Patch after local reproduction

During local reproduction on kind, underlay pod creation could fail in IPv6 single-stack and dual-stack because the gateway readiness check kept probing the U2O interconnection IP and never received a reply.

What we observed during troubleshooting:

- neighbor discovery to the U2O IP was working
- the failure was specifically `underlay pod -> U2OInterconnectionIP`
- in overlay-only mode, that traffic could still fall into the broad `30050 ipX.src == <underlay-cidr> reroute <physical-gw>` fallback

Based on that reproduction result, this PR adds the `30060 ipX.src == <underlay-cidr> && ipX.dst == <underlay-cidr> allow` patch so traffic to the U2O interconnection IP no longer falls through to the physical-gateway fallback. The higher-priority `31000` excluded-IP reroute is still preserved so `underlay -> node/provider IP` behavior does not regress.

If `overlayOnlyRouting` is disabled again, stale overlay-only policies are removed and the legacy U2O policies are restored.

## Expected trace behavior

| Source | Destination | Expected policy behavior |
| --- | --- | --- |
| underlay | overlay pod | hit existing `31000 dst overlay-cidr allow` |
| underlay | same underlay subnet pod | same subnet local/L2 path, no overlay-only policy |
| underlay | peer underlay CIDR | hit `30050 src underlay reroute physical-gw` |
| underlay | U2O interconnection IP | hit `30060 src/dst underlay-cidr allow` |
| underlay | ServiceIP -> overlay backend | DNAT to overlay pod IP, then hit `31000 dst overlay-cidr allow` |
| underlay | ServiceIP -> underlay backend | DNAT to underlay pod IP, then hit `30060 src/dst underlay-cidr allow` |
| underlay | node IP | hit `31000 dst excludedIP && src underlay reroute physical-gw`, not baseline `30000 dst nodeIP reroute joinIP` |
| underlay | join IP | keep baseline, hit `31000 dst join-cidr allow` |
| overlay | overlay pod | existing overlay route, no overlay-only policy |
| overlay | underlay pod | hit `29400 overlay -> underlay allow` |
| overlay | ServiceIP -> overlay backend | DNAT to overlay pod IP, no overlay-only policy |
| overlay | ServiceIP -> underlay backend | DNAT to underlay pod IP, then hit `29400 overlay -> underlay allow` |
| overlay | node IP | keep baseline, hit `30000 dst nodeIP reroute joinIP` |
| overlay | join IP | keep baseline, hit `31000 dst join-cidr allow` |

## Tests

- Added validator/unit coverage for `u2oFeatures.overlayOnlyRouting`.
- Added controller unit tests for overlay CIDR address set sync, overlay-only policy generation, and stale policy cleanup when switching between legacy and overlay-only modes.
- Expanded underlay E2E coverage for policy presence/absence after toggling `overlayOnlyRouting` off and on.
- Added reachability checks for underlay-to-overlay, overlay-to-underlay, and same-underlay-subnet ClusterIP traffic while `overlayOnlyRouting` is enabled.
- Expanded underlay E2E `ovn-trace` coverage for the expected matrix above.
- Fixed missing cleanup for the same-subnet overlay pod and overlay backend ClusterIP service in the overlay-only case.
- Changed the relevant `ovn-trace` helpers to fail after a single trace attempt and print the full trace on mismatch.

Local verification run in this iteration:

```text
go test ./pkg/controller -run 'TestAddPolicyRouteForU2OInterconn'
go test ./test/e2e/kube-ovn/underlay -run '^$'
```

`make lint` was not run per maintainer request during this iteration.

Fixes #(issue-number)

### DIFF
--- a/charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
+++ b/charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
@@ -2649,6 +2649,14 @@ spec:
               routeTable:
                 description: Route table associated with the subnet.
                 type: string
+              u2oFeatures:
+                description: Feature configurations for underlay to overlay interconnection.
+                properties:
+                  overlayOnlyRouting:
+                    description: OverlayOnlyRouting controls whether only overlay
+                      CIDRs use U2O routing.
+                    type: boolean
+                type: object
               u2oInterconnection:
                 description: Enable underlay to overlay interconnection.
                 type: boolean
--- a/charts/kube-ovn/templates/kube-ovn-crd.yaml
+++ b/charts/kube-ovn/templates/kube-ovn-crd.yaml
@@ -2649,6 +2649,14 @@ spec:
               routeTable:
                 description: Route table associated with the subnet.
                 type: string
+              u2oFeatures:
+                description: Feature configurations for underlay to overlay interconnection.
+                properties:
+                  overlayOnlyRouting:
+                    description: OverlayOnlyRouting controls whether only overlay
+                      CIDRs use U2O routing.
+                    type: boolean
+                type: object
               u2oInterconnection:
                 description: Enable underlay to overlay interconnection.
                 type: boolean
--- a/dist/images/install.sh
+++ b/dist/images/install.sh
@@ -2900,6 +2900,14 @@ spec:
               routeTable:
                 description: Route table associated with the subnet.
                 type: string
+              u2oFeatures:
+                description: Feature configurations for underlay to overlay interconnection.
+                properties:
+                  overlayOnlyRouting:
+                    description: OverlayOnlyRouting controls whether only overlay
+                      CIDRs use U2O routing.
+                    type: boolean
+                type: object
               u2oInterconnection:
                 description: Enable underlay to overlay interconnection.
                 type: boolean
--- a/pkg/apis/kubeovn/v1/subnet.go
+++ b/pkg/apis/kubeovn/v1/subnet.go
@@ -136,6 +136,8 @@ type SubnetSpec struct {
 	U2OInterconnectionIP string `json:"u2oInterconnectionIP,omitempty"`
 	// Enable underlay to overlay interconnection.
 	U2OInterconnection bool `json:"u2oInterconnection,omitempty"`
+	// Feature configurations for underlay to overlay interconnection.
+	U2OFeatures U2OFeatures `json:"u2oFeatures,omitempty"`
 	// Enable LoadBalancer on this subnet.
 	EnableLb *bool `json:"enableLb,omitempty"`
 	// Enable ECMP for centralized gateway.
@@ -151,6 +153,11 @@ type SubnetSpec struct {
 	NamespaceSelectors []metav1.LabelSelector `json:"namespaceSelectors,omitempty"`
 	// Node network name for underlay.
 	NodeNetwork string `json:"nodeNetwork,omitempty"`
+}
+
+type U2OFeatures struct {
+	// OverlayOnlyRouting controls whether only overlay CIDRs use U2O routing.
+	OverlayOnlyRouting bool `json:"overlayOnlyRouting,omitempty"`
 }
 
 type ACL struct {
--- a/pkg/controller/subnet.go
+++ b/pkg/controller/subnet.go
@@ -74,6 +74,10 @@ func readyToRemoveFinalizer(subnet *kubeovnv1.Subnet) bool {
 	return false
 }
 
+func u2oOverlayOnlyRoutingEnabled(subnet *kubeovnv1.Subnet) bool {
+	return subnet != nil && subnet.Spec.U2OFeatures.OverlayOnlyRouting
+}
+
 func (c *Controller) enqueueUpdateSubnet(oldObj, newObj any) {
 	oldSubnet := oldObj.(*kubeovnv1.Subnet)
 	newSubnet := newObj.(*kubeovnv1.Subnet)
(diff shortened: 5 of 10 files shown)


patriziobassi:
hi @changluyi will the subnet ACL still apply or won't they anymore (i guess the second option)?
and what about the LoadBalancer services?

thank you

changluyi:
> hi @changluyi will the subnet ACL still apply or won't they anymore (i guess the second option)? and what about the LoadBalancer services?
> 
> thank you

Hi @patriziobassi, subnet ACL is no longer needed in this mode — the LR policies handle the routing logic directly. LB services behavior is unchanged. Thanks!

patriziobassi:
> > hi @changluyi will the subnet ACL still apply or won't they anymore (i guess the second option)? and what about the LoadBalancer services?
> > thank you
> 
> Hi @patriziobassi, subnet ACL is no longer needed in this mode — the LR policies handle the routing logic directly. LB services behavior is unchanged. Thanks!

we are using ACLs to logically split a Layer 2 subnet and assign a ippool to a namespace.
for instance the kubeovn subnet is:
10.1.0.0/24

so we split in 10.1.0.32/27, with the corresponding ippool assigned to namespace A, 10.1.0.64/27 assigned to namespace B and so on. As it's Layer2 pods would be able to communicate, but using ACLs we can completely deny, pretending they are separate networks.
I guess this won't be possible anymore in this mode, but we need to use K8s network policies or similar approach.

changluyi:
> > > hi @changluyi will the subnet ACL still apply or won't they anymore (i guess the second option)? and what about the LoadBalancer services?
> > > thank you
> > 
> > 
> > Hi @patriziobassi, subnet ACL is no longer needed in this mode — the LR policies handle the routing logic directly. LB services behavior is unchanged. Thanks!
> 
> we are using ACLs to logically split a Layer 2 subnet and assign a ippool to a namespace. for instance the kubeovn subnet is: 10.1.0.0/24
> 
> so we split in 10.1.0.32/27, with the corresponding ippool assigned to namespace A, 10.1.0.64/27 assigned to namespace B and so on. As it's Layer2 pods would be able to communicate, but using ACLs we can completely deny, pretending they are separate networks. I guess this won't be possible anymore in this mode, but we need to use K8s network policies or similar approach.

  For underlay-to-underlay traffic across different underlay subnets, yes, this mode intentionally does not rely on the previous logical ACL/routing behavior anymore. The traffic
  is rerouted to the physical gateway, so isolation/routing between those underlay subnets should be handled by the physical network.

  For the case you described, if the IPPools are carved from the same Kube-OVN subnet, that is same-L2 east-west traffic. overlayOnlyRouting should not bypass logical switch ACLs
  or NetworkPolicy ACLs for same-subnet traffic. Those ACLs should still be the right place to enforce namespace/IPPool isolation inside one L2 subnet.

oilbeater (approved):

Issue #5073, opened 2025-03-11
[Feature Request] Allow cert-manager to sign IPSec CSRs


cruickshankpg:
### Description

Currently when IPSec is enabled kube-ovn-controller is responsible for:
1. enabling IPSec in OVNNB
2. generating the ovn-ipsec-ca secret if it doesn't exist
3. signing CSRs from kube-ovn-cni pods

In our k8s deployments we use cert-manager to sign all our other CSRs and ideally we would use it for IPSec as well.

I think I can achieve this currently by removing permissions for `certificates.k8s.io` apiGroups in the `system:ovn` ClusterRole. kube-ovn-controller appears healthy but it does then periodically log errors:

```
0311 15:56:39.748324       7 reflector.go:561] k8s.io/client-go@v1.5.2/tools/cache/reflector.go:243: failed to list *v1.CertificateSigningRequest: certificatesigningrequests.certificates.k8s.io is forbidden: User "system:serviceaccount:kube-system:ovn" cannot list resource "certificatesigningrequests" in API group "certificates.k8s.io" at the cluster scope
E0311 15:56:39.748424       7 reflector.go:158] "Unhandled Error" err="k8s.io/client-go@v1.5.2/tools/cache/reflector.go:243: Failed to watch *v1.CertificateSigningRequest: failed to list *v1.CertificateSigningRequest: certificatesigningrequests.certificates.k8s.io is forbidden: User \"system:serviceaccount:kube-system:ovn\" cannot list resource \"certificatesigningrequests\" in API group \"certificates.k8s.io\" at the cluster scope" logger="UnhandledError"
```

It would be nice if there was a config flag to disable the csrInformer and not generate the ovn-ipsec-ca secret but still enable ipsec in ovnnb. And then a helm value to set the config flag and toggle the ClusterRole permissions.

### Who will benefit from this feature?

IPSec users that use cert-manager

oilbeater:
@changluyi can you take a look at this?

cruickshankpg:
Looks like the kube-ovn-cni pods should also create cert-manager [CertificateRequests](https://cert-manager.io/docs/usage/certificaterequest/) rather than k8s CertificateSigningRequests. cert-manager does have [experimental support](https://cert-manager.io/docs/usage/certificaterequest/) for signing CSRs.

cruickshankpg:
hey @changluyi have you made a start on this? I'm happy to try and implement this

changluyi:
[...]

PR #6743, opened 2026-05-16
fix: avoid panic decoding invalid CA private keys

immanuwell:
# Pull Request

## What type of this PR

- Bug fixes
- Tests

## What this does

Tiny footgun fix tbh: if the ovn-ipsec-ca Secret has a bad cakey value, decodePrivateKey tries to print block.Type even when pem.Decode returned nil. That panics the controller instead of going down the existing CorruptCAKey error path.

This drops the stray stdout print and adds tests for valid PKCS#8 RSA keys + invalid PEM.

## How to reproduce

On master, add/run the invalid PEM regression test:

```bash
go test ./pkg/controller -run TestDecodePrivateKeyReturnsErrorForInvalidPEM -count=1
```

It crashes with a nil pointer panic. With this patch, it returns an error, no drama.

## Related

Related context: #5073 and #5972, both around IPSec CSR/signing behavior.

## Tests

```bash
go test ./pkg/controller -run TestDecodePrivateKey -count=1 -v
go test ./pkg/controller -count=1
make lint
```

### DIFF
--- a/pkg/controller/signer.go
+++ b/pkg/controller/signer.go
@@ -297,7 +297,6 @@ func decodeCertificate(pemBytes []byte) (*x509.Certificate, error) {
 func decodePrivateKey(pemBytes []byte) (*rsa.PrivateKey, error) {
 	block, _ := pem.Decode(pemBytes)
 	if block == nil || block.Type != "PRIVATE KEY" {
-		fmt.Println(block.Type)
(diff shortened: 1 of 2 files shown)


oilbeater:
@immanuwell Thanks and please signoff the commit to pass the DCO check.

PR #6804, opened 2026-06-02
fix(controller): correct egress sync status field on ACL update failure

### LINKED ISSUE #6803: [BUG]: security group egress sync status incorrectly reports ingress failure
## Bug Description

In `pkg/controller/security_group.go` (line 241), when the egress ACL update fails, `IngressLastSyncSuccess` is incorrectly set to `false` instead of `EgressLastSyncSuccess`. This is a clear copy-paste bug from the ingress block above.

## Impact

- **False-positive alerts on ingress sync**: Operators may think ingress rules are broken when they are actually fine.
- **False-negative alerts on egress sync**: Real egress sync failures are not reflected in the status, so they can be missed.
- **Increased MTTR**: Operators look at the wrong status field during debugging and health checks.

## Affected Code

```go
if egressNeedUpdate {
    if err = c.OVNNbClient.UpdateSgACL(sg, ovnnb.ACLDirectionFromLport); err != nil {
        sg.Status.IngressLastSyncSuccess = false   // BUG: should be EgressLastSyncSuccess
        ...
    }
    sg.Status.EgressLastSyncSuccess = true         // correct
}
```

## Fix

Change line 241 to `sg.Status.EgressLastSyncSuccess = false`.

harisudarsan1:
Fixes #6803

When the egress ACL update fails in `handleAddOrUpdateSg`, `IngressLastSyncSuccess` was incorrectly set to `false` instead of `EgressLastSyncSuccess`. This is a copy-paste bug from the ingress block above.

The incorrect status field causes:
- Ingress status to become falsely negative — monitoring/alerting thinks ingress rules are broken when they are fine.
- Egress status to become falsely positive — real egress sync failures are missed.

This change corrects line 241 to set `EgressLastSyncSuccess = false` in the egress error path.

### DIFF
--- a/pkg/controller/security_group.go
+++ b/pkg/controller/security_group.go
@@ -238,7 +238,7 @@ func (c *Controller) handleAddOrUpdateSg(key string, force bool) error {
 	}
 	if egressNeedUpdate {
 		if err = c.OVNNbClient.UpdateSgACL(sg, ovnnb.ACLDirectionFromLport); err != nil {
-			sg.Status.IngressLastSyncSuccess = false
+			sg.Status.EgressLastSyncSuccess = false
 			c.patchSgStatus(sg)
 			klog.Error(err)
 			return err


oilbeater (approved):

oilbeater:
Thanks! @harisudarsan1

PR #6827, opened 2026-06-05
perf(daemon): list subnets and pods once per gateway setter

oilbeater:
## Summary
- `runGateway` runs every 3s; each of `setIPSet` / `setPolicyRouting` / `setIptables` independently re-listed the full subnet informer cache — up to **15 full subnet lists per dual-stack tick** (and pods 2–4×).
- List subnets (and pods) **once per setter** and thread the slice through the helpers, cutting it to **3 subnet lists/tick** and **1–2 pod lists/tick**.
- Scoped per-setter (not one global snapshot) to keep each setter's existing independent-snapshot boundary — minimal behavioral change.
- Dropped the now-redundant `error` returns from helpers whose only error source was the removed `List` call; callers simplified accordingly. `StartTProxyTCPPortProbe` (separate path) now lists pods itself.

## Test plan
- [x] `go build ./pkg/daemon/...` + `go vet`
- [x] `make lint` — 0 issues (golangci-lint + modernize)
- [x] `make ut` — Ginkgo suite + all `pkg` `go test` pass
- [ ] No functional change expected; gateway iptables/ipset/policy-routing reconcile behavior is identical (same data, fewer cache reads)

### DIFF
--- a/pkg/daemon/gateway.go
+++ b/pkg/daemon/gateway.go
@@ -94,11 +94,16 @@ func (c *Controller) isSubnetNeedNat(subnet *kubeovnv1.Subnet, protocol string) 
 func (c *Controller) getSubnetsNeedNAT(subnets []*kubeovnv1.Subnet, protocol string) []string {
 	var subnetsNeedNat []string
 	for _, subnet := range subnets {
-		if c.isSubnetNeedNat(subnet, protocol) {
-			cidrBlock, err := getCidrByProtocol(subnet.Spec.CIDRBlock, protocol)
-			if err == nil && cidrBlock != "" {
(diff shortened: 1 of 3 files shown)


oilbeater:
Thanks @Copilot — addressed all 4 comments in aadf40701.

`getSubnetsNeedNAT`, `getSubnetsDistributedGateway`, `getDefaultVpcSubnetsCIDR` and `StartTProxyTCPPortProbe` now log the `getCidrByProtocol` / `getTProxyConditionPod` error instead of silently skipping, matching the existing handling in `reconcileNatOutGoingPolicyIPset` and `generateNatOutgoingPolicyChainRules`. The error branch only fires on malformed CIDRs (not on normal single-stack subnets), so there's no per-tick log spam.

PR #4895, opened 2025-01-02
[performance] in large-scale clusters, init node router policy too slow

cmdy:
# Pull Request

- [ ] Make sure you have followed [Kube-OVN Code Style](https://github.com/kubeovn/kube-ovn/blob/master/CODE_STYLE.md).

## What type of this PR

Examples of user facing changes:

- Features
- Bug fixes
- Docs
- Tests

## Which issue(s) this PR fixes

Fixes #4849

PR #4884 merge into master branch

PR #6829, opened 2026-06-05
perf(ovs): list NAT and static routes via single WhereCache scan

oilbeater:
## Summary
- Rewrite `listLogicalRouterNatByFilter` and `listLogicalRouterStaticRoutesByFilter` to do a single `WhereCache().List()` filtered by a UUID set, instead of looping `GetXByUUID` per UUID in `lr.Nat` / `lr.StaticRoutes`.
- This mirrors the already-merged pattern in `listLogicalRouterPoliciesByFilter` (#4895) and the in-tree `batchListLogicalRouterStaticRoutesForDelete`.
- Drops the per-row `context.WithTimeout` allocation and the second deep clone done for every row; only matched rows are cloned. Also removes the now-unused `GetLogicalRouterStaticRouteByUUID` helper.

## Notes
- Behavior is equivalent except result ordering, which becomes non-deterministic (map iteration). Audited all callers of `ListNats` / `ListLogicalRouterStaticRoutes` / `ListLogicalRouterStaticRoutesByOption`: every consumer iterates the full list by key or only checks `len()==0` — none index `[0]` or rely on order. The single-result getters (`GetNat`, `GetLogicalRouterStaticRoute`) filter by a unique key and error on `>1`, so they are unaffected.
- `GetNATByUUID` is kept (still used by `pkg/controller/vpc.go` and the `OVNNbClient` interface).

## Test plan
- [x] `go build ./pkg/...`
- [x] `go vet ./pkg/ovs`
- [x] `make lint` (0 issues)
- [x] Targeted unit tests: `Test_GetNat`, `Test_ListNats`, `Test_DeleteNats`, `Test_AddNat`, `Test_GetNATByUUID`, `Test_*LogicalRouterStaticRoute*` (incl. ecmp, edge cases, batch delete) — all pass

### DIFF
--- a/pkg/ovs/ovn-nb-logical_router_route.go
+++ b/pkg/ovs/ovn-nb-logical_router_route.go
@@ -455,6 +455,10 @@ func (c *OVNNbClient) listLogicalRouterStaticRoutesByFilter(lrName string, filte
 		return nil, err
 	}
 
+	if len(lr.StaticRoutes) == 0 {
+		return nil, nil
+	}
+
 	uuidSet := set.New(lr.StaticRoutes...)
 	predicate := func(route *ovnnb.LogicalRouterStaticRoute) bool {
 		if !uuidSet.Has(route.UUID) {
--- a/pkg/ovs/ovn-nb-logical_router_route.go
+++ b/pkg/ovs/ovn-nb-logical_router_route.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"slices"
 
-	"github.com/ovn-kubernetes/libovsdb/client"
 	"github.com/ovn-kubernetes/libovsdb/model"
 	"github.com/ovn-kubernetes/libovsdb/ovsdb"
 	"github.com/scylladb/go-set/strset"
@@ -323,20 +322,6 @@ func (c *OVNNbClient) ClearLogicalRouterStaticRoute(lrName string) error {
 	return nil
 }
 
-// GetLogicalRouterStaticRouteByUUID get logical router static route by UUID
-func (c *OVNNbClient) GetLogicalRouterStaticRouteByUUID(uuid string) (*ovnnb.LogicalRouterStaticRoute, error) {
-	ctx, cancel := context.WithTimeout(context.Background(), c.Timeout)
-	defer cancel()
-
-	route := &ovnnb.LogicalRouterStaticRoute{UUID: uuid}
-	if err := c.Get(ctx, route); err != nil {
-		klog.Error(err)
-		return nil, err
-	}
-
-	return route, nil
-}
-
 // GetLogicalRouterStaticRoute get logical router static route by some attribute,
 // a static route is uniquely identified by router(lrName), policy and ipPrefix when route is not ecmp
 // a static route is uniquely identified by router(lrName), policy, ipPrefix and nexthop when route is ecmp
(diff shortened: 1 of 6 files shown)


oilbeater:
### Benchmark: per-UUID `Get` loop vs single `WhereCache().List()`

Measured both code paths against an **identical in-memory OVN NB cache** (libovsdb test server), listing all `snat` rules on a logical router. `old` = the previous `for uuid := range lr.Nat { GetNATByUUID(uuid) }` loop; `new` = this PR's `WhereCache(predicate).List()`. Same dataset feeds both; `go test -bench -benchmem`.

| Scenario | Path | ns/op | B/op | allocs/op |
|---|---|---:|---:|---:|
| **N≈T** — 100 NAT on one LR | old `Get` loop | 8,851,446 | 383 KB | 6,747 |
| | **new `WhereCache`** | **632,476** | **123 KB** | **1,476** |
| | | **~14× faster** | 3.1× less | 4.6× fewer |
| **N≈T** — 1000 NAT on one LR | old `Get` loop | 87,242,330 | 3.83 MB | 67,060 |
| | **new `WhereCache`** | **6,200,770** | **1.33 MB** | **14,105** |
| | | **~14× faster** | 2.9× less | 4.8× fewer |
| **N≪T** — 10 on target LR + 2000 on another | old `Get` loop | 944,665 | **40 KB** | **714** |
| | new `WhereCache` | 799,022 | 229 KB | 4,209 |
| | | ~1.2× faster | 5.6× more mem | 5.9× more |

**Takeaways**

- **The cost was per-row constant, not complexity.** The old path scales at a fixed ~67 allocs and ~87 µs *per NAT rule* (6,747→67,060 allocs as rows go 100→1000). Each row paid for a `context.WithTimeout` allocation plus two deep reflective clones (`RowByModel` clone + `CloneInto`). The new path runs one cheap predicate per row and deep-clones only the matched rows.
- **Hot path (N≈T) is a clean win:** ~14× faster and ~3× less memory at both sizes — e.g. listing 1000 NAT rules drops 87.2 ms → 6.2 ms. This is exactly the per-call cost inside the EIP-SNAT / route-reconcile loops on the cluster router, where almost all rows live on a single router.
- **Boundary case (N≪T) is honest about its trade:** a router holding a tiny slice of a large table makes the full-table scan allocate ~5.6× more memory; wall-time stayed comparable here at T≈2000 but could regress at much larger T. The converted helpers' hot callers operate in the N≈T regime, so this is acceptable.

Static routes use the same code shape (`listLogicalRouterStaticRoutesByFilter`), so the same characteristics apply.

oilbeater:
Thanks for the review! Addressed the feedback:

**Copilot — full cache scan when the router has zero NAT / static routes** (both inline comments): valid catch — the old per-UUID loop returned immediately for an empty `lr.Nat` / `lr.StaticRoutes`, while the new code would still scan the whole table. Restored the early return in both helpers (`ecb55f89a`):

```go
if len(lr.Nat) == 0 {
    return nil, nil
}
```

Added unit coverage for the empty-router path in both `Test_GetNat` and `Test_ListLogicalRouterStaticRoutes`.

**Benchmark** (`4acffc1b2`): added the benchmark behind the numbers in the comment above, so the comparison is reproducible (`go test ./pkg/ovs -run=^$ -bench=BenchmarkListNAT -benchmem`). Also widened two test helpers to `testing.TB` so benchmarks can reuse them.

**`OVN METALLB E2E (ipv6)` failure**: unrelated known flake — `[It] should support metallb and underlay combine` timed out on `service fc00:…::64:80 should be reachable` (curl exit 7, IPv6 NDP/underlay dataplane convergence). This PR only touches NB-client NAT/static-route cache reads, which cannot affect metallb underlay dataplane. The new push re-runs CI.

Issue #6823, opened 2026-06-04
[BUG] KubeVirt live migration deadlock: migration setup skipped when target pod not yet scheduled, never retried (primary CNI)


jiri-jezek1:
### Kube-OVN Version

v1.15.10
In this case the whole infrastructure is cozystack version 1.4.2
As CNI its used kube-ovn with cilium in chained regime - https://kubeovn.github.io/docs/v1.12.x/en/advance/with-cilium/

### Kubernetes Version

v1.34.6

### Operation-system/Kernel Version

"Ubuntu 24.04.4 LTS"
6.8.0-100-generic

### Description

When Kube-OVN is used as the **primary CNI** and `--enable-live-migration-optimize=true`
(the default), every KubeVirt live migration fails after ~15 minutes. The migration
target pod is stuck in `ContainerCreating` throughout. `kube-ovn-cni` logs the same
error every ~40 seconds:

```
E ovs_linux.go:232] ovs interface <hash>_h is not ready after 30s
E handler.go:334]   configure nic eth0 for pod <vm>/<namespace> failed: ovs interface <hash>_h is not ready after 30s
```

The OVN NB logical switch port for the VM keeps `options: {requested-chassis=<source-node>}`
throughout the entire attempt — it is never updated to the target node.

### Steps To Reproduce

1. Deploy Kube-OVN as primary CNI with `--enable-live-migration-optimize=true`
2. Run a KubeVirt VM on any node
3. Create a `VirtualMachineInstanceMigration` to migrate the VM to a different node
4. Observe: target pod stays in `ContainerCreating` indefinitely; VMIM stays in `Scheduling`
5. After ~15 minutes, VMIM transitions to `Failed`

The failure is 100% reproducible across all VMs, all source nodes, and both migration
directions. 

### Current Behavior

`handleAddOrUpdateVMIMigration` processes the `MigrationScheduling` phase with [this guard](https://github.com/kubeovn/kube-ovn/blob/release-1.15/pkg/controller/kubevirt.go#L186-L190) 

```go
if sourceNode == "" || targetPod.Spec.NodeName == "" || sourceNode == targetPod.Spec.NodeName {
    klog.Warningf("VM pod %s/%s migration setup skipped, source node: %s, target node: %s ...")
    return nil   // ← dead end: no retry, no re-enqueue
}
``` 

When a VMIM first enters `Scheduling`, the target pod has been created but
[...]

PR #6834, opened 2026-06-05
fix(controller): re-enqueue VMI migration when target pod is not yet …

jiri-jezek1:
…scheduled

# Pull Request

## What type of this PR

- Bug fixes

  `handleAddOrUpdateVMIMigration` sets `requested-chassis` on the VM's OVN logical
  switch port during the `MigrationScheduling` phase. It finds the target pod via
  the `kubevirt.io/migrationJobUID` label and, if the target pod isn't scheduled
  yet (`targetPod.Spec.NodeName == ""`) or doesn't exist yet, logs a warning and
  `return nil` — dropping the key from the work queue.

  While the VMIM stays in `Scheduling` no further phase-change event is emitted,
  so migration setup is never retried. The OVN port stays bound to the source
  chassis, the target pod hangs in `ContainerCreating` (`ovs interface ... is not
  ready after 30s`), and the migration fails after ~15 min. With Kube-OVN as the
  primary CNI this is a reproducible deadlock whenever the `Scheduling` event is
  processed before the target pod is scheduled.
  
  This re-enqueues the VMI migration via `AddAfter(key, 2*time.Second)` in those
  transient cases, so setup runs once scheduling completes. The success path and
  the genuine same-node skip are unchanged; real OVN errors still return an error
  and retry with the existing rate-limited backoff.

  ## Testing

  Verified on a live cluster (Kube-OVN v1.15.10, primary CNI, KubeVirt). Before:
  live migrations deadlocked 100% of the time. After: migration completes in ~20s
  with no manual intervention. Controller logs show the guard being hit
  (`migration setup skipped ... target node: ""`) followed ~2s later by the
  re-enqueue completing setup (`successfully set migrate options ...`) and the
  VMIM reaching `Succeeded`.

Fixes #(issue-number)  #6823

### DIFF
--- a/pkg/controller/kubevirt.go
+++ b/pkg/controller/kubevirt.go
@@ -186,6 +186,15 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
 			if sourceNode == "" || targetPod.Spec.NodeName == "" || sourceNode == targetPod.Spec.NodeName {
 				klog.Warningf("VM pod %s/%s migration setup skipped, source node: %s, target node: %s (migration job UID: %s)",
 					targetPod.Namespace, targetPod.Name, sourceNode, targetPod.Spec.NodeName, vmiMigration.UID)
+                // The source or target node may not be known yet while the migration is still
+                // in the Scheduling phase; this produces no further phase-change event, so
+                // return an error to requeue with rate limiting and retry once scheduling
+                // completes, instead of dropping the event.
+                // https://github.com/kubeovn/kube-ovn/issues/6823
+                if sourceNode == "" || targetPod.Spec.NodeName == "" {
+                        return fmt.Errorf("VM pod %s/%s migration setup deferred, source node %q, target node %q not ready yet (migration job UID %s)",
+                                targetPod.Namespace, targetPod.Name, sourceNode, targetPod.Spec.NodeName, vmiMigration.UID)
+                }
 				return nil
 			}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/kubevirt.go
```diff
@@ -186,6 +186,13 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
 			if sourceNode == "" || targetPod.Spec.NodeName == "" || sourceNode == targetPod.Spec.NodeName {
 				klog.Warningf("VM pod %s/%s migration setup skipped, source node: %s, target node: %s (migration job UID: %s)",
 					targetPod.Namespace, targetPod.Name, sourceNode, targetPod.Spec.NodeName, vmiMigration.UID)
+				// The source or target node may not be known yet while the migration is still
```
oilbeater, line 189: What about just return error here, it will automatically requeue with ratelimit.

oilbeater (changes requested):

pkg/controller/kubevirt.go
```diff
@@ -203,6 +210,9 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
 		} else {
 			klog.Warningf("target pod not yet created for migration job UID %s in phase %s, waiting for pod creation",
 				vmiMigration.UID, vmiMigration.Status.Phase)
+			// The target pod has not been created yet; re-enqueue so setup is retried once it
+			// appears. https://github.com/kubeovn/kube-ovn/issues/6823
+			c.addOrUpdateVMIMigrationQueue.AddAfter(key, 2*time.Second)
```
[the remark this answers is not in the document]
jiri-jezek1: ok i kept is as it was

pkg/controller/kubevirt.go
jiri-jezek1: Ok make more sence, thanks for review. Fixed it now

oilbeater (approved):

oilbeater:
Thanks! @jiri-jezek1

PR #6882, opened 2026-06-17
fix(controller): enable ARP learning on VPC routers that have peerings

LoneExile:
## Description

VPC peering requires ARP learning on the logical router.

When traffic arrives from a peered VPC, it enters via the peer LRP. If `always_learn_from_arp_request=false`, the destination pod's MAC is never learned into `mac_binding`, so the router mis-forwards the packet (tunnels it instead of delivering locally).

Currently `learnFromARPRequest` is only set for `EnableExternal` and the U2O interconnection case. This change also enables it when the VPC has any `VpcPeerings`.

## Impact

- Fixes cross-VPC L2 delivery for peered VPCs.
- No behavior change for VPCs without peerings.
- Minimal and safe.


oilbeater (approved):

oilbeater:
Thanks! @LoneExile

PR #6885, opened 2026-06-17
feat: Add kubeovn support for dynamic ip allocation/byo dhcp

zhao-xuan:
# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)

### DIFF
--- a/pkg/controller/init.go
+++ b/pkg/controller/init.go
@@ -773,10 +773,12 @@ func (c *Controller) syncSubnetCR() error {
 			klog.Warningf("subnet %s is not ready", subnet.Name)
 			continue
 		}
-		subnet, err = c.calcSubnetStatusIP(subnet)
-		if err != nil {
-			klog.Errorf("failed to calculate subnet %s used ip: %v", cachedSubnet.Name, err)
-			return err
+		if subnet.Spec.CIDRBlock != "" {
+			subnet, err = c.calcSubnetStatusIP(subnet)
+			if err != nil {
+				klog.Errorf("failed to calculate subnet %s used ip: %v", cachedSubnet.Name, err)
+				return err
+			}
 		}
 
 		// only sync subnet spec enableEcmp when subnet.Spec.EnableEcmp is false and c.config.EnableEcmp is true
--- a/pkg/controller/pod.go
+++ b/pkg/controller/pod.go
@@ -2186,6 +2186,24 @@ func (c *Controller) acquireAddress(pod *v1.Pod, podNet *kubeovnNet) (string, st
 	key := cache.NewObjectName(pod.Namespace, podName).String()
 	portName := ovs.PodNameToPortName(podName, pod.Namespace, podNet.ProviderName)
 
+	// Handle underlay subnets without CIDR (BYO-DHCP / external DHCP) - only allocate a MAC
+	if podNet.Subnet.Spec.Vlan != "" && podNet.Subnet.Spec.CIDRBlock == "" {
+		klog.Infof("allocating MAC-only for pod %s in underlay subnet %s without CIDR", key, podNet.Subnet.Name)
+
+		annoMAC := pod.Annotations[fmt.Sprintf(util.MacAddressAnnotationTemplate, podNet.ProviderName)]
+		if annoMAC != "" {
+			if _, err := net.ParseMAC(annoMAC); err != nil {
+				return "", "", "", podNet.Subnet, err
+			}
+			// Use the existing MAC annotation
+			return "", "", annoMAC, podNet.Subnet, nil
+		}
+
+		// Generate a new MAC address
+		mac := util.GenerateMac()
+		return "", "", mac, podNet.Subnet, nil
+	}
+
 	var checkVMPod bool
 	isStsPod, _, _ := isStatefulSetPod(pod)
 	// if pod has static vip
--- a/pkg/controller/pod.go
+++ b/pkg/controller/pod.go
@@ -2180,6 +2180,41 @@ func (c *Controller) validatePodIP(podName, subnetName, ipv4, ipv6 string) (bool
 	return true, true, nil
 }
 
+// acquireMacOnlyAddress allocates only a MAC address (no IP) for a pod NIC on an
+// underlay subnet without a CIDR (BYO-DHCP / external DHCP). The guest obtains its
+// IP from an external DHCP server. MAC allocation goes through IPAM so it is
+// idempotent per NIC (no churn on retries) and the subnet/MAC are tracked, which
+// prevents the GC from cleaning the pod's mac-only IP/LSP resources.
+func (c *Controller) acquireMacOnlyAddress(pod *v1.Pod, podNet *kubeovnNet, key, portName string) (string, string, string, *kubeovnv1.Subnet, error) {
+	klog.Infof("allocating MAC-only address for pod %s in mac-only subnet %s", key, podNet.Subnet.Name)
+
+	// Prefer a per-interface MAC annotation, then fall back to the provider-level one.
+	var macPointer *string
+	if podNet.NadName != "" && podNet.NadNamespace != "" && podNet.InterfaceName != "" {
+		if macStr := pod.Annotations[perInterfaceMACAnnotationKey(podNet.NadName, podNet.NadNamespace, podNet.InterfaceName)]; macStr != "" {
+			if _, err := net.ParseMAC(macStr); err != nil {
+				return "", "", "", podNet.Subnet, err
+			}
+			macPointer = &macStr
+		}
+	}
+	if macPointer == nil {
+		if annoMAC := pod.Annotations[fmt.Sprintf(util.MacAddressAnnotationTemplate, podNet.ProviderName)]; annoMAC != "" {
+			if _, err := net.ParseMAC(annoMAC); err != nil {
+				return "", "", "", podNet.Subnet, err
+			}
+			macPointer = &annoMAC
+		}
+	}
+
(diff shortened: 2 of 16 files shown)


pkg/controller/subnet.go
```diff
@@ -212,17 +212,22 @@ func (c *Controller) validateSubnetVlan(subnet *kubeovnv1.Subnet) error {
...
+		if err := formatGateway(subnet); err != nil {
+			klog.Error(err)
+			return err
+		}
 
-	formatExcludeIPs(subnet)
+		formatExcludeIPs(subnet)
+	}
 
 	subnet.Spec.Protocol = util.CheckProtocol(subnet.Spec.CIDRBlock)
```
oilbeater: I think we need a new protocol for mac-only subnet.

pkg/controller/pod.go
```diff
@@ -2186,6 +2186,24 @@ func (c *Controller) acquireAddress(pod *v1.Pod, podNet *kubeovnNet) (string, st
 	key := cache.NewObjectName(pod.Namespace, podName).String()
 	portName := ovs.PodNameToPortName(podName, pod.Namespace, podNet.ProviderName)
 
+	// Handle underlay subnets without CIDR (BYO-DHCP / external DHCP) - only allocate a MAC
+	if podNet.Subnet.Spec.Vlan != "" && podNet.Subnet.Spec.CIDRBlock == "" {
+		klog.Infof("allocating MAC-only for pod %s in underlay subnet %s without CIDR", key, podNet.Subnet.Name)
+
+		annoMAC := pod.Annotations[fmt.Sprintf(util.MacAddressAnnotationTemplate, podNet.ProviderName)]
```
oilbeater: We should consider per-interface mac first here.

pkg/controller/subnet.go
```diff
@@ -568,16 +573,20 @@ func (c *Controller) handleAddOrUpdateSubnet(key string) error {
...
-		klog.Errorf("calculate subnet %s used ip failed, %v", cachedSubnet.Name, err)
-		return err
+		// availableIPStr valued from ipam, so leave update subnet.status after ipam process
+		subnet, err = c.calcSubnetStatusIP(subnet)
+		if err != nil {
+			klog.Errorf("calculate subnet %s used ip failed, %v", cachedSubnet.Name, err)
+			return err
+		}
+	} else {
+		klog.Infof("subnet %s has no cidrBlock, skipping IPAM (BYO-DHCP / external DHCP)", subnet.Name)
```
oilbeater: We should at least add a subnet entry in ipam. Otherwise gc https://github.com/kubeovn/kube-ovn/blob/master/pkg/controller/gc.go#L311 will clean all mac-only resources.

oilbeater (changes requested):

oilbeater:
I think we also need an e2e to make sure the mac-only interface creation is success.

pkg/webhook/subnet.go
```diff
@@ -69,6 +69,16 @@ func (v *ValidatingHook) SubnetUpdateHook(ctx context.Context, req admission.Req
...
 
+	// Prevent converting an underlay subnet without CIDR (BYO-DHCP) to one with CIDR and vice versa
+	oldIsUnderlayWithoutCIDR := oldSubnet.Spec.Vlan != "" && oldSubnet.Spec.CIDRBlock == ""
+	newIsUnderlayWithoutCIDR := o.Spec.Vlan != "" && o.Spec.CIDRBlock == ""
+	if oldIsUnderlayWithoutCIDR && !newIsUnderlayWithoutCIDR {
+		return ctrlwebhook.Denied("cannot add cidrBlock to an underlay subnet that was created without one")
+	}
+	if !oldIsUnderlayWithoutCIDR && newIsUnderlayWithoutCIDR {
+		return ctrlwebhook.Denied("cannot remove cidrBlock from an underlay subnet that was created with one")
+	}
```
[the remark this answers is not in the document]
zhao-xuan: Addressed in 9d162ee. The update guard now keys off the actual `cidrBlock` transition rather than the underlay-without-CIDR flag: it denies only when an existing MAC-only subnet *gains* a cidrBlock (`oldIsUnderlayWithoutCIDR && o.Spec.CIDRBlock != ""`) or an existing CIDR subnet *loses* it (`oldSubnet.Spec.CIDRBlock != "" && newIsUnderlayWithoutCIDR`). Clearing the vlan while cidrBlock stays empty no longer trips the misleading "cannot add cidrBlock…" denial, and each message now matches the actual mutation.

pkg/util/validator.go
```diff
@@ -15,29 +15,42 @@ import (
...
-	if err := CIDRGlobalUnicast(subnet.Spec.CIDRBlock); err != nil {
-		klog.Error(err)
-		return err
-	}
-	if CheckProtocol(subnet.Spec.CIDRBlock) == "" {
-		return fmt.Errorf("CIDRBlock: %q format error", subnet.Spec.CIDRBlock)
-	}
 	excludeIps := subnet.Spec.ExcludeIps
 	for _, ipr := range excludeIps {
 		// v6 ip address can not use upper case
```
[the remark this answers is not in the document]
zhao-xuan: Addressed in 9d162ee. `ValidateSubnet` now rejects a non-empty `excludeIps` up front when the subnet is underlay-without-CIDR (`excludeIps must be empty for underlay subnet %s without cidrBlock`). Because that check returns early, the CIDR-dependent excludeIps validation loop is never reached in MAC-only mode.

pkg/controller/pod.go
```diff
@@ -2186,6 +2186,24 @@ func (c *Controller) acquireAddress(pod *v1.Pod, podNet *kubeovnNet) (string, st
...
+				return "", "", "", podNet.Subnet, err
+			}
+			// Use the existing MAC annotation
+			return "", "", annoMAC, podNet.Subnet, nil
+		}
+
+		// Generate a new MAC address
+		mac := util.GenerateMac()
+		return "", "", mac, podNet.Subnet, nil
+	}
```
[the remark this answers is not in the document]
zhao-xuan: Addressed in d25fe6c. MAC allocation is no longer a fresh random per call. It now goes through IPAM — `c.ipam.GetRandomAddress(key, portName, macPointer, subnet.Name, ...)` — which reserves/stores the MAC in the persistent allocator (NicToMac / MacToPod) keyed by portName. Retries therefore reuse the same MAC instead of flipping it, and a supplied MAC is validated for conflicts via GetStaticMac. This also keeps the pod's MAC tracked so GC won't reap it.

pkg/daemon/ovs_linux.go
```diff
@@ -1277,88 +1289,92 @@ func configureNic(link, ip string, macAddr net.HardwareAddr, mtu int, detectIPv4
...
-	ipAddMap := make(map[string]netlink.Addr)
-	ipAddrs, err := util.AddrList(nodeLink, unix.AF_UNSPEC)
-	if err != nil {
-		err = fmt.Errorf("failed to list addresses on link %s: %w", link, err)
-		klog.Error(err)
-		return err
-	}
+	// MAC-only mode (BYO-DHCP / external DHCP): skip all IP address configuration when no
+	// IP was allocated. The interface keeps only its MAC and the VM gets its address from DHCP.
+	if ip != "" {
```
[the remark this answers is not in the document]
zhao-xuan: Addressed in 9d162ee. The address list/flush now runs up front regardless of mode: existing non-link-local addresses are collected into `ipDelMap` and deleted even when `ip == ""` (MAC-only). Only the "add assigned IP" path is skipped, so the interface is left with just its MAC and no leftover non-link-local addresses that could cause stray routing/ARP.

pkg/controller/subnet.go
zhao-xuan: Done in d25fe6c. Added a dedicated `ProtocolMac = "Mac"` protocol. `formatSubnet` sets `subnet.Spec.Protocol = kubeovnv1.ProtocolMac` for underlay-without-CIDR subnets, and the validator, IPAM and daemon paths branch on this protocol. The subnet now reports `Protocol: Mac` and reaches Ready (covered by the new e2e in 7ff9202).

pkg/controller/pod.go
zhao-xuan: Done in d25fe6c. `acquireAddress` now checks the per-interface MAC annotation first via `perInterfaceMACAnnotationKey(nadName, nadNamespace, ifaceName)`, and only falls back to the provider-level MAC annotation when that is empty, before handing the MAC to IPAM.

pkg/controller/subnet.go
zhao-xuan: Done in d25fe6c. `handleAddOrUpdateSubnet` now registers the MAC-only subnet in IPAM via `c.ipam.AddOrUpdateSubnet(name, "", "", nil)` (CIDR-less branch that creates a `NewMacOnlySubnet` entry with no IP ranges). Because the subnet entry exists in IPAM, `gc.go` `checkIPOwnerExists` (the L311 path you linked) no longer treats the mac-only IPs/LSPs as orphaned, so GC won't clean them.

pkg/controller/init.go
oilbeater, line 463: GetStaticAddress will fail if ip is empty, which will ignore adding mac to the ipam when init.

oilbeater (changes requested):
Please run `make lint` and signoff the commit to fix the E2E failures.

zhao-xuan:
@oilbeater I've fixed and signed off as requested, requesting for another round of review.
FYI, the pipeline seems to be flaky?

oilbeater (approved):

oilbeater:
Thanks! @zhao-xuan

PR #7039, opened 2026-07-14
fix(ovn-central): self-recover from stuck joining cluster state

### LINKED ISSUE #6907: [BUG] Ovn-central permanently stuck in "joining cluster" after peer server ID change: health check blind spot prevents self-recovery
### Kube-OVN Version

v1.14.20

### Kubernetes Version

v1.35.0

Azure Linux 3 6.6.139.1-1.azl3

### Description

In a 3-node `ovn-central` deployment, one pod can enter a permanent `Status: joining
cluster` state on both the NB and SB Raft clusters with no automated recovery. The pod
passes all liveness and readiness probes indefinitely, Kubernetes never restarts it, and
the cluster silently runs with degraded 2/3 quorum until manually intervened.

This was observed in production where a pod was stuck for at least 7 days (evidenced by
34,090+ disconnection events on NB). The two healthy nodes continued serving traffic, but
the cluster was one node failure away from total quorum loss with no alerting.

I believe the root cause is a chain of three compounding bugs. A recovery mechanism already exists
in the code (`ovn_db_pre_start` stale DB cleanup in `start-db.sh`), but it is
unreachable because the health check does not recognise `joining cluster` as unhealthy —
so Kubernetes never restarts the pod to trigger the cleanup. Even if the cleanup did run,
it leaves the stale raft header file (`.hdr`) on disk, causing the very next startup to
immediately rebuild an equally broken database from the same stale peer IDs, looping back
to the stuck state.

All issues confirmed present in `master` at commit `69bf350e7`.

---

### Bug 1: Health check does not detect `Status: joining cluster`

**File:** `dist/images/ovn-healthcheck.sh`

**Severity:** High — this is the root cause of permanent stuck state

The liveness/readiness probe script only exits non-zero for two conditions:

```bash
# Catches: "Status: failed"
if ! echo ${nb_status} | grep -v "failed"; then exit 1; fi

# Catches: "Status: disconnected" + "Role: candidate"
if echo ${nb_status} | grep "disconnected" && echo ${nb_role} | grep "candidate"; then exit 1; fi
```

`Status: joining cluster` is not caught. A pod permanently stuck joining passes all
probes indefinitely, so Kubernetes never restarts it and the self-healing loop
(Bug 2 below) never triggers.

**Observed impact:** A pod was confirmed stuck in `joining cluster` for at least 7 days
(34,090+ disconnections on NB, with a separate SB issue on the same pod). The cluster ran
with degraded 2/3 quorum for this entire period with no alerting or automatic recovery.

**Suggested fix:** Add `joining` as a health check failure condition:

```bash
if echo ${nb_status} | grep "joining"; then
    echo "nb health check failed: stuck in joining state"
    exit 1
fi
```

To avoid restarting a pod that is legitimately mid-join at startup, this should be gated
behind the pod's `initialDelaySeconds`. The existing `initialDelaySeconds: 30` is likely
too short for large clusters; a value of 90–180 seconds is more appropriate, giving a
total grace period of ~120–210 seconds before a stuck-joining pod is killed — consistent
with the 120-second threshold already used in `ovn_db_pre_start`.

---

## Bug 2: Stale DB cleanup is unreachable without Bug 1, and incomplete even when it runs

**File:** `dist/images/start-db.sh`, function `ovn_db_pre_start`

**Severity:** Medium — recovery mechanism exists but has two independent flaws

The startup script checks whether a clustered DB file has been in `has not joined the
cluster` state for more than 120 seconds, and if so deletes it so the next start can
attempt a clean join:

```bash
if echo $msg | grep -q 'has not joined the cluster'; then
    local birth_time=$(stat --format=%W $db_file)
    local now=$(date +%s)
    if [ $(($now - $birth_time)) -ge 120 ]; then
        echo "ovn db file $db_file exists for more than 120s, remove it."
        rm -f "$db_file" || return 1
    fi
    return
fi
```

This logic runs exactly once — at pod startup. The intended recovery chain is:

1. Health check fails → Kubernetes kills the pod
2. Pod restarts → `ovn_db_pre_start` detects a >120s-old unjoined DB → deletes it → clean join

**Flaw 1 — unreachable:** Because Bug 1 prevents step 1, this cleanup never runs. A pod
can be stuck indefinitely even though the recovery mechanism exists in the code.

**Flaw 2 — incomplete:** Even when the cleanup does run (e.g. after a manual pod delete),
it only removes the `.db` file. The `.hdr` raft header file is left on disk. On the next
start, the `elif [ -e "$hdr_file" ]` branch immediately calls `ovsdb-tool rejoin-cluster`
using the stale header, rebuilding an equally broken DB with the same dead peer IDs. The
node re-enters `joining cluster` on the very next startup. See Bug 3.

**Race condition:** If the pod restarts quickly (e.g. rapid crash loop), the DB file may
be less than 120 seconds old and the cleanup is skipped entirely. This was directly
observed: DB files had `birth == modified` ~100 seconds before the current startup, just
under the 120-second threshold, causing the cleanup to be skipped.

**Combined fix for Bugs 2 and 3:** Delete the `.hdr` file alongside the `.db` file:

```bash
if [ $(($now - $birth_time)) -ge 120 ]; then
    echo "ovn db file $db_file exists for more than 120s, remove it."
    rm -f "$db_file" || return 1
    rm -f "$hdr_file" || true    # prevents .hdr rebuilding the same broken DB
fi
```

---

## Bug 3: Stale `.hdr` file causes immediate re-breakdown after DB cleanup

**File:** `dist/images/start-db.sh` (uses `.hdr`), `pkg/ovn_leader_checker/ovn.go`
(`backupRaftHeader`)

**Severity:** Medium — causes the stuck-joining state to persist across restarts

The `backupRaftHeader` function in `pkg/ovn_leader_checker/ovn.go` periodically writes
`/etc/ovn/ovnnb_db.hdr` containing the server ID, cluster ID, local address, and the
current peer configuration. When `ovn_db_pre_start` finds no `.db` file but an `.hdr`
file exists, it unconditionally calls:

```bash
elif [ -e "$hdr_file" ]; then
    ovsdb-tool rejoin-cluster "$db_file" "$hdr_file" "$local_addr" ${remote_addr[*]}
```

If the `.hdr` was written when a peer had server ID `ffba`, and that peer has since been
replaced with a new server ID (`8999`) at the same IP, the rebuilt DB connects to `ffba`
— which no longer exists. The node re-enters `joining cluster` and `Connections:` shows
`->ffba` permanently.

This means the Bug 2 stale DB cleanup, even when it fires correctly, only clears the
`.db` file. The `.hdr` file persists and causes the next startup to immediately recreate
an identically broken DB. The two bugs form a loop: clean up DB → restart → `.hdr`
rebuilds broken DB → stuck again → health check passes → no further restart.

The `.hdr` file is never invalidated when cluster membership changes.

**Suggested fix:** See Bug 2 combined fix — delete `.hdr` alongside `.db` in the
stale-join cleanup. This is a one-line change to `start-db.sh`.

---

## Bug 4: No mechanism to reconcile Raft membership against running server IDs at the same addresses

**Severity:** Medium — architectural gap that allows the cluster to silently accumulate stale members

When a node is reprovisioned (reimaged) and comes back with the same IP but a new server
ID, two things happen:

1. The new node issues `join-cluster`, which adds it as a new member — the leader runs an
   `add-server` configuration change
2. The old server ID at the same IP remains permanently in the Raft configuration until
   manually kicked via `ovs-appctl cluster/kick`

There is no mechanism in kube-ovn to detect this case. The stale server ID is never
kicked, and any pods that held a join-in-progress DB referencing the old ID are left
permanently stuck.

In a 3-node cluster, an unkicked stale entry raises the effective cluster size to 4,
increasing the quorum threshold from 2 to 3 and making the cluster more fragile than
intended until the entry is manually removed.

**Suggested fix:** A reconciler should query `cluster/status` on a healthy member and
kick any server ID whose registered address IP appears in `NODE_IPS` but whose server ID
does not match the currently running `ovsdb-server` at that IP.

### Steps To Reproduce

The core issue is that `ovn-central` has no self-recovery from a stuck join. The simplest
way to trigger it:

1. Deploy a 3-node `ovn-central` cluster (default `hostPath` DB storage)
2. Reprovision any one of the three OVN central nodes so that its DB files are wiped
   (e.g. node reimaged, or `/etc/origin/ovn/` manually cleared)
3. The replacement pod starts fresh and calls `join-cluster` targeting another node.
   The joining pod's DB now contains the current cluster peer server IDs.
4. Before the join completes, reprovision a **second** OVN central node at the same IP.
   This second node gets a new server ID. The old server ID is never kicked from the
   Raft cluster.
5. The first pod (from step 3) restarts for any reason (OOM, node pressure, upgrade).
   Its DB still references the old server ID from step 4, which no longer exists.
6. The pod enters `Status: joining cluster`, connects to the IP where the old server ID
   used to live, gets no valid response, and loops indefinitely.
7. All health probes pass. The pod is never restarted. The cluster stays in degraded
   2/3 quorum with no alerting.

**Simpler single-event trigger** (sufficient to demonstrate issue):

1. On any OVN central node, stop the pod, wipe `/etc/origin/ovn/`, and restart
2. The pod joins successfully and writes its DB
3. Stop the pod again within 120 seconds of the DB file being created
4. Restart — `ovn_db_pre_start` skips cleanup (file is <120s old), the pod
   re-enters `joining cluster`, and all health checks pass indefinitely

### Current Behavior

- The stuck pod shows `Status: joining cluster`, `Log: [1, 1]`, `Leader: unknown`,
  `Servers:` empty, and `Connections: -><dead-server-id>` in `cluster/status`
- The pod passes all health checks — `ovn-healthcheck.sh` does not detect `joining` as
  an unhealthy state
- Kubernetes never restarts the pod; the stuck state is permanent until manual intervention
- The healthy 2-node cluster continues serving traffic but has no fault tolerance
- NB and SB can be stuck independently for different reasons with no per-database health
  reporting — both fail silently

Specifically, `ovn-healthcheck.sh` only catches:
```bash
# Catches: "Status: failed"
if ! echo ${nb_status} | grep -v "failed"; then exit 1; fi

# Catches: "Status: disconnected" + "Role: candidate"
if echo ${nb_status} | grep "disconnected" && echo ${nb_role} | grep "candidate"; then exit 1; fi
```
`Status: joining cluster` is not caught.

Additionally, the stale DB cleanup in `ovn_db_pre_start` only deletes the `.db` file, not
the `.hdr` file. On the subsequent restart, the `elif [ -e "$hdr_file" ]` branch
unconditionally calls `ovsdb-tool rejoin-cluster` from the stale header, rebuilding an
identically broken DB.

### Expected Behavior

- A pod that has been in `Status: joining cluster` for longer than the startup grace period
  should fail its health check, causing Kubernetes to restart it
- On restart, `ovn_db_pre_start` should detect the stale DB (>120s old, unjoined), delete
  both the `.db` and `.hdr` files, and perform a clean `join-cluster` against the current
  live cluster members
- The pod should rejoin the cluster as a new member within one restart cycle and restore
  3/3 quorum automatically

09mulcs:
# Pull Request

- [ X ] Make sure you have followed [Kube-OVN Code Style](https://github.com/kubeovn/kube-ovn/blob/master/CODE_STYLE.md).

## What type of this PR

- Bug fixes

## Description

Fixes #6907.

A 3-node `ovn-central` pod could enter a permanent `Status: joining cluster` state on the NB/SB Raft clusters with no automated recovery. The pod would pass all health checks indefinitely, leaving the cluster silently degraded at 2/3 quorum until manually intervened.

Three compounding bugs caused this:

**Bug 1 — health check blind spot** (`dist/images/ovn-healthcheck.sh`)  
The liveness probe only detected `Status: failed` and `disconnected+candidate`. `Status: joining cluster` was not treated as unhealthy, so Kubernetes never restarted the stuck pod. Fixed by adding `joining` as a failure condition.

**Bug 2 — stale DB cleanup incomplete** (`dist/images/start-db.sh`)  
When `ovn_db_pre_start` detected a >120s-old unjoined DB 36907and deleted it, the `.hdr` raft header file was left on disk. On the next startup, `ovsdb-tool rejoin-cluster` would rebuild an identically broken DB from the stale peer IDs in the header, immediately re-entering `joining cluster`. Fixed by deleting the `.hdr` alongside the `.db`.

**Bug 3 — liveness probe kills pod before cleanup threshold** (`charts/`, `dist/images/install.sh`)  
With `failureThreshold: 5` × `periodSeconds: 15`, the liveness probe could kill a pod after only 105s — before the 120s threshold in `ovn_db_pre_start` that triggers the stale DB cleanup. Increased `failureThreshold` to 7 (135s) to guarantee the cleanup always runs before the pod is restarted.

### Recovery chain after this fix

1. Pod enters `joining cluster` → health check fails after 135s → Kubernetes restarts pod
2. On restart, `ovn_db_pre_start` finds a >120s-old unjoined DB → deletes both `.db` and `.hdr`
3. Pod performs a clean `join-cluster` → rejoins as a new member → 3/3 quorum restored

Fixes #6907

### DIFF
--- a/charts/kube-ovn-v2/templates/central/central-deployment.yaml
+++ b/charts/kube-ovn-v2/templates/central/central-deployment.yaml
@@ -200,7 +200,7 @@ spec:
                 - /kube-ovn/ovn-healthcheck.sh
             initialDelaySeconds: 30
             periodSeconds: 15
-            failureThreshold: 8
+            failureThreshold: 10
             timeoutSeconds: 45
       nodeSelector:
         kubernetes.io/os: "linux"
--- a/charts/kube-ovn-v2/templates/central/central-deployment.yaml
+++ b/charts/kube-ovn-v2/templates/central/central-deployment.yaml
@@ -200,7 +200,7 @@ spec:
                 - /kube-ovn/ovn-healthcheck.sh
             initialDelaySeconds: 30
             periodSeconds: 15
-            failureThreshold: 5
+            failureThreshold: 8
             timeoutSeconds: 45
       nodeSelector:
         kubernetes.io/os: "linux"
--- a/charts/kube-ovn/templates/central-deploy.yaml
+++ b/charts/kube-ovn/templates/central-deploy.yaml
@@ -227,7 +227,7 @@ spec:
                 - /kube-ovn/ovn-healthcheck.sh
             initialDelaySeconds: 30
             periodSeconds: 15
-            failureThreshold: 8
+            failureThreshold: 10
             timeoutSeconds: 45
       {{- $nodeSelector := index .Values "ovn-central" "nodeSelector" }}
       {{- if $nodeSelector }}
--- a/charts/kube-ovn/templates/central-deploy.yaml
+++ b/charts/kube-ovn/templates/central-deploy.yaml
@@ -227,7 +227,7 @@ spec:
                 - /kube-ovn/ovn-healthcheck.sh
             initialDelaySeconds: 30
             periodSeconds: 15
-            failureThreshold: 5
+            failureThreshold: 8
             timeoutSeconds: 45
       {{- $nodeSelector := index .Values "ovn-central" "nodeSelector" }}
       {{- if $nodeSelector }}
--- a/dist/images/install.sh
+++ b/dist/images/install.sh
@@ -7815,7 +7815,7 @@ $(ovn_central_tls_env)
                 - /kube-ovn/ovn-healthcheck.sh
             initialDelaySeconds: 30
             periodSeconds: 15
-            failureThreshold: 5
+            failureThreshold: 8
             timeoutSeconds: 45
       nodeSelector:
         kubernetes.io/os: "linux"
--- a/dist/images/install.sh
+++ b/dist/images/install.sh
@@ -7815,7 +7815,7 @@ $(ovn_central_tls_env)
                 - /kube-ovn/ovn-healthcheck.sh
             initialDelaySeconds: 30
             periodSeconds: 15
-            failureThreshold: 8
+            failureThreshold: 10
             timeoutSeconds: 45
       nodeSelector:
         kubernetes.io/os: "linux"
--- a/dist/images/ovn-healthcheck.sh
+++ b/dist/images/ovn-healthcheck.sh
@@ -48,3 +48,12 @@ if echo ${sb_status} | grep "disconnected" && echo ${sb_role} | grep "candidate"
     echo "sb health check failed"
     exit 1
 fi
+
+if echo ${nb_status} | grep -q "joining"; then
+    echo "nb health check failed: stuck in joining state"
+    exit 1
+fi
+if echo ${sb_status} | grep -q "joining"; then
+    echo "sb health check failed: stuck in joining state"
+    exit 1
+fi
--- a/dist/images/start-db.sh
+++ b/dist/images/start-db.sh
@@ -262,6 +262,9 @@ function ovn_db_pre_start() {
                 if [ $(($now - $birth_time)) -ge 120 ]; then
                     echo "ovn db file $db_file exists for more than 120s, remove it."
                     rm -f "$db_file" || return 1
+                    # also remove the raft header so the next startup does a clean
+                    # join-cluster instead of rebuilding a broken DB from stale peer IDs
+                    rm -f "$hdr_file" || return 1
                 fi
                 return
             fi


SkalaNetworks:
Hi @09mulcs, I'll take a look at this PR

In the meantime, can you please fix your commit signature to satisfy the CI/CD

The error:
-----
There is one commit incorrectly signed off. This means that the author of this commit failed to include a Signed-off-by line in the commit message.

To avoid having PRs blocked in the future, always include Signed-off-by: Author Name <authoremail@example.com> in every commit message. You can also do this automatically by using the -s flag (i.e., git commit -s).

Here is how to fix the problem so that this code can be merged.

SkalaNetworks:
@09mulcs 

>Bug 3 — liveness probe kills pod before cleanup threshold (charts/, dist/images/install.sh)
>With failureThreshold: 5 × periodSeconds: 15, the liveness probe could kill a pod after only 105s — before the 120s >threshold in ovn_db_pre_start that triggers the stale DB cleanup. Increased failureThreshold to 7 (135s) to guarantee >the cleanup always runs before the pod is restarted.

I see you set it to 8 and not 7, and 8*15 is 120, so there's a possibility this doesn't properly fix the problem. Don't we need to increase the time it takes to > 120 so that we don't kill the pod before the script has a change to fix the DB?

failureThreshold 10 (150s) seems like a good enough bargain.

charts/kube-ovn-v2/templates/central/central-deployment.yaml
```diff
@@ -200,7 +200,7 @@ spec:
                 - /kube-ovn/ovn-healthcheck.sh
             initialDelaySeconds: 30
             periodSeconds: 15
-            failureThreshold: 5
+            failureThreshold: 8
```
SkalaNetworks: As said in my comment, this is barely 120s, there's a possibility the script doesn't have to kick in and is killed by Kubernetes

charts/kube-ovn/templates/central-deploy.yaml
```diff
@@ -227,7 +227,7 @@ spec:
                 - /kube-ovn/ovn-healthcheck.sh
             initialDelaySeconds: 30
             periodSeconds: 15
-            failureThreshold: 5
+            failureThreshold: 8
```
SkalaNetworks: Same

dist/images/install.sh
```diff
@@ -7795,7 +7795,7 @@ $(ovn_central_tls_env)
                 - /kube-ovn/ovn-healthcheck.sh
             initialDelaySeconds: 30
             periodSeconds: 15
-            failureThreshold: 5
+            failureThreshold: 8
```
SkalaNetworks: Same

SkalaNetworks (changes requested):

charts/kube-ovn-v2/templates/central/central-deployment.yaml
09mulcs: Increased to 10

dist/images/install.sh
09mulcs: Increased to 10

charts/kube-ovn/templates/central-deploy.yaml
09mulcs: Increased to 10

SkalaNetworks (approved):

09mulcs:
@SkalaNetworks  seem to have a bunch of test failures unrelated to my changes, have you seen these before or know how to work around?

SkalaNetworks:
@09mulcs can you try rebasing on main to ensure this is not something upstream?

09mulcs:
@SkalaNetworks  alas, rebasing on latest master doesn't seem to have helped

SkalaNetworks:
Seems to be a problem in the master branch, you can't do anything about it. This PR seems fine to me, thank you for you contribution.

PR #7106, opened 2026-08-01
fix(controller): reject V6Ip changes in enqueueUpdateOvnEip guard

### LINKED ISSUE #6800: [BUG] OvnEip V6Ip change not rejected by update guard
### Kube-OVN Version

v1.17.0

### Kubernetes Version

v1.31.0

Ubuntu 24.04, kernel 6.8.0-100-generic (Kind cluster)

### Description

`enqueueUpdateOvnEip` in `pkg/controller/ovn_eip.go` guards against unsupported IP/MAC changes on existing OvnEip CRs. However, the guard only checks `V4Ip` and `MacAddress` — it does not check `V6Ip`.

This means a user or automation can patch the `v6Ip` field on an existing OvnEip without any rejection or reset, causing silent inconsistency between the CR and the actual OVN Logical Router Port configuration.

The relevant code (line 42-47):

```go
if oldEip.Spec.V4Ip != "" && oldEip.Spec.V4Ip != newEip.Spec.V4Ip ||
    oldEip.Spec.MacAddress != "" && oldEip.Spec.MacAddress != newEip.Spec.MacAddress {
    klog.Infof("not support change ip or mac for eip %s", key)
    c.resetOvnEipQueue.Add(key)
    return
}
```

`V6Ip` is not checked. If it changes, the controller silently enqueues the update instead of resetting the EIP.

### Steps To Reproduce

```bash
# 1. Have a dualstack external subnet and a VPC with enableExternal: true
#    (OvnEip "tenant-vpc-external" exists with v4Ip and v6Ip)

# 2. Check current state
kubectl get ovn-eip tenant-vpc-external -o jsonpath='{.spec.v6Ip}'
# Output: fd00:100::2

# 3. Patch the V6Ip
kubectl patch ovn-eip tenant-vpc-external --type=merge -p '{"spec":{"v6Ip":"fd00:100::99"}}'

# 4. Check controller logs
kubectl -n kube-system logs deploy/kube-ovn-controller --tail=10 | grep tenant-vpc-external
```

### Current Behavior

The V6Ip change is silently accepted. Controller logs show:

```
I ovn_eip.go:51] enqueue update ovn eip tenant-vpc-external
I ovn_eip.go:214] wait ovn eip tenant-vpc-external to be ready only in the handle add process
```

No "not support change ip" message. The CR now has `v6Ip: fd00:100::99` but the OVN LRP still uses the old `fd00:100::2`. IPv6 routing is silently broken.

### Expected Behavior

The controller should detect the V6Ip change and reject it, same as it does for V4Ip:

```
I ovn_eip.go:46] not support change ip or mac for eip tenant-vpc-external
```

And the EIP should be reset to its original value.

pujitha24:
Motivation:
enqueueUpdateOvnEip guards against unsupported mutations to an
existing OvnEip by routing them to resetOvnEipQueue, which quietly
marks the EIP not-ready without erroring. That guard only checked
V4Ip and MacAddress, not V6Ip. A V6Ip-only spec change on a Ready
OvnEip therefore skipped the guard and instead went to
updateOvnEipQueue -> handleUpdateOvnEip, which already rejects the
Status/Spec V6Ip mismatch but returns a persistent error. Since
runWorker keeps re-queuing on error, this produces indefinite
rate-limited retries and repeated klog.Error log lines instead of
the same clean, silent rejection the V4Ip/MacAddress paths get.

This does not cause a crash or outage: the change is still
ultimately rejected (both by this downstream check and, when
deployed, by the validating webhook in pkg/webhook/ovn_nat_gateway.go
for Ready OvnEips). User-visible behavior for a successful update is
unchanged. The fix only closes the gap between how V6Ip and V4Ip
mutations are handled, eliminating the retry-loop/log-spam path for
V6Ip.

Approach:
Add a `V6Ip != "" && V6Ip != newV6Ip` clause to the guard in
enqueueUpdateOvnEip, mirroring the existing V4Ip clause exactly, so
V6Ip changes on an existing EIP are caught at the same point as
V4Ip/MacAddress changes and routed to resetOvnEipQueue.

Validation:
This sandbox's host is native macOS, and pkg/util (transitively
imported by pkg/controller) does not build natively on darwin
(pkg/util/ndp.go uses the Linux-only unix.ETH_P_IPV6 constant) -
verified this is pre-existing and unrelated to this change via
`git stash && go test -run TestNothing ./pkg/controller/...`, which
fails identically before this diff. No Docker/VM/qemu is available
in this sandbox to execute a cross-compiled Linux test binary, so
`go test` itself could not be run and is not claimed to have been
run. Instead:
  - `GOOS=linux GOARCH=amd64 go build ./pkg/controller/...` passes
  - `GOOS=linux GOARCH=amd64 go vet ./pkg/controller/...` passes
  - `GOOS=linux GOARCH=amd64 go test -c -o /dev/null ./pkg/controller/...`
    passes, proving the new test compiles and links against the real
    Controller struct/queues
  - `GOOS=linux GOARCH=amd64 golangci-lint run ./pkg/controller/...`
    reports 0 issues
  - The corrected boolean guard expression was extracted verbatim into
    a standalone throwaway Go program (no repo deps) and run natively
    with `go run` against 5 scenarios (v6 change, v4 change, mac
    change, v6/v4 assigned from empty, no change); all 5 produced the
    expected accept/reject verdict. The scratch file was removed
    afterward and is not part of this commit.
Added Test_enqueueUpdateOvnEip to pkg/controller/ovn_eip_test.go,
covering: V6Ip change routed to resetOvnEipQueue (not
updateOvnEipQueue), V4Ip change routed to resetOvnEipQueue
(regression guard), and V4Ip/V6Ip initial assignment from empty
routed to updateOvnEipQueue (must not be rejected).

Fixes #6800

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/pkg/controller/ovn_eip.go
+++ b/pkg/controller/ovn_eip.go
@@ -43,6 +43,7 @@ func (c *Controller) enqueueUpdateOvnEip(oldObj, newObj any) {
 	}
 	oldEip := oldObj.(*kubeovnv1.OvnEip)
 	if oldEip.Spec.V4Ip != "" && oldEip.Spec.V4Ip != newEip.Spec.V4Ip ||
+		oldEip.Spec.V6Ip != "" && oldEip.Spec.V6Ip != newEip.Spec.V6Ip ||
 		oldEip.Spec.MacAddress != "" && oldEip.Spec.MacAddress != newEip.Spec.MacAddress {
 		klog.Infof("not support change ip or mac for eip %s", key)
 		c.resetOvnEipQueue.Add(key)
--- a/pkg/controller/ovn_eip_test.go
+++ b/pkg/controller/ovn_eip_test.go
@@ -10,6 +10,77 @@ import (
 	"github.com/kubeovn/kube-ovn/pkg/util"
 )
 
+func Test_enqueueUpdateOvnEip(t *testing.T) {
+	newTestController := func() *Controller {
+		return &Controller{
+			updateOvnEipQueue: newTypedRateLimitingQueue[string]("UpdateOvnEip", nil),
+			resetOvnEipQueue:  newTypedRateLimitingQueue[string]("ResetOvnEip", nil),
+		}
+	}
+
+	t.Run("v6 ip change is rejected via resetOvnEipQueue, not updateOvnEipQueue", func(t *testing.T) {
+		ctrl := newTestController()
+		oldEip := &kubeovnv1.OvnEip{
+			ObjectMeta: metav1.ObjectMeta{Name: "eip-v6"},
+			Spec:       kubeovnv1.OvnEipSpec{V6Ip: "fc00:1::a"},
+		}
+		newEip := oldEip.DeepCopy()
+		newEip.Spec.V6Ip = "fc00:1::b"
+
+		ctrl.enqueueUpdateOvnEip(oldEip, newEip)
+
+		require.Equal(t, 1, ctrl.resetOvnEipQueue.Len())
+		require.Equal(t, 0, ctrl.updateOvnEipQueue.Len())
+		key, _ := ctrl.resetOvnEipQueue.Get()
+		require.Equal(t, "eip-v6", key)
+	})
+
+	t.Run("v4 ip change is rejected via resetOvnEipQueue", func(t *testing.T) {
+		ctrl := newTestController()
+		oldEip := &kubeovnv1.OvnEip{
+			ObjectMeta: metav1.ObjectMeta{Name: "eip-v4"},
+			Spec:       kubeovnv1.OvnEipSpec{V4Ip: "192.168.0.5"},
+		}
+		newEip := oldEip.DeepCopy()
+		newEip.Spec.V4Ip = "192.168.0.6"
+
+		ctrl.enqueueUpdateOvnEip(oldEip, newEip)
+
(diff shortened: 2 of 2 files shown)


pujitha24:
Thanks for the review. Fixed the workqueue leak Copilot flagged: `newTestController` in `Test_enqueueUpdateOvnEip` now takes `t *testing.T` and registers `t.Cleanup` to `ShutDown()` both `updateOvnEipQueue` and `resetOvnEipQueue`, mirroring the pattern already used in `ip_test.go`. All four subtests were updated to pass `t` through.

On the failing `Iptables VPC NAT Gateway E2E` check: it fails on an `iperf`-based bandwidth QoS assertion in `test/e2e/iptables-eip-qos`, a suite unrelated to this PR's `enqueueUpdateOvnEip` change. I couldn't find the same failure on recent master runs, so this looks like a bandwidth-measurement flake rather than something caused by this diff — happy to take another look if it keeps failing after a re-run.

pujitha24:
Closing this to keep my open PR queue manageable - it has been open a while without review, and I would rather not leave stale PRs sitting in your queue. The change itself still applies; happy to reopen and rebase if it is useful to you.

Issue #7092, opened 2026-07-29
[Feature Request] Add VpcEgressGateway Prometheus metrics and sidecar scraping support


zhangzujian:
### Description

`VpcEgressGateway` already exposes useful operational state through its CRD status and programs OVN logical router policies, ECMP nexthops and optional BFD sessions. Operators still lack a coherent Prometheus and flow-telemetry surface for detecting degraded gateways and investigating NAT traffic.

This issue tracks independently implementable observability targets. Each target should remain opt-in where it changes the gateway workload and should provide its own tests.

#### Target 1: Export controller-level `VpcEgressGateway` status metrics

Expose metrics derived only from the Kubernetes object and workload status:

```text
kube_ovn_vpc_egress_gateway_info{namespace,name,vpc,external_subnet,traffic_policy,bfd_enabled}
kube_ovn_vpc_egress_gateway_ready{namespace,name}
kube_ovn_vpc_egress_gateway_phase{namespace,name,phase}
kube_ovn_vpc_egress_gateway_replicas{namespace,name,type="desired|ready"}
kube_ovn_vpc_egress_gateway_workload_nodes{namespace,name,node}
kube_ovn_vpc_egress_gateway_policy_count{namespace,name,type="selector|policy"}
```

Acceptance criteria:

- metrics are exported by `kube-ovn-controller`;
- deleted gateways do not leave stale time series after the next export cycle;
- unit tests cover ready, not-ready, processing and deleted gateway cases.

#### Target 2: Export OVN logical route policy and ECMP nexthop metrics

Expose metrics derived from OVN NB logical router policies owned by a gateway:

```text
kube_ovn_vpc_egress_gateway_nexthops{namespace,name,address_family}
kube_ovn_vpc_egress_gateway_ovn_policy_count{namespace,name,address_family,priority,type}
```

Acceptance criteria:

- policies are selected by gateway external IDs;
- `Cluster` and `Local` traffic policies are covered;
- IPv4, IPv6 and dual-stack gateways are represented correctly;
- stale or missing policies are observable without changing reconcile results.

#### Target 3: Export BFD session metrics

Expose BFD state for sessions owned by a gateway:

```text
kube_ovn_vpc_egress_gateway_bfd_session_status{namespace,name,address_family,nexthop,status}
kube_ovn_vpc_egress_gateway_bfd_session_count{namespace,name,status}
```

Acceptance criteria:

- BFD rows are selected by gateway external IDs;
[...]

PR #7116, opened 2026-08-03
Add observability for VPC egress gateways

zhangzujian:
## Summary

Add an optional native observability sidecar for each `VpcEgressGateway` workload.

- expose internal and external interface byte, packet, drop, and error counters;
- expose low-cardinality conntrack NAT flow, packet, and byte metrics;
- emit versioned JSON Lines NAT flow lifecycle records to container stdout;
- reconcile a hot-reloadable ConfigMap and, when metrics are enabled, a per-gateway headless Service and ServiceMonitor;
- keep all observability capabilities disabled by default and independent from data-plane readiness.

Part of #7092.

## API

All settings live under `spec.observability`:

```yaml
observability:
  resources: {}
  interfaceMetrics:
    enabled: true
  conntrack:
    metrics:
      enabled: true
    log:
      enabled: true
      events: [start, end]
      rateLimit:
        recordsPerSecond: 100
        burst: 1000
      filters:
        include: []
        exclude: []
  serviceMonitor:
    labels: {}
    annotations: {}
```

Collector switches, log filters, and rate limits are reloaded through the controller-managed ConfigMap without replacing gateway pods. Enabling observability for the first time, disabling it completely, changing the observer resources, or changing the workload image still updates the Deployment.

## Runtime behavior

- Kubernetes 1.29 or later is required for restartable init containers. The controller uses a zero-replica server-side dry-run to confirm that the API server preserves `restartPolicy: Always`, so a disabled `SidecarContainers` feature gate is also detected. Unsupported results are cached; transient probe failures are retried. The controller skips new observer injection for definitive unsupported results, records `ObservabilityConfigured=False`, and continues reconciling the data plane. On a transient capability, ConfigMap, or Service error, it preserves the last-known-good observer PodSpec and auxiliary resources instead of replacing a forwarding Pod.
- The dedicated observer binary runs as UID/GID 65534 with only `NET_ADMIN` admitted to the capability bounding set, all other capabilities dropped, and a read-only root filesystem. The official binary carries only the `CAP_NET_ADMIN` file capability, and `allowPrivilegeEscalation` is enabled so that the trusted capability becomes effective after the non-root launcher `exec`.
- The launcher falls back to `sleep infinity` when an explicitly selected older gateway image does not contain the observer binary.
- `/metrics` and `/healthz` listen on port 10666. Every gateway executes the observer binary for its liveness probe and checks the same endpoint over loopback, so health does not depend on node-to-Pod reachability. The probe also accepts the older-image `sleep infinity` fallback when the binary is absent. A private Prometheus registry avoids `go_*`, `process_*`, and `promhttp_*` metrics.
- Interface collection resolves the primary and external Multus interfaces once from `network-status`, then reads `/proc/net/dev` on each scrape.
- Conntrack collection subscribes with one decoder worker before the initial dump, preserving NEW/DESTROY delivery order. It maintains an on-demand bounded 65,536-entry cache and uses bounded 4,096-entry event and log queues. Interface-only observers do not preallocate conntrack cache buckets, and log backpressure never blocks netlink event processing.
- Flow logs use schema `v1`, expose `start` and `end` events only, write JSON Lines to stdout, and keep diagnostics on stderr.
- Conntrack accounting is never enabled or changed by the observer. Packet and byte values are omitted when accounting is unavailable.

## Monitoring resources

When either metrics collector is enabled, the controller creates one headless Service and one ServiceMonitor for the gateway. The Service uses `publishNotReadyAddresses: true` and numeric `targetPort: 10666` for native-sidecar compatibility.

The Prometheus Operator CRD is a soft dependency. If it is absent, the gateway remains operational and reports `ServiceMonitorReady=False`; the delayed reconcile loop creates the ServiceMonitor after the CRD appears. User labels and annotations are supported, while required selector labels cannot be overridden.

This change only adds the controller RBAC required to manage these resources. It does not add Helm ServiceMonitor templates or values.

## Testing

- observer unit tests cover NAT record classification, complete start/end JSON lifecycle records, accounting omission, include/exclude semantics, structured port ranges, interface resolution caching, counter deltas, initial dump/event ordering, hot-reloaded runtime settings, private registry contents, and the loopback health-check client;
- controller fake and dynamic-client tests cover restartable sidecar security/resources, binary-aware loopback liveness probing, API-server capability probing and retry, last-known-good PodSpec/resource preservation after transient failures, hot-switch pod-spec stability, per-gateway ConfigMap/Service/ServiceMonitor creation, unsupported Kubernetes behavior, and the missing-ServiceMonitor-CRD soft failure;
- the VpcEgressGateway E2E case uses two replicas and validates external interface counter growth under real SNAT traffic, conntrack metrics, schema-valid matching `start`/`end` flow records after explicitly deleting the selected active conntrack entry, hot reload on both replicas without pod replacement, collector failure isolation, interface-only RSS at or below 20 MiB, observer restart, and optional ServiceMonitor discovery across the existing IPv4, IPv6, and dual-stack jobs.

Local verification completed:

```text
go test ./pkg/vegobserver ./pkg/controller -count=1
go test -race ./pkg/vegobserver -count=1
go test ./pkg/apis/kubeovn/v1 ./pkg/client/applyconfiguration/... -count=1
go test ./test/e2e/vpc-egress-gateway -run '^$' -count=0
go vet ./pkg/vegobserver ./pkg/controller ./cmd/vpc-egress-gateway-observer ./test/e2e/vpc-egress-gateway
make verify-crd
```


zhangzujian:
Documentation is available in kubeovn/docs#419.

pkg/vegoobserver/conntrack_collector.go
```diff
@@ -0,0 +1,370 @@
...
+		if err := eventConnection.Close(); err != nil {
+			returnErr = errors.Join(returnErr, fmt.Errorf("close conntrack event netlink: %w", err))
+		}
+	}()
+	if err := eventConnection.SetReadBuffer(4 << 20); err != nil {
+		c.metrics.errors.WithLabelValues(append(c.identity.labels(), "conntrack", "set_read_buffer")...).Inc()
+		writeDiagnostic(c.metrics, c.identity, diagnostics, "conntrack", "set conntrack read buffer: %v\n", err)
+	}
+	events := make(chan conntrack.Event, DefaultEventBuffer)
+	errorsChannel, err := eventConnection.Listen(events, 2, netfilter.GroupsCT)
```
zhangzujian: [P1] Preserve conntrack lifecycle ordering

`Listen(events, 2, ...)` starts two decoder workers. The underlying netlink connection serializes `Receive` only until each datagram is returned; the workers then decode and send independently, so a NEW followed by DESTROY can reach `processEvent` in reverse order. The current handlers will emit `end` before `start`, then reinsert the already-destroyed flow and leave `activeFlows`/the cache with a ghost entry. Please use a single ordered worker (or explicitly sequence events) and add a deterministic NEW/DESTROY ordering test.

pkg/controller/vpc_egress_gateway_observability.go
```diff
@@ -0,0 +1,411 @@
...
+		return vpcEgressObserverState{}
+	}
+	if err := c.reconcileVpcEgressObserverConfigMap(gw, resourceName, labels, data); err != nil {
+		c.setVpcEgressGatewayObservabilityCondition(gw, kubeovnv1.ObservabilityConfigured, corev1.ConditionFalse, "ConfigMapReconcileFailed", err.Error())
+		if observabilityMetricsEnabled(gw.Spec.Observability) {
+			c.setVpcEgressGatewayObservabilityCondition(gw, kubeovnv1.ServiceMonitorReady, corev1.ConditionFalse, "ConfigMapReconcileFailed", err.Error())
+		}
+		c.deleteOwnedService(gw, resourceName)
+		c.deleteOwnedServiceMonitor(gw, resourceName)
+		return vpcEgressObserverState{}
```
zhangzujian: [P1] Keep the last-known-good observer on transient resource failures

After observability is already running, a transient ConfigMap update failure returns an empty state here; the Service failure branch below does the same. The caller then hashes and updates a Deployment with the observer removed. Because gateway Deployments use `maxUnavailable: 1` and `maxSurge: 0`, a single-replica gateway loses its only forwarding Pod for an observability control-plane error. Please preserve the last-known-good injected state/resources, or stop the workload update while persisting the condition and requeueing. A post-enable transient-failure test would cover this isolation contract.

pkg/controller/vpc_egress_gateway_observability.go
```diff
@@ -0,0 +1,411 @@
...
+		resources = corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("20m"), corev1.ResourceMemory: resource.MustParse("64Mi")},
+			Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("200m"), corev1.ResourceMemory: resource.MustParse("256Mi")},
+		}
+	}
+	launcher := fmt.Sprintf("if [ -x /kube-ovn/vpc-egress-gateway-observer ]; then exec /kube-ovn/vpc-egress-gateway-observer --config %s --network-status %s; fi; exec sleep infinity", vpcEgressObserverConfigPath, vpcEgressObserverNetworkStatusPath)
+	healthCheck := "if [ ! -x /kube-ovn/vpc-egress-gateway-observer ]; then exit 0; fi; exec /kube-ovn/vpc-egress-gateway-observer --health-check"
+	livenessProbe := &corev1.Probe{ProbeHandler: corev1.ProbeHandler{Exec: &corev1.ExecAction{Command: []string{"/bin/sh", "-ec", healthCheck}}}, PeriodSeconds: 10, FailureThreshold: 3}
+	if useHTTPProbe {
+		livenessProbe.ProbeHandler = corev1.ProbeHandler{HTTPGet: &corev1.HTTPGetAction{Path: "/healthz", Port: intstr.FromInt32(vpcEgressObserverPort)}}
```
zhangzujian: [P1] Make the older-image fallback healthy in the default VPC

The launcher intentionally falls back to `sleep infinity` when the selected gateway image lacks the observer binary, but this branch replaces the conditional exec probe with an unconditional HTTP probe. Nothing listens on port 10666 in fallback mode, so default-VPC sidecars are restarted forever instead of remaining in the promised fail-open sleep state. Please retain a binary-aware probe for this case (or only select HTTP when the binary is known to exist) and add the default-VPC older-image test.

zhangzujian (commented):
I found three correctness issues that should be fixed before merge. This is my own PR, so GitHub does not allow an approval or request-changes review; I am submitting them as blocking COMMENT findings.

pkg/vegoobserver/conntrack_collector.go
zhangzujian: Fixed in e3035c1b4. Conntrack event subscription now uses a single decoder worker, preserving NEW/DESTROY delivery order. The regression test records the Listen worker count and requires exactly one worker.

pkg/controller/vpc_egress_gateway_observability.go
zhangzujian: Fixed in e3035c1b4. Transient capability-probe, ConfigMap, and Service failures now reuse the observer container and volumes from the current Deployment and retain the last-known-good auxiliary resources. Regression tests also verify that the existing image/resources are preserved instead of changing the forwarding PodSpec.

pkg/controller/vpc_egress_gateway_observability.go
zhangzujian: Fixed in e3035c1b4. All gateways now use the binary-aware exec liveness probe, which checks /healthz over loopback when the observer exists and succeeds when an older image uses the sleep fallback. The default-VPC regression test covers the fallback command.

PR #7118, opened 2026-08-03
fix: use fully qualified CRD names in cleanup.sh

### LINKED ISSUE #7117: [BUG] cleanup.sh uses short CRD names, can delete other projects' resources (e.g. Calico IPPools)
### Kube-OVN Version

all versions

### Kubernetes Version

1.30

Debian GNU/Linux 12 (bookworm) 6.1.0-17-amd64

### Description

`dist/images/cleanup.sh` iterates over Kube-OVN CRs using short resource names, e.g.:

```bash
for ippool in $(kubectl get ippool -o name); do
  kubectl delete --ignore-not-found $ippool
done
```

Short names are resolved by the API server's discovery/shortname matching, which is not scoped to the `kubeovn.io` API group. If another CRD in the cluster exposes the same short name or plural, `kubectl get ippool` can resolve to that CRD instead.

The concrete case is Calico, which installs `ippools.crd.projectcalico.org`. Running `cleanup.sh` on such a cluster lists and deletes Calico IP pools, breaking cluster networking well beyond removing Kube-OVN.

The same pattern applies to other loops in the script whose short names are generic and collision-prone: `vip`, `eip`, `fip`, `snat`, `dnat`, `vlan`, `provider-network`, and `subnet` / `vpc` / `ip` in the finalizer-removal loop (where a wrong match strips finalizers off foreign resources).

### Steps To Reproduce

1. Create a cluster with Calico installed alongside Kube-OVN (any CNI/operator that defines an `ippools` CRD reproduces this).
2. Confirm the ambiguity: `kubectl get ippool -o name` resolves to `ippools.crd.projectcalico.org` rather than `ippools.kubeovn.io`.
3. Create a Calico IP pool (or note the existing ones): `kubectl get ippools.crd.projectcalico.org`.
4. Run `dist/images/cleanup.sh`.
5. Check the Calico IP pools again: `kubectl get ippools.crd.projectcalico.org`.

### Current Behavior

The cleanup script deletes the Calico IP pools, because the short name `ippool` resolved to `ippools.crd.projectcalico.org`. Non-Kube-OVN resources are destroyed and cluster networking is broken.

### Expected Behavior

`cleanup.sh` only touches resources in the `kubeovn.io` API group and leaves other projects' CRs untouched, regardless of which other CRDs are installed. The loops should use fully qualified names, e.g.:

```bash
for ippool in $(kubectl get ippools.kubeovn.io -o name); do
  kubectl delete --ignore-not-found $ippool
done
```

`kubectl get ... -o name` already emits fully qualified `resource.group/name` output, so the following `delete`/`patch` calls stay unambiguous. The CRD deletion block later in the same script already uses fully qualified names, so this only makes the loops consistent with it.

hawwwdi:
### Which issue(s) this PR fixes

Fixes #7117

### What this PR does / why we need it

`dist/images/cleanup.sh` looked up Kube-OVN CRs by short resource name (`ippool`, `vip`, `eip`, `snat`, `dnat`, `vlan`, `provider-network`, ...). Short names are resolved by API server discovery matching and are not scoped to the `kubeovn.io` API group, so on a cluster that also has, for example, Calico's `ippools.crd.projectcalico.org`, `kubectl get ippool -o name` can resolve to the foreign CRD — and the script then deletes resources that do not belong to Kube-OVN.

This PR switches every `kubectl get` loop in the script to the fully qualified `<plural>.kubeovn.io` form, and does the same for the finalizer-removal loop (`subnets.kubeovn.io`, `vpcs.kubeovn.io`, `ips.kubeovn.io`), where an ambiguous match would strip finalizers off foreign resources.

`kubectl get ... -o name` already emits fully qualified `resource.group/name` output, so the subsequent `delete` / `patch` calls stay unambiguous. The CRD deletion block later in the same script already used fully qualified names — this only makes the loops consistent with it.

Resources updated: `vpc-nat-gateways`, `vpc-dnses`, `vips`, `iptables-snat-rules`, `iptables-dnat-rules`, `iptables-fip-rules`, `iptables-eips`, `ovn-dnat-rules`, `ovn-snat-rules`, `ovn-fips`, `ovn-eips`, `switch-lb-rules`, `ippools`, `vlans`, `provider-networks`, plus `subnets` / `vpcs` / `ips` in the finalizer loop.

### Does this PR introduce a user-facing change?

```release-note
Fixed cleanup.sh deleting non-Kube-OVN custom resources (such as Calico IPPools) when another CRD in the cluster shares a short name with a Kube-OVN CRD.
```

### DIFF
--- a/dist/images/cleanup.sh
+++ b/dist/images/cleanup.sh
@@ -11,63 +11,63 @@ while :; do
   sleep 1
 done
 
-for gw in $(kubectl get vpc-nat-gw -o name); do
+for gw in $(kubectl get vpc-nat-gateways.kubeovn.io -o name); do
   kubectl delete --ignore-not-found $gw
 done
 
-for vd in $(kubectl  get vpc-dns -o name); do
+for vd in $(kubectl  get vpc-dnses.kubeovn.io -o name); do
   kubectl delete --ignore-not-found $vd
 done
 
-for vip in $(kubectl get vip -o name); do
+for vip in $(kubectl get vips.kubeovn.io -o name); do
    kubectl delete --ignore-not-found $vip
 done
 
-for snat in $(kubectl get snat -o name); do
+for snat in $(kubectl get iptables-snat-rules.kubeovn.io -o name); do
    kubectl delete --ignore-not-found $snat
 done
 
-for dnat in $(kubectl get dnat -o name); do
+for dnat in $(kubectl get iptables-dnat-rules.kubeovn.io -o name); do
    kubectl delete --ignore-not-found $dnat
 done
 
-for fip in $(kubectl get fip -o name); do
+for fip in $(kubectl get iptables-fip-rules.kubeovn.io -o name); do
    kubectl delete --ignore-not-found $fip
 done
 
-for eip in $(kubectl get eip -o name); do
+for eip in $(kubectl get iptables-eips.kubeovn.io -o name); do
    kubectl delete --ignore-not-found $eip
 done
 
-for odnat in $(kubectl get odnat -o name); do
+for odnat in $(kubectl get ovn-dnat-rules.kubeovn.io -o name); do
    kubectl delete --ignore-not-found $odnat
 done
 
-for osnat in $(kubectl get osnat -o name); do
+for osnat in $(kubectl get ovn-snat-rules.kubeovn.io -o name); do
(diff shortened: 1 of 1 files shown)


zhangzujian:
Thanks for your report and fix. Please add DCO in your commit(s).

zbb88888 (approved):

PR #7135, opened 2026-08-06
fix(gateway): prevent probe disconnects from restarting bfdd

zhangzujian:
## What this PR does

- apply an OpenBFDD source patch that uses `MSG_NOSIGNAL` only when sending replies to control clients
- stop sending the remainder of a control reply after the first send failure
- keep the BFD peer packet path and process-wide signal disposition unchanged
- set the VPC gateway BFD readiness probe timeout to 10 seconds, matching the liveness probe
- add regression coverage for patch packaging, application, and the readiness timeout

The existing OpenBFDD compile patch is now actually applied rather than only checked. It adds the pthread/time includes required to rebuild the pinned source on the current Ubuntu toolchain and does not change runtime behavior. The existing TTL patch behavior is unchanged.

## Why this is needed

`bfdd-control status` uses a TCP control connection. When an exec probe times out, kubelet terminates `bfdd-control`, which closes that connection. OpenBFDD sends control replies with a plain `send(2)` call. If it writes after the probe disconnects, the daemon receives `SIGPIPE` and exits with status 141, after which the container restart policy restarts it.

The BFD peer path is separate: `Session::send()` sends UDP packets with `MSG_NOSIGNAL` already. Applying the same flag only to `CommandProcessorImp::doMessageReply()` fixes the control disconnect without ignoring `SIGPIPE` for the whole daemon or changing peer session behavior.

With the OpenBFDD commit bundled in the image (`e35f43ad8d2b3f084e96a84c392528a90d05a287`), disconnecting the control client while a response was being written terminated the unpatched daemon in 5/5 reproductions. After applying this source patch, two daemons established an Up BFD session and the test sent 20 valid, multi-line `status` requests followed by immediate TCP RST disconnects. The daemon stayed alive, both peer sessions remained Up, and `/proc/<pid>/status` confirmed that `SIGPIPE` was not ignored process-wide.

The explicit 10-second readiness timeout also prevents the Kubernetes default one-second timeout from unnecessarily aborting a slow status request.

Command logging remains disabled by default because readiness and liveness probes would otherwise generate about 32 routine command records per minute per gateway Pod. Send failures continue to use the existing socket error logging.

## Tests

- apply both patches to the pinned OpenBFDD source and build it with `make -j2`
- run two patched OpenBFDD daemons with an Up BFD peer session; inject 20 valid `status` requests followed by TCP RST and verify both daemons and the peer session remain healthy
- `bash -n dist/images/start-bfdd.sh`
- `go test ./pkg/controller -run 'TestGenGatewayBFDDContainer|TestOpenBFDDControlReplyPatch|TestBFDDHealthcheck' -count=1`
- `go test ./pkg/controller -run 'TestGenGatewayBFDDContainer|TestOpenBFDDControlReplyPatch' -count=10`
- `git diff --check`

Full lint and image builds run in GitHub Actions.

### DIFF
--- a/dist/images/Dockerfile.base
+++ b/dist/images/Dockerfile.base
@@ -176,11 +176,13 @@ RUN mkdir -p /usr/src/openbfdd && \
 
 ADD OpenBFDD-compile.patch /usr/src/
 ADD OpenBFDD-allow-ttl-254.patch /usr/src/
+ADD OpenBFDD-control-no-sigpipe.patch /usr/src/
 
 RUN cd /usr/src/openbfdd && \
     rm -vf missing && \
-    git apply --no-apply /usr/src/OpenBFDD-compile.patch && \
+    git apply /usr/src/OpenBFDD-compile.patch && \
     git apply --no-apply /usr/src/OpenBFDD-allow-ttl-254.patch && \
+    git apply /usr/src/OpenBFDD-control-no-sigpipe.patch && \
     autoupdate && \
     ./autogen.sh && \
     ./configure --enable-silent-rules && \
--- a/dist/images/OpenBFDD-control-no-sigpipe.patch
+++ b/dist/images/OpenBFDD-control-no-sigpipe.patch
@@ -1,0 +1,43 @@
+From 9b195569201a77a23d7a80491d672ab64e08bf68 Mon Sep 17 00:00:00 2001
+From: zhangzujian <zhangzujian.7@gmail.com>
+Date: Thu, 6 Aug 2026 11:00:00 +0800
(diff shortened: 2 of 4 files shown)


zhangzujian (commented):
Reviewed the incremental change at `18b43f41` against `master@50556fa8`. Standards: no findings. Spec: the control-only `MSG_NOSIGNAL` handling, stop-after-send-failure behavior, compile-patch application, and 10-second readiness timeout match the PR contract. The latest checks are 80 successful and 3 path-skipped, with no failures. No blocking findings.

PR #7136, opened 2026-08-06
fix(gateway): bound bfdd control requests

zhangzujian:
### What changed

- make accepted control sockets non-blocking and close clients that do not send a complete command within 500 ms
- receive fragmented control commands until their double-NUL terminator
- give `bfdd-control` connect, write, and read operations one five-second monotonic deadline
- send complete replies with a non-blocking partial-write loop and `MSG_NOSIGNAL`
- roll back synchronous beacon operations when scheduler signaling fails
- emit rate-limited stage timings for requests that take more than 100 ms or fail while replying

### Why

OpenBFDD processes each control port serially. An idle client could occupy that thread for about ten seconds, while both the client and reply path otherwise had unbounded blocking I/O. These paths can make `bfdd-control status` exceed a probe deadline even when the BFD session set is small.

The change is limited to the TCP control path. It does not change BFD peer UDP sockets, packet transmission, session timers, or process-wide signal handling.

### Dependency

This PR is stacked on #7135 so that the minimal SIGPIPE fix remains independently reviewable. After #7135 merges, this PR will contain only the control-path hardening commit.

### Validation

- applied both Kube-OVN OpenBFDD patches to pinned OpenBFDD commit `e35f43ad8d2b3f084e96a84c392528a90d05a287` and built with `make -j2`
- normal zero-session `status`: about 5 ms
- idle-client head-of-line test: about 429 ms instead of about ten seconds
- fake server that never replies: client failed after about 5.01 seconds
- fragmented command was reassembled and answered successfully
- TCP RST during a reply left the daemon alive and produced one rate-limited failure record
- targeted Kube-OVN controller tests passed

Fixes the remaining control-path risks investigated after #7135.


dist/images/OpenBFDD-control-hardening.patch
```diff
@@ -0,0 +1,500 @@
...
++    m_inCommandName.clear();
++    m_operationQueueDepth = 0;
++    m_replyFailed = false;
++    m_replyError = 0;
++  }
++
++  void logRequestIssue()
++  {
++    if (m_requestStarted.empty() || m_commandReceived.empty())
++      return;
```
zhangzujian, line 126: The 500 ms idle-client path never produces the advertised slow timing because `m_commandReceived` is empty and this function returns immediately. That accepted request still exceeds the 100 ms threshold. Please log it with an unknown/missing command stage, or narrow the PR contract if only completed commands are intended.

dist/images/OpenBFDD-control-hardening.patch
```diff
@@ -0,0 +1,500 @@
...
++      if (remainingLength < previousLength)
++        continue;
++    }
++    else
++    {
++      int error = socket.GetLastError();
++      if (error != EAGAIN && error != EWOULDBLOCK && error != EINTR)
++        return false;
++      if (error == EINTR)
++        continue;
```
zhangzujian, line 427: `EINTR` retries bypass `WaitForSocket`, and this loop does not check the deadline before the next send. A repeated signal stream can therefore keep `SendAll` beyond the shared five-second monotonic deadline. Please check `MonoNow()` against `deadline` on every iteration; the server reply loop has the same `EINTR` gap.

zhangzujian (commented):
Reviewed the standalone `381f4994` layer against #7135. The I/O hardening is directionally sound, but the inline findings show two places where the implementation does not fully meet its bounded/observable behavior. Standards also finds that the updated `QueueOperation` signaling-failure contract is only enforced for synchronous operations; #7137 fixes the asynchronous branch, but this standalone layer still leaves that error ignored. The duplicated daemon/client select-deadline loops are a non-blocking maintainability concern. The failing iptables E2E is unrelated (`iperf` connectivity), while the image builds and relevant jobs succeeded.

PR #7137, opened 2026-08-06
fix(gateway): bound bfdd status operations

zhangzujian:
### What changed

- run `status` through a heap-owned operation with explicit control-thread and scheduler-thread references
- stop the control wait after four seconds using a monotonic deadline
- allow a late scheduler callback to finish safely and discard its result after the control client has returned
- initialize every queued operation field and roll back both synchronous and asynchronous operations when scheduler signaling fails
- make `bfdd-control` return a non-zero exit code when the daemon closes without a reply or returns `Unable to complete request`

### Why

The existing synchronous `QueueOperation` stores both the pending operation and status callback data on the control thread stack. Adding a timeout directly to its condition wait would let the scheduler access freed stack memory after the control thread returned.

The status-specific operation in this change owns its request, result, condition, and lifecycle on the heap. One reference belongs to the control thread and one to the scheduler callback. A timeout releases only the control reference, so a delayed callback never accesses the old request stack.

This remains limited to the TCP control/scheduler seam. BFD peer UDP sockets, packet transmission, session timers, and process-wide signal handling are unchanged.

### Dependency

This PR is stacked on #7136, which is itself stacked on #7135. Each layer remains independently reviewable; after the earlier PRs merge, this PR will contain only the status-operation lifetime change.

### Validation

- applied the combined patch to pinned OpenBFDD commit `e35f43ad8d2b3f084e96a84c392528a90d05a287` and built with `make -j2`
- injected a five-second scheduler callback delay: `status` exited non-zero after about 4.007 seconds
- after the timed-out request, the late callback completed, the daemon stayed alive, and a follow-up control command succeeded
- 200 status calls with concurrency two completed without failures
- targeted Kube-OVN controller tests passed

### DIFF
--- a/dist/images/OpenBFDD-control-hardening.patch
+++ b/dist/images/OpenBFDD-control-hardening.patch
@@ -1,19 +1,26 @@
+From 2c0be5a6f979968736525e52f114ca9a962ef7ef Mon Sep 17 00:00:00 2001
 From: zhangzujian <zhangzujian.7@gmail.com>
-Date: Thu, 6 Aug 2026 15:00:00 +0800
-Subject: [PATCH] Harden control request I/O
+Date: Thu, 6 Aug 2026 16:00:00 +0800
+Subject: [PATCH] Bound control I/O and status operations
 
 Bound control client I/O with a monotonic deadline, prevent idle clients from
 occupying a serial command thread for ten seconds, receive fragmented commands,
 and send complete replies without blocking or raising SIGPIPE.
 
+Run status requests through a heap-owned operation with explicit shared
+lifetime. The control thread may stop waiting after four seconds while a late
+scheduler callback safely completes and discards its result. Use a monotonic
+condition variable so wall-clock changes cannot extend that deadline.
+
 Also reject operations when scheduler signaling fails, initialize every queued
-operation field, and emit rate-limited stage timings for slow or failed control
-requests, including clients that never finish a command.
+operation field, propagate server-side request failures through the client exit
+status, and emit rate-limited stage timings for slow, timed out, or failed
+control requests, including clients that never finish a command.
 
 Signed-off-by: zhangzujian <zhangzujian.7@gmail.com>
----
+
 diff --git a/Beacon.cpp b/Beacon.cpp
-index d1b0658..f752d48 100755
+index d1b0658..757c29f 100755
 --- a/Beacon.cpp
 +++ b/Beacon.cpp
 @@ -322 +322,2 @@ void Beacon::KillSession(Session *session)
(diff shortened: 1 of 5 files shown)


dist/images/OpenBFDD-control-hardening.patch
```diff
@@ -0,0 +1,733 @@
...
++      if (remainingNs <= 0)
++      {
++        m_operationTimedOut = true;
++        return false;
++      }
++
++      uint32_t waitMs = uint32_t((remainingNs + TimeSpec::NSecPerMs - 1) /
++                                 TimeSpec::NSecPerMs);
++      waitMs = min(waitMs, uint32_t(100));
++      if (lock.LockWait(operation->condition, waitMs) < 0)
```
zhangzujian, line 423: `LockWait` ultimately uses `pthread_cond_timedwait` with a `CLOCK_REALTIME` absolute timeout. A wall-clock rollback during one nominal 100 ms wait can therefore hold this thread past the required four-second monotonic deadline. Please use a condition variable configured for `CLOCK_MONOTONIC` or another monotonic wait primitive.

dist/images/OpenBFDD-control-hardening.patch
```diff
@@ -0,0 +1,733 @@
...
++    if (!WaitForSocket(socket, true, deadline, "sending command to beacon"))
++      return false;
++  }
++  return true;
++}
++
++bool ReadReply(Socket &socket, vector<char> &buffer, const char *outPrefix,
++               const TimeSpec &deadline)
++{
++  const char *failurePrefix = "Unable to complete request";
```
zhangzujian: The new status failure paths emit `Unable to complete status request...`, but this parser recognizes only the exact prefix `Unable to complete request`. Queue/signal, wait, exception, and shutdown failures therefore leave `requestFailed` false and make `bfdd-control` exit 0. Please use one canonical failure prefix or recognize both forms, and add exit-code coverage.

dist/images/OpenBFDD-control-hardening.patch
```diff
@@ -0,0 +1,733 @@
...
++      messageReply("Unable to complete status request because the beacon is shutting down.\n");
++      return false;
++    }
++    return true;
++  }
++
+@@ -1400,2 +1645,17 @@ protected:
+-      MultiStatusCallbackInfo info;
+-      info.level = level;
++      StatusOperation *statusOperation =
```
zhangzujian: This change grows the already over-limit `handle_Status` from about 101 to 130 lines, further beyond the 100-line limit in `CODE_STYLE.md`. The all-session and single-session branches also duplicate allocation, RAII, queue-failure, and wait handling. Please extract that lifecycle into a helper.

zhangzujian (commented):
Reviewed the standalone `1a050048` layer against #7136. The heap-owned two-reference lifetime is balanced across success, timeout, late completion, and queue failure. Three inline findings remain: failure replies can still exit zero, the four-second wait is not fully monotonic at the blocking primitive, and `handle_Status` grows further beyond the documented function-length limit with duplicated all/single lifecycle code. The two failed Actions jobs are unrelated: IPv6 underlay failed only during cleanup with a local stream timeout, and iptables E2E failed on `iperf` connectivity.

PR #7138, opened 2026-08-06
fix(ovn-central): tolerate transient duplicate leaders

zhangzujian:
## What this PR does

Require three consecutive duplicate-leader observations before ovn-leader-checker restarts ovn-central. A normal observation resets the counter independently for each OVN database.

## Why this is needed

During invalid RAFT-header recovery, a rejoining ovsdb-server can temporarily report itself as a one-member leader before it applies the live cluster snapshot. The checker previously restarted on the first observation. In the IPv6 SSL HA failure from run 31082266072, the recovered member and the existing leader then restarted each other, delaying convergence until both disruptive tests timed out.

Persistent duplicate leaders still trigger the existing restart behavior after three probe cycles.

## Validation

- go test ./pkg/ovn_leader_checker -count=1
- go test -race ./pkg/ovn_leader_checker -count=1
- go vet ./pkg/ovn_leader_checker
- golangci-lint run ./pkg/ovn_leader_checker/...
- modernize -test ./pkg/ovn_leader_checker/...
- regression test repeated 100 times

Signed-off-by: zhangzujian <zhangzujian.7@gmail.com>

### DIFF
--- a/pkg/ovn_leader_checker/ovn.go
+++ b/pkg/ovn_leader_checker/ovn.go
@@ -62,7 +62,7 @@ type Configuration struct {
 	duplicateLeaderObservations map[string]int
 }
 
-func (c *Configuration) observeDuplicateLeader(database string, duplicate bool) (int, bool) {
+func (c *Configuration) observeDuplicateLeader(database string, duplicate bool, requiredObservations int) (int, bool) {
 	if !duplicate {
 		delete(c.duplicateLeaderObservations, database)
 		return 0, false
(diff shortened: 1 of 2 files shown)


zhangzujian:
Validation update: all 12 Kube-OVN HA E2E matrix jobs passed in run 31087605788. The exact previously failing SSL=true, compact=true, IPv6 job passed: https://github.com/kubeovn/kube-ovn/actions/runs/31087605788/job/92572937208. Its HA suite completed both disruptive recovery specs with 2 passed and 0 failed.

PR #7156, opened 2026-08-07
feat: decode NetworkPolicy ACL samples

zhangzujian:
Related to #7146.

## Summary

- parse decimal or hexadecimal ACL sample metadata and 64-bit local-sampling cookies
- decode the OVN observation domain into its application byte and 24-bit datapath key
- resolve the OVN sampling application, Sample row, and owning Kube-OVN NetworkPolicy ACL through one narrow interface
- emit stable allow and default-deny event models while preserving non-exclusive default-deny attribution
- reject unknown applications, missing metadata, ambiguous owners, and inconsistent ACL external IDs

## Stack

1. #7149 - configuration
2. #7150 - stable metadata allocator
3. #7148 - OVN sampling object reconciler
4. #7151 - NetworkPolicy ACL attachment
5. #7152 - sampling-only retry and enforcement isolation
6. #7153 - ownership-aware disable and cleanup
7. #7154 - controller availability metrics
8. #7155 - node-local psample delivery
9. **this PR - cookie and metadata event decoder**
10. #7157 - Linux psample listener

Base: `feature/acl-sampling-node-delivery`

## Validation

All Go commands ran in a transient systemd scope with `MemoryMax=2G`, `MemorySwapMax=0`, `GOMAXPROCS=2`, `GOMEMLIMIT=1700MiB`, and `GOFLAGS=-p=1`.

- `go test ./pkg/aclsampling ./pkg/ovs -count=1`
- `go test -race ./pkg/aclsampling ./pkg/ovs -run "Test(ParseSampleReference|ResolveNetworkPolicyACLSample)" -count=1`
- `go vet ./pkg/aclsampling ./pkg/ovs`
- `git diff --cached --check`

Local golangci-lint was intentionally not run because of its high memory usage; GitHub CI remains the lint gate.

PR #7155, opened 2026-08-07
feat: deliver ACL samples locally on nodes

zhangzujian:
Related to #7146.

## Summary

- reconcile an owned `Flow_Sample_Collector_Set` for `br-int` and remove only owned rows when ACL sampling is disabled
- reject unowned collector-set ID collisions without adopting or modifying external rows
- require the `local_group_id` schema, the Linux kernel `system` datapath, and runtime `psample=true` capability
- resolve runtime capability through `Open_vSwitch.datapaths` and the referenced `Datapath.capabilities` row
- keep the regular Vswitch client compatible with older schemas by using raw transactions for optional tables
- clear IPFIX from the owned collector set so the initial implementation remains local-delivery-only
- retry node reconciliation independently every minute and expose node availability and failure metrics

## Stack

1. #7149 - configuration
2. #7150 - stable metadata allocator
3. #7148 - OVN sampling object reconciler
4. #7151 - NetworkPolicy ACL attachment
5. #7152 - sampling-only retry and enforcement isolation
6. #7153 - ownership-aware disable and cleanup
7. #7154 - controller availability metrics
8. **this PR - node-local psample delivery**
9. #7156 - cookie and metadata event decoder

Base: `feature/acl-sampling-metrics`

## Verification

All local Go commands were executed in a systemd scope with `MemoryMax=2G`, `MemorySwapMax=0`, `GOMAXPROCS=2`, and package parallelism limited to one.

- `go test ./pkg/ovs ./pkg/daemon -count=1`
- targeted fake-OVSDB tests for lifecycle, IPFIX clearing, unowned conflicts, legacy schemas, userspace datapaths, and missing psample capability
- targeted `go test -race ./pkg/ovs ./pkg/daemon`
- `go vet ./pkg/ovs ./pkg/daemon`
- `git diff --check`

Local golangci-lint was intentionally not run because of its high memory usage; GitHub CI remains the lint gate.

PR #7154, opened 2026-08-07
feat: report ACL sampling availability

zhangzujian:
Related to #7146.

## Summary

- reconcile ACL sampling objects on controller startup without blocking controller availability
- trigger ownership-aware cleanup on startup when ACL sampling is disabled
- expose `acl_sampling_controller_available` for the latest reconciliation state
- count `reconcile`, `prepare`, and `attach` failures with `acl_sampling_controller_failures_total`
- restore the availability gauge after a successful ACL attachment retry

## Stack

1. #7149 - configuration
2. #7150 - stable metadata allocator
3. #7148 - OVN sampling object reconciler
4. #7151 - NetworkPolicy ACL attachment
5. #7152 - sampling-only retry and enforcement isolation
6. #7153 - ownership-aware disable and cleanup
7. **this PR - controller availability metrics**
8. #7155 - node-local psample delivery

Base: `feature/acl-sampling-cleanup`

## Verification

- `GOMAXPROCS=2 go test ./pkg/controller -count=1`
- targeted startup reconciliation and metrics unit tests
- targeted `go test -race`
- `GOMAXPROCS=2 go vet ./pkg/ovs ./pkg/controller`
- `git diff --check`

Local golangci-lint was intentionally not run because of its high memory usage; GitHub CI remains the lint gate.

PR #7153, opened 2026-08-07
feat: clean up owned ACL sampling state

zhangzujian:
Related to #7146.

## Summary

- make disabled controller reconciliation clear only ACLs marked as Kube-OVN NetworkPolicy sampling ACLs
- wait for non-root Sample rows to be garbage collected after ACL references are removed
- delete only owned collectors that are no longer referenced by any retained Sample
- preserve unowned applications, collectors, and ACL sample references
- delete owned applications last while preserving matching external applications that were reused
- treat an unsupported sampling schema as a no-op while the feature is disabled

## Stack

1. #7149 - configuration
2. #7150 - stable metadata allocator
3. #7148 - OVN sampling object reconciler
4. #7151 - NetworkPolicy ACL attachment
5. #7152 - sampling-only retry and enforcement isolation
6. **this PR - ownership-aware disable and cleanup**
7. #7154 - controller availability metrics

Base: `feature/acl-sampling-network-policy`

## Verification

- `GOMAXPROCS=2 go test ./pkg/ovs -count=1`
- targeted cleanup integration tests for unowned objects and externally retained owned collectors
- targeted `go test -race`
- `GOMAXPROCS=2 go vet ./pkg/ovs ./pkg/controller`

Local golangci-lint was intentionally not run because of its high memory usage; GitHub CI remains the lint gate.

PR #7157, opened 2026-08-07
feat: listen for local ACL samples

zhangzujian:
Related to #7146.

## Summary

- discover the Linux `psample` generic-netlink family and its `packets` multicast group at runtime
- subscribe with `NETLINK_ADD_MEMBERSHIP`, including multicast group IDs above the legacy 32-bit bind mask
- accept only kernel messages and filter events by the configured local psample group
- parse packet bytes, interface indexes, size, sequence, protocol, timestamp, and probability metadata
- decode the 8-byte big-endian OVS cookie into the OVN observation domain, application byte, datapath key, and ACL metadata
- stop blocking receives cleanly on context cancellation and provide an explicit unsupported implementation outside Linux

## Stack

1. #7149 - configuration
2. #7150 - stable metadata allocator
3. #7148 - OVN sampling object reconciler
4. #7151 - NetworkPolicy ACL attachment
5. #7152 - sampling-only retry and enforcement isolation
6. #7153 - ownership-aware disable and cleanup
7. #7154 - controller availability metrics
8. #7155 - node-local psample delivery
9. #7156 - cookie and metadata event decoder
10. **this PR - Linux psample listener**

Base: `feature/acl-sampling-event-decode`

## Validation

All Go commands ran in a transient systemd scope with `MemoryMax=2G`, `MemorySwapMax=0`, `GOMAXPROCS=2`, `GOMEMLIMIT=1700MiB`, and `GOFLAGS=-p=1`.

- `go test ./pkg/aclsampling ./pkg/ovs -count=1`
- targeted `go test -race ./pkg/aclsampling ./pkg/ovs`
- `go vet ./pkg/aclsampling ./pkg/ovs`
- `git diff --cached --check`

Local golangci-lint was intentionally not run because of its high memory usage; GitHub CI remains the lint gate.

Next: #7158 - ACL sample debug commands

### DIFF
--- a/pkg/aclsampling/event.go
+++ b/pkg/aclsampling/event.go
@@ -102,6 +102,10 @@ func ParseSampleReference(value string) (SampleReference, error) {
 	if err != nil {
 		return SampleReference{}, fmt.Errorf("invalid ACL sample cookie or metadata %q: %w", value, err)
 	}
+	return sampleReferenceFromUint64(raw)
+}
+
+func sampleReferenceFromUint64(raw uint64) (SampleReference, error) {
 	if raw == 0 {
 		return SampleReference{}, errors.New("ACL sample metadata must be greater than zero")
 	}
--- a/pkg/aclsampling/psample.go
+++ b/pkg/aclsampling/psample.go
@@ -1,0 +1,18 @@
+package aclsampling
+
+// PacketSample is an ACL sample received from the Linux psample generic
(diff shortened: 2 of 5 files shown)


zhangzujian:
Superseded by #7169, which uses an upstream head branch and is part of GitHub native stack #7175. The commit and review scope are unchanged.

PR #7182, opened 2026-08-08
controller, daemon: address pod event review findings

zhangzujian:
## What this PR does

Addresses follow-up review findings from #6989 and its release-1.16 backport #7180:

- propagate VIP lister and patch errors during pod network release, and report whether VIP state changed
- extract stale pod network state collection to keep reconciliation focused
- consolidate default and Multus pod QoS reconciliation
- scope QoS test dependencies to each controller instead of package-global mutable hooks
- split the oversized stale-port provider test into focused cases
- cover VIP lister failures, release failure Events, and retry retention

## Tests

- `go test ./pkg/controller ./pkg/daemon -count=1`
- `CGO_ENABLED=1 go test -race ./pkg/daemon -run "TestHandleUpdatePod|TestEnqueueUpdatePodOnlyQueuesRelevantAnnotationChanges" -count=1`
- `golangci-lint run ./pkg/controller/... ./pkg/daemon/...`
- `git diff --check`

### DIFF
--- a/pkg/controller/gc.go
+++ b/pkg/controller/gc.go
@@ -291,7 +291,7 @@ func (c *Controller) gcVip() error {
 			_, err := c.podsLister.Pods(namespace).Get(podName)
 			if err != nil {
 				if k8serrors.IsNotFound(err) {
-					if err = c.releaseVip(vip.Name); err != nil {
+					if _, err = c.releaseVip(vip.Name); err != nil {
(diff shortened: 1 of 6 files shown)


zhangzujian:
Superseded by three independent pull requests: #7185 for VIP release error propagation, #7184 for stale pod network state collection, and #7183 for Pod QoS reconciliation. Their combined Git tree is identical to this pull request.

PR #7220, opened 2026-08-12
fix: reconcile terminating OVN EIP/FIP/DNAT/SNAT from the add handler on restart

zbb88888:
OVN EIP/FIP/DNAT/SNAT deletion cleanup only runs in handleUpdateOvn*. On restart the informer re-lists existing objects and fires only AddFunc; with resync=0 a terminating object lands in the add queue, hits the already-ok short-circuit and never removes its finalizer. Route terminating objects to the update queue via a new enqueueUpdateIfTerminatingWithFinalizer helper that mirrors the OVN update handlers: terminating objects without a finalizer are skipped (cleanup done, awaiting apiserver deletion) to avoid a double delete. enqueueAddOvnEip keeps its requeueRouterLBRulesForEip side effect regardless.

test: verify OVN enqueue-add routing for terminating objects

Cover live->add, terminating+finalizer->update, terminating-without-finalizer->skip for OvnEip/OvnFip/OvnDnatRule/OvnSnatRule via a shared assertEnqueueAddRoutingWithFinalizer.

refactor: reuse terminating-route helper in OVN update handlers

The four enqueueUpdateOvn* deletion branches duplicated the terminating->update-queue logic now provided by enqueueUpdateIfTerminatingWithFinalizer. Reuse the helper to remove the duplication; the enqueue log unifies to V(3) "enqueue update to clean".

test: cover router LB rule requeue for a terminating OVN eip

enqueueAddOvnEip must still requeue an EIP's RouterLBRules when the EIP is terminating; the side effect runs before the terminating route returns.

# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -369,6 +369,21 @@ func enqueueUpdateIfTerminating(queue workqueue.TypedRateLimitingInterface[strin
 	return false
 }
 
+// enqueueUpdateIfTerminatingWithFinalizer behaves like enqueueUpdateIfTerminating but skips objects
+// whose finalizer has already been removed (their cleanup is done and they only await API server
+// deletion), mirroring the OVN enqueueUpdate* deletion branch to avoid a double delete. It returns
+// true when the object is terminating, so the add handler stops regardless of whether it enqueued.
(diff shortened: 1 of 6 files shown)


changluyi (approved):
Reviewed the terminating-resource add-path routing and its unit coverage. No blocking issues found; pkg/controller tests pass.

PR #7222, opened 2026-08-12
feat: add native Hardware VTEP binding support

emamihe:
## Summary
Kube-OVN can already provide L2/L3 networking for pods and KubeVirt VMs, but there was no first-class way to extend a Subnet onto a Hardware VTEP–capable physical switch so bare-metal servers on that switch join the same L2 domain.

Today that usually means out-of-band switch config plus a separately managed `ovn-controller-vtep`, with no CR, no controller ownership of the OVN VTEP LSP, and no Ready signal tied to real chassis binding. Applying only an OVN NB `type=vtep` port is also not enough: without VTEP `Logical_Switch` / `vlan_bindings` and `ovn-controller-vtep`, the SB Port_Binding never binds and bare-metal reachability never comes up.

This PR makes Hardware VTEP a native Kube-OVN feature:

- Add cluster-scoped `VtepBinding` CRD so a Kube-OVN Subnet can extend to a Hardware VTEP physical switch via an OVN NB `type=vtep` LSP.
- When `VTEP_DB_ADDR` is set, `kube-ovn-controller` writes Hardware VTEP `Logical_Switch` and `Physical_Port.vlan_bindings` (switch NMS still owns `Physical_Switch`/`Physical_Port`).
- Deploy optional `ovn-controller-vtep` (`ENABLE_HARDWARE_VTEP=true`) and mark `status.ready` only after SB `Port_Binding` is chassis-bound.
- Keep VTEP packages in the base image; add docs, unit/fake-VTEP coverage, and e2e for NB LSP creation.

Target path:

```text
VtepBinding CR → OVN NB type=vtep LSP → ovn-controller-vtep → Hardware VTEP OVSDB → physical switch port → bare-metal host
```

## How to enable
```bash
helm upgrade --install kube-ovn charts/kube-ovn -n kube-system \
  --set func.ENABLE_HARDWARE_VTEP=true \
  --set networking.VTEP_DB_ADDR='tcp:[SWITCH_IP]:6640'
```

Docs: `docs/hardware-vtep.md`

## Out of scope
- NX-API / vendor CLI switch config
- Creating Physical_Switch / Physical_Port rows
- Configuring IP addresses inside bare-metal servers
- BGP/EVPN/MLAG and bare-metal DHCP/IPAM

## Test plan
- [ ] Unit: VTEP Logical_Switch + vlan_bindings (in-memory OVSDB)
- [ ] Unit: SB Port_Binding chassis Ready helper
- [ ] Helm/values: `ENABLE_HARDWARE_VTEP` renders `ovn-controller-vtep`
- [ ] E2E `[group:vtep]`: `VtepBinding` creates NB `type=vtep` LSP; Ready stays false without chassis
- [ ] Rebuild base image and confirm `ovn-controller-vtep` is present
- [ ] Manual: with real/fake VTEP DB + chassis binding, `status.ready=true` and `status.chassis` populated


dist/images/Dockerfile.base
```diff
@@ -194,7 +194,7 @@ RUN mkdir /packages/ && \
     cp /usr/src/python3-openvswitch*deb /packages && \
     cp /usr/src/ovn-*deb /packages && \
     cp /usr/src/ovs/tutorial/ovs-sandbox /packages && \
-    cd /packages && rm -f *source* *doc* *datapath* *docker* *vtep* *test* *dev*
+    cd /packages && rm -f *source* *doc* *datapath* *docker* *test* *dev*
```
zhangzujian, line 197:
```txt
#76 1.849 dpkg: dependency problems prevent configuration of openvswitch-vtep:
#76 1.849  openvswitch-vtep depends on lsb-release; however:
#76 1.849   Package lsb-release is not installed.
```

pkg/controller/controller.go
```diff
@@ -686,6 +700,18 @@ func Run(ctx context.Context, config *Configuration) {
...
 	}
+	if config.VtepDbAddr != "" {
+		if controller.VtepClient, err = ovs.NewVtepClient(
+			config.VtepDbAddr,
+			config.OvsDbConnectTimeout,
+			config.OvsDbInactivityTimeout,
+			config.OvnTimeout,
+			config.OvsDbConnectMaxRetry,
+		); err != nil {
+			util.LogFatalAndExit(err, "failed to create hardware VTEP client")
```
zhangzujian: Could we avoid making the optional Hardware VTEP DB a fatal dependency of the main controller? If the switch or NMS is unreachable during startup, this exits `kube-ovn-controller` and stops reconciliation for unrelated networking resources. Please let the controller start, report the connection failure on the affected `VtepBinding` objects, and retry or reconnect in the background.

pkg/controller/vtep_binding.go
```diff
@@ -0,0 +1,322 @@
...
+			return nil
+		}
+		klog.Errorf("failed to reconcile vtep binding %s: %v", binding.Name, err)
+		if patchErr := c.patchVtepBindingStatusCondition(binding, "ReconcileFailed", err.Error()); patchErr != nil {
+			klog.Error(patchErr)
+		}
+		return err
+	}
+
+	if err = c.patchVtepBindingStatusCondition(binding, "VTEPAttachmentReady", ""); err != nil {
```
zhangzujian: This marks the binding Ready once, but after success no delayed requeue is scheduled and no SB `Port_Binding` watcher enqueues the CR. If the chassis binding disappears later, `status.ready` and `status.chassis` remain stale indefinitely. Please observe SB binding changes or schedule periodic revalidation even after the binding becomes Ready.

pkg/apis/kubeovn/v1/vtep-binding.go
```diff
@@ -0,0 +1,222 @@
...
+	// Maps to options:vtep-logical-switch. Defaults to subnet when empty. Immutable.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="spec.vtepLogicalSwitch is immutable"
+	VtepLogicalSwitch string `json:"vtepLogicalSwitch,omitempty"`
+
+	// PhysicalPort is the Hardware VTEP Physical_Port.name. When --vtep-db-addr
+	// is configured, Kube-OVN writes Physical_Port.vlan_bindings for this port.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	PhysicalPort string `json:"physicalPort"`
```
zhangzujian, line 104: `physicalPort` and `vlanID` remain mutable, but update reconciliation only ensures the new tuple. It never removes the previous `Physical_Port.vlan_bindings` entry, and deletion cleans only the current spec, so an update can leave the old VLAN forwarding into this Subnet permanently. Please either make both fields immutable or implement an explicit migration using previous or observed state, with update and deletion tests.

pkg/controller/vtep_binding.go
```diff
@@ -0,0 +1,322 @@
...
+func (c *Controller) validateVtepBindingConflict(binding *kubeovnv1.VtepBinding, vtepLogicalSwitch string) error {
+	bindings, err := c.vtepBindingsLister.List(labels.Everything())
+	if err != nil {
+		return fmt.Errorf("failed to list vtep bindings: %w", err)
+	}
+	for _, other := range bindings {
+		if other.Name == binding.Name || !other.DeletionTimestamp.IsZero() {
+			continue
+		}
+		if other.Spec.PhysicalSwitch == binding.Spec.PhysicalSwitch &&
```
zhangzujian: The conflict key is incomplete. `Physical_Port.vlan_bindings` is keyed by VLAN on a concrete physical port, so two CRs using the same `(physicalSwitch, physicalPort, vlanID)` but different logical switches pass this check and overwrite each other. Please reject that tuple as a conflict, and also make removal verify that the current mapping still references this binding's Logical Switch UUID before deleting it.

pkg/controller/vtep_binding.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+	return nil
+}
+
+func (c *Controller) handleDeleteVtepBinding(binding *kubeovnv1.VtepBinding) error {
+	c.vtepBindingKeyMutex.LockKey(binding.Name)
+	defer func() { _ = c.vtepBindingKeyMutex.UnlockKey(binding.Name) }()
+
+	klog.Infof("handle delete vtep binding %s", binding.Name)
+	return c.cleanupVtepBinding(binding)
```
zhangzujian: The finalizer path already calls `cleanupVtepBinding` before the object is deleted. `DeleteFunc` then enqueues the old object and performs the same cleanup again. If a same-name `VtepBinding` has been recreated and reconciled first, this stale delete can remove the new object's deterministic LSP and VLAN mapping. Please avoid the second cleanup or guard it with UID and ownership checks.

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
```diff
@@ -7065,3 +7065,183 @@ spec:
...
     subresources:
       status: {}
+---
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.20.1
+  name: vtep-bindings.kubeovn.io
```
zhangzujian, line 7098: Adding the CRD to `kube-ovn-v2` is only partial integration. The v2 Chart defines neither `ENABLE_HARDWARE_VTEP` nor `VTEP_DB_ADDR`, does not wire the DB address into `kube-ovn-controller`, and ships no `ovn-controller-vtep` workload. As a result, v2 users cannot complete chassis binding or reach `status.ready=true` through the supported Helm interface. Please add the full v2 values, controller wiring, workload, and tests, or explicitly exclude v2 rather than exposing a partially functional CRD.

zhangzujian (commented):
Review comments for findings 2-7 from the requested review.

pkg/controller/controller.go
emamihe: Done. Connecting to the Hardware VTEP DB is no longer fatal. If the switch/NMS is unreachable at startup, kube-ovn-controller still starts and keeps reconciling other resources. Affected VtepBinding objects report VTEPDBNotConnected, and the controller retries the VTEP client connection in the background.

pkg/controller/vtep_binding.go
emamihe: Done. After a binding becomes Ready, we schedule a delayed requeue (~30s) so chassis binding is revalidated. If the SB Port_Binding chassis disappears later, status.ready / status.chassis are cleared and the object is marked NotReady again.

pkg/apis/kubeovn/v1/vtep-binding.go
emamihe: Done. Made both physicalPort and vlanID immutable via CRD XValidation (self == oldSelf) and matching webhook checks, so updates cannot leave stale VLAN mappings. Changing the physical attachment requires recreating the CR.

pkg/controller/vtep_binding.go
emamihe: Done. Conflict detection now also rejects duplicate (physicalSwitch, physicalPort, vlanID) tuples (in addition to (physicalSwitch, vtepLogicalSwitch)). On remove, we only clear vlan_bindings[vlanID] when it still references this binding’s Logical Switch UUID.

pkg/controller/vtep_binding.go
emamihe: Done. Cleanup runs only through the finalizer path; DeleteFunc is a no-op. LSP deletion is further guarded by ownership checks (vtep-binding name and UID in external_ids) so a stale delete cannot remove a recreated binding’s LSP.

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
emamihe: Done. Completed v2 Helm support: features.enableHardwareVtep, networking.vtepDbAddr, VTEP_DB_ADDR on kube-ovn-controller, the ovn-controller-vtep Deployment, and Helm tests. Docs updated with v2 install examples.

charts/kube-ovn/templates/ovn-controller-vtep-deploy.yaml
```diff
@@ -0,0 +1,141 @@
...
+      maxUnavailable: 1
+    type: RollingUpdate
+  selector:
+    matchLabels:
+      app: ovn-controller-vtep
+  template:
+    metadata:
+      labels:
+        app: ovn-controller-vtep
+        component: network
```
zhangzujian, line 28: Please include this workload in the uninstall lifecycle. `cleanup.sh` never deletes `ovn-controller-vtep`, but later waits until every `component=network` pod is gone, so an enabled deployment can make cleanup wait forever. The same lifecycle should delete `VtepBinding` objects while the controller can run their finalizers, strip finalizers only as a last resort, and remove the CRD.

pkg/controller/vtep_binding.go
```diff
@@ -0,0 +1,418 @@
...
+		}
+	}
+	if err := c.handleDelVtepBindingFinalizer(binding); err != nil {
+		klog.Errorf("failed to remove finalizer for vtep binding %s: %v", binding.Name, err)
+		return err
+	}
+	return nil
+}
+
+func (c *Controller) reconcileVtepBinding(binding *kubeovnv1.VtepBinding) error {
```
zhangzujian, line 280: This controller creates deterministic NB LSPs plus Hardware VTEP `Logical_Switch`/`vlan_bindings`, but the GC pipeline has no VTEP pass. Force-deleted CRs or historical cleanup failures therefore leave forwarding state indefinitely. Please add ownership-aware GC for orphaned LSPs, Kube-OVN-owned unreferenced Logical Switch rows, and stale VLAN bindings.

pkg/controller/controller.go
```diff
@@ -454,6 +461,7 @@ func Run(ctx context.Context, config *Configuration) {
 	iptablesDnatRuleInformer := kubeovnInformerFactory.Kubeovn().V1().IptablesDnatRules()
 	iptablesSnatRuleInformer := kubeovnInformerFactory.Kubeovn().V1().IptablesSnatRules()
 	vlanInformer := kubeovnInformerFactory.Kubeovn().V1().Vlans()
+	vtepBindingInformer := kubeovnInformerFactory.Kubeovn().V1().VtepBindings()
```
zhangzujian, line 464: `ENABLE_HARDWARE_VTEP` is documented and defaulted as disabled, but this informer, its handler/cache sync, and the worker are unconditional. A CR can still drive the NB `type=vtep` path when the feature is disabled. Please gate the informer/worker/reconcile path, or define and document the actual gate semantics, so the default-disabled feature is genuinely inactive.

pkg/controller/controller.go
```diff
@@ -1430,6 +1462,7 @@ func (c *Controller) startWorkers(ctx context.Context) {
 	}
 	go wait.Until(runWorker("add/update ippool", c.addOrUpdateIPPoolQueue, c.handleAddOrUpdateIPPool), time.Second, ctx.Done())
 	go wait.Until(runWorker("add vlan", c.addVlanQueue, c.handleAddVlan), time.Second, ctx.Done())
+	go wait.Until(runWorker("add/update vtep binding", c.addOrUpdateVtepBindingQueue, c.handleAddOrUpdateVtepBinding), time.Second, ctx.Done())
```
zhangzujian: Could the VTEP setup and startup wiring be extracted into focused helpers? This addition further extends `Run` and `startWorkers`, which are already far beyond the repository's approximately 100-line function guideline, making the feature lifecycle and gating harder to audit.

pkg/controller/vtep_binding_test.go
```diff
@@ -0,0 +1,92 @@
...
+	binding.Spec.VtepLogicalSwitch = "custom-ls"
+	require.Equal(t, "custom-ls", binding.VtepLogicalSwitchName())
+}
+
+func TestGetVtepLogicalSwitchPortName(t *testing.T) {
+	t.Parallel()
+	require.Equal(t, "vtep.rack1-tenant-a", ovs.GetVtepLogicalSwitchPortName("rack1-tenant-a"))
+}
+
+func TestValidateVtepBindingConflict(t *testing.T) {
```
zhangzujian, line 49: These tests cover helpers and conflict validation, but not the controller lifecycle introduced here. Please add controller-level cases for finalizer add/delete, DB-disconnected deletion, reconnect, periodic Ready revalidation, same-name LSP ownership, and disabled-feature behavior.

pkg/controller/vtep_binding.go
```diff
@@ -0,0 +1,418 @@
...
+		return err
+	}
+	// Keep revalidating chassis binding so Ready/Chassis do not go stale.
+	c.addOrUpdateVtepBindingQueue.AddAfter(key, vtepBindingReadyRequeueDelay)
+	return nil
+}
+
+func (c *Controller) cleanupVtepBinding(binding *kubeovnv1.VtepBinding) error {
+	vtepLogicalSwitch := binding.VtepLogicalSwitchName()
+	if vtepClient := c.getVtepClient(); vtepClient != nil {
```
zhangzujian: When `VTEP_DB_ADDR` is configured but the client is nil, this block is skipped and the finalizer is still removed below. After a controller restart during a DB outage, deleting an existing binding permanently leaks its `Logical_Switch` and `vlan_bindings`. Please keep the CR terminating and retry cleanup after reconnect.

pkg/ovs/vtep-logical_switch.go
```diff
@@ -0,0 +1,122 @@
...
+	vtepOtherConfigBindingKey = VtepBindingKey
+)
+
+// EnsureLogicalSwitch creates the Hardware VTEP Logical_Switch if missing.
+func (c *VtepClient) EnsureLogicalSwitch(name, bindingName string) (*vtep.LogicalSwitch, error) {
+	ls, err := c.GetLogicalSwitch(name, true)
+	if err != nil {
+		return nil, err
+	}
+	if ls != nil {
```
zhangzujian, line 30: `Logical_Switch` names are global in the Hardware VTEP DB, and bindings on different physical switches may reuse the same name. Reusing the row without recording the additional owner makes deletion order-dependent: if the original owner is removed first, the last binding cannot delete the now-unreferenced row. Please track shared ownership or allow any Kube-OVN-owned, unreferenced row to be deleted safely.

pkg/controller/vtep_binding.go
```diff
@@ -0,0 +1,418 @@
...
+	newBinding := binding.DeepCopy()
+	controllerutil.AddFinalizer(newBinding, util.KubeOVNControllerFinalizer)
+	patch, err := util.GenerateMergePatchPayload(binding, newBinding)
+	if err != nil {
+		klog.Errorf("failed to generate patch payload for vtep binding %s: %v", binding.Name, err)
+		return err
+	}
+	if _, err = c.config.KubeOvnClient.KubeovnV1().VtepBindings().Patch(context.Background(), binding.Name,
+		types.MergePatchType, patch, metav1.PatchOptions{}, ""); err != nil {
+		if k8serrors.IsNotFound(err) {
```
zhangzujian, line 465: `NotFound` here means the CR disappeared before the finalizer was installed, but returning nil lets the caller continue into reconcile and create NB/VTEP state with no object or finalizer to own it. Please return a terminal object-gone result and stop reconciliation.

pkg/ovs/ovn-nb-logical_switch_port.go
```diff
@@ -201,6 +201,73 @@ func (c *OVNNbClient) CreateLocalnetLogicalSwitchPort(lsName, lspName, provider,
...
+	maps.Copy(finalExternalIDs, externalIDs)
+	finalExternalIDs[LogicalSwitchKey] = lsName
+	finalExternalIDs["vendor"] = util.CniTypeName
+
+	options := map[string]string{
+		"vtep-physical-switch": physicalSwitch,
+		"vtep-logical-switch":  vtepLogicalSwitch,
+	}
+
+	if lsp != nil {
```
zhangzujian, line 234: An existing same-name LSP is updated without checking binding name/UID ownership. It may be a foreign port or a stale port attached to another Logical Switch; this code overwrites it without moving `Logical_Switch.ports`, and cleanup later also treats missing owner markers as deletable. Please require exact ownership and transactionally move only an owned stale port to the target switch.

pkg/ovs/vtep.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	if port.VLANBindings == nil {
+		port.VLANBindings = make(map[int]string)
+	}
+	if existing, ok := port.VLANBindings[vlanID]; ok && existing == ls.UUID {
+		return nil
+	}
+	port.VLANBindings[vlanID] = ls.UUID
+	ops, err := c.Where(port).Update(port, &port.VLANBindings)
```
zhangzujian: `Where(port).Update(...VLANBindings)` replaces the entire cached map. Different bindings on the same physical port use different CR locks, so concurrent reconciles for different VLANs can write stale snapshots and erase or resurrect each other's entries. Please use OVSDB map mutations/CAS or serialize by physical port for both add and remove.

pkg/controller/vtep_binding.go
```diff
@@ -0,0 +1,418 @@
...
+		binding.Spec.PhysicalSwitch,
+		vtepLogicalSwitch,
+		externalIDs,
+	); err != nil {
+		return fmt.Errorf("failed to create vtep logical switch port %s: %w", lspName, err)
+	}
+
+	if c.config.VtepDbAddr != "" {
+		vtepClient := c.getVtepClient()
+		if vtepClient == nil {
```
zhangzujian: This returns before the SB `Port_Binding` lookup, so a controller restart during a VTEP DB management outage reports Ready=false even when the existing LSP is still chassis-bound and forwarding. Please derive Ready from the SB attachment and report DB reconciliation health in a separate condition.

pkg/controller/vtep_binding.go
```diff
@@ -0,0 +1,418 @@
...
+			return nil
+		}
+		klog.Errorf("failed to reconcile vtep binding %s: %v", binding.Name, err)
+		if patchErr := c.patchVtepBindingStatusCondition(binding, "ReconcileFailed", err.Error()); patchErr != nil {
+			klog.Error(patchErr)
+		}
+		return err
+	}
+
+	if err = c.patchVtepBindingStatusCondition(binding, "VTEPAttachmentReady", ""); err != nil {
```
zhangzujian: A successful reconcile sets Ready but never clears a previously set Error condition. After a transient failure the object can expose Ready=true and Error=true simultaneously. Please clear Error on successful or otherwise non-error reconciliation and add a recovery test.

zhangzujian (commented):
Follow-up review findings for the latest revision.

zhangzujian (commented):
Follow-up review findings for HEAD 0d50f8768.

The previous gate/GC/cleanup/Ready-vs-VTEPDBReady items look addressed: `--enable-hardware-vtep` now owns the handler/worker, uninstall deletes `ovn-controller-vtep` and `VtepBinding`, and Ready is derived from SB independently of VTEP DB health.

Remaining issues below are about the documented no-`VTEP_DB_ADDR` path, Ready semantics, `vlan_bindings` assignment, and error logging.

Nits (not blocking if you prefer to leave them):
- `ValidateVtepBinding` also requires the Subnet to already exist. That is stricter than the written spec; fine if intentional, but then please document it.
- Conflict checks are duplicated between the webhook and `validateVtepBindingConflict`.
- `EnsureVtepBinding` / `RemoveVtepBinding` still pass the same 5-tuple that `VtepLiveBinding` already models.

dist/images/start-ovn-controller-vtep.sh
```diff
@@ -0,0 +1,48 @@
...
+set -euo pipefail
+ENABLE_SSL=${ENABLE_SSL:-false}
+OVN_DB_IPS=${OVN_DB_IPS:-}
+OVN_SB_ADDR=${OVN_SB_ADDR:-}
+KUBE_OVN_SB_PORT=${KUBE_OVN_SB_PORT:-6642}
+VTEP_DB_ADDR=${VTEP_DB_ADDR:-}
+
+if [[ -z "${VTEP_DB_ADDR}" ]]; then
+  echo "VTEP_DB_ADDR must be set to the Hardware VTEP OVSDB remote (e.g. tcp:[switch-ip]:6640)"
+  exit 1
```
zhangzujian, line 11: `docs/hardware-vtep.md` says `ovn-controller-vtep` is deployed when `--enable-hardware-vtep` is on, and that the no-`VTEP_DB_ADDR` path still waits for SB chassis binding via that process. Helm/install follow that: the Deployment is rendered whenever the feature flag is true.

This script then exits if `VTEP_DB_ADDR` is empty, so the documented mode crashloops the Pod. Please either:

- require `VTEP_DB_ADDR` whenever Hardware VTEP is enabled and stop deploying a workload that cannot start, or
- allow the process to start without a VTEP DB and update the start script / values validation to match the doc.

pkg/ovs/ovn-sb-port_binding.go
```diff
@@ -0,0 +1,72 @@
...
+	return &list[0], nil
+}
+
+// IsPortBindingChassisBound reports whether the Port_Binding is chassis-bound
+// (and not explicitly down).
+func IsPortBindingChassisBound(pb *ovnsb.PortBinding) bool {
+	if pb == nil || pb.Chassis == nil || *pb.Chassis == "" {
+		return false
+	}
+	if pb.Up != nil && !*pb.Up {
```
zhangzujian: `status.ready` is documented as following SB chassis attachment only. This extra `Up == false` check keeps a chassis-bound VTEP LSP NotReady.

`ovn-controller-vtep` can leave `up=false` while `chassis` is already set. Please derive Ready from a non-empty chassis (or document `up=true` as an explicit extra Ready condition) and add a test for chassis-bound + `up=false`.

pkg/ovs/vtep.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+	if port.VLANBindings != nil {
+		if existing, ok := port.VLANBindings[vlanID]; ok && existing == ls.UUID {
+			return nil
+		}
+	}
+
+	ops, err := c.Where(port).Mutate(port, model.Mutation{
+		Field:   &port.VLANBindings,
+		Mutator: ovsdb.MutateOperationInsert,
```
zhangzujian, line 135: The documented write is `Physical_Port.vlan_bindings[vlanID] = Logical_Switch`. This only no-ops when the current value already equals `ls.UUID`; otherwise it `MutateOperationInsert`s.

If the key already maps to another UUID (stale leftover, or a recreated Logical_Switch), insert will not replace it, so `EnsureVtepBinding` cannot make the documented assignment and `VTEPDBReady` stays false. Please delete-then-insert, or mutate/update the existing key, when the stored UUID differs.

pkg/controller/vtep_binding.go
```diff
@@ -0,0 +1,488 @@
...
+		return
+	}
+	if err := vtepClient.EnsureVtepBinding(
+		binding.Spec.PhysicalSwitch,
+		binding.Spec.PhysicalPort,
+		vtepLogicalSwitch,
+		binding.Name,
+		binding.Spec.VlanID,
+	); err != nil {
+		binding.Status.NotVTEPDBReady("VTEPDBReconcileFailed", err.Error())
```
zhangzujian: Keeping Ready independent of VTEP DB health is fine, but this path still swallows the error: it writes `VTEPDBReconcileFailed` / `VTEPDBNotConnected` and returns without `klog.Error`.

Please log the unexpected VTEP DB / client errors here (and the nil-client case above). Status-only reporting is easy to miss in controller logs.

pkg/controller/vtep_binding.go
```diff
@@ -0,0 +1,488 @@
...
+	if !binding.DeletionTimestamp.IsZero() {
+		return c.cleanupVtepBinding(binding)
+	}
+
+	klog.Infof("handle add/update vtep binding %s", binding.Name)
+	binding.Status.EnsureStandardConditions()
+
+	if err = c.reconcileVtepBinding(binding); err != nil {
+		if errors.Is(err, errVtepBindingPending) {
+			binding.Status.NotReady("WaitingForChassis", err.Error())
```
zhangzujian: This is the common path (and the e2e case): NB LSP exists, Ready stays false until chassis binding. It only patches status, so there is no Event for `WaitingForChassis`.

The same gap exists after Ready flips back to false when the chassis disappears, and `reconcileVtepDBStatus` / `cleanupVtepBinding` never call `Eventf` for `VTEPDBNotConnected`, `VTEPDBReconcileFailed`, or cleanup success/failure.

Please emit Events for at least:

- `WaitingForChassis` and later chassis loss
- VTEP DB not connected / reconcile failed
- cleanup success and cleanup failure

Also give `VTEPAttachmentReady` a non-empty message (LSP / chassis). `patchVtepBindingStatusCondition` already records Warning/Normal for hard reconcile failures; pending and DB-health updates skip it.

zhangzujian (commented):
Follow-up: VtepBinding CR events are only partially wired.

`VTEPAttachmentReady` and `ReconcileFailed` exist, but the common NotReady / VTEP DB / cleanup transitions never emit Events, so `kubectl describe vtepbinding` will not explain why Ready is false. Details on the pending path below.

pkg/controller/gc.go
```diff
@@ -1329,6 +1330,69 @@ func (c *Controller) gcRouterLBRules() error {
 	return nil
 }
 
+func (c *Controller) gcVtepBinding() error {
+	if !c.hardwareVtepEnabled() || c.vtepBindingsLister == nil {
```
zhangzujian: CODE_STYLE CRD #4: `gc.go` should clean inconsistent resources even when the feature is disabled.

This returns immediately when `--enable-hardware-vtep` is off or the lister is nil, so leftover `type=vtep` LSPs and Hardware VTEP `vlan_bindings` are never collected after the flag is turned off. `gcNetworkPolicy` still sweeps leftovers when NP is disabled.

Please always list Kube-OVN-owned VTEP LSPs (and VTEP DB state if a client is connected). If the feature is off, treat the live set as empty and delete those orphans; if it is on, keep using live CRs.

test/e2e/kube-ovn/vtep/vtep.go
```diff
@@ -0,0 +1,102 @@
...
+				PhysicalSwitch: "fake-vtep-switch",
+				PhysicalPort:   "Ethernet1/1",
+				VlanID:         100,
+			},
+		}
+		binding = bindingClient.Create(binding)
+
+		lspName := ovs.GetVtepLogicalSwitchPortName(bindingName)
+		ginkgo.By("Waiting for status.logicalSwitchPort=" + lspName)
+		binding = bindingClient.WaitUntil(bindingName, func(b *apiv1.VtepBinding) (bool, error) {
```
zhangzujian, line 72: This waits on `status.logicalSwitchPort`, but Hardware VTEP is default-off and `install.sh` / Helm refuse to enable it without `VTEP_DB_ADDR`. The default kind/install e2e job never starts the worker, so this `WaitUntil` hangs until timeout.

Please skip unless `--enable-hardware-vtep` is actually on, or give the test job a path that enables NB VTEP LSP reconciliation without requiring a real VTEP DB / `ovn-controller-vtep`.

pkg/ovs/ovn-sb-port_binding.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if logicalPort == "" {
+		return nil, errors.New("logical port name is empty")
+	}
+
+	var list []ovnsb.PortBinding
+	if err := c.WhereCache(func(pb *ovnsb.PortBinding) bool {
+		return pb.LogicalPort == logicalPort
+	}).List(ctx, &list); err != nil {
+		return nil, fmt.Errorf("list port binding for logical port %s: %w", logicalPort, err)
```
zhangzujian, line 30: CODE_STYLE error handling: unexpected errors should be logged, not only wrapped.

Sibling SB helpers in `ovn-sb-chassis.go` call `klog.Error(err)` (or a contextual `klog.Errorf`) before returning. Please do the same here and in `GetChassisNameByUUID`.

pkg/webhook/vtep_binding.go
```diff
@@ -0,0 +1,93 @@
...
+		if binding.Spec.PhysicalPort != oldBinding.Spec.PhysicalPort {
+			return errors.New("physicalPort is immutable")
+		}
+		if binding.Spec.VlanID != oldBinding.Spec.VlanID {
+			return errors.New("vlanID is immutable")
+		}
+	}
+
+	subnet := &ovnv1.Subnet{}
+	if err := v.cache.Get(ctx, client.ObjectKey{Name: binding.Spec.Subnet}, subnet); err != nil {
```
zhangzujian, line 79: The docs say the validating webhook rejects a `VtepBinding` whose Subnet does not exist, and that `(physicalSwitch, vtepLogicalSwitch)` / `(physicalSwitch, physicalPort, vlanID)` must be unique.

This `cache.Get` / conflict list implements that, but the tests only cover empty fields and immutability. Please add cases for a missing Subnet and for both conflict keys.

Also, Helm v1 and `install.sh` still do not deploy the validating webhook (only the v2 chart / `yamls/webhook.yaml` do). If webhook-less installs are supported, the controller path remains the only admission, and that should be explicit in the docs.

zhangzujian (commented):
Follow-up review for HEAD 2614827b0.

The previous gate items look addressed: Helm/install require `VTEP_DB_ADDR` when Hardware VTEP is enabled, Ready is chassis-only (`Port_Binding.up=false` does not block), stale `vlan_bindings` are replaced, VTEP DB errors are logged, and pending/cleanup Events are emitted.

One issue is not in the new hunks, so it is here instead of an inline comment:

`gcChassis` still deletes every SB Chassis that is not tied to a Kubernetes Node annotation. Hardware VTEP chassis (`Chassis.vtep_logical_switches`) are not exempt. Periodic GC can remove the chassis that `status.ready` depends on, then the 30s requeue marks the binding NotReady. Please skip VTEP chassis (or otherwise keep them) and add a test.

Inline comments cover GC after the feature is turned off, the default-off e2e, SB error logging, and webhook coverage.

emamihe:
The review items from 2614827 are addressed.

Chassis GC. gcChassis now skips SB chassis with a non-empty vtep_logical_switches column, so periodic node GC cannot delete the chassis that status.ready depends on. Covered by TestGcChassisSkipsHardwareVTEPChassis.

VTEP GC when the feature is off. gcVtepBinding always lists Kube-OVN-owned type=vtep LSPs (and VTEP DB state if a client is connected). With --enable-hardware-vtep off, the live set is empty and leftovers are deleted, including LSPs that still match leftover CRs. Covered by TestGcVtepBindingDeletesOrphansWhenDisabled.

Default-off e2e. The VTEP e2e skips unless kube-ovn-controller has ENABLE_HARDWARE_VTEP=true or --enable-hardware-vtep, so the default kind/install job no longer waits on status.logicalSwitchPort.

SB errors. GetPortBindingByLogicalPort and GetChassisNameByUUID log unexpected errors with klog.Error before returning, matching other SB helpers.

Webhook coverage and docs. Tests cover a missing Subnet and both conflict keys. Docs state that Helm v1 and install.sh do not deploy the validating webhook, and that the controller is the admission path on those installs.

zhangzujian (changes requested):
Two blocking findings remain on HEAD `7ca80b7c9`:

1. **Generated CRDs are out of sync.** `make verify-crd` fails in the arm64 build because `hack/gen-crd.sh` changes `charts/kube-ovn-v2/crds/kube-ovn-crd.yaml`, `charts/kube-ovn/templates/kube-ovn-crd.yaml`, and `dist/images/install.sh`. Please regenerate and commit the generated artifacts before merge. Evidence: https://github.com/kubeovn/kube-ovn/actions/runs/32337927470/job/96474140919

2. **Old finalizer cleanup can remove a replacement binding's VLAN mapping.** `VtepBindingConflict` skips objects with `DeletionTimestamp`, so a new binding with a different name can claim the same `(physicalSwitch, physicalPort, vlanID, vtepLogicalSwitch)` while the old object is terminating. The old cleanup then calls `RemoveVtepBinding` (`pkg/controller/vtep_binding.go:236-253`), which clears the mapping by Logical_Switch UUID (`pkg/ovs/vtep.go:170-193`) and cannot distinguish the replacement CR. This can remove the live binding's `vlan_bindings` entry and/or logical switch. Keep the tuple reserved until finalizer cleanup completes, or make cleanup verify the owning binding UID/live binding before deleting shared state.

Please address both and add regression coverage.

zhangzujian:
--- FAIL: TestOvnClientTestSuite (4.20s)
    --- FAIL: TestOvnClientTestSuite/Test_SharedLogicalSwitchDeleteAndGC (0.00s)
        panic.go:336: test panicked: runtime error: invalid memory address or nil pointer dereference
            goroutine 511 [running]:
            runtime/debug.Stack()
            	/opt/hostedtoolcache/go/1.26.6/x64/src/runtime/debug/stack.go:26 +0x5e
            github.com/stretchr/testify/suite.failOnPanic(0x5c79a4fb448, {0x1a8e140, 0x2f84680})
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:93 +0x37
            github.com/stretchr/testify/suite.Run.func2.1()
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:200 +0x259
            panic({0x1a8e140?, 0x2f84680?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/runtime/panic.go:860 +0x13a
            github.com/kubeovn/kube-ovn/pkg/ovs.(*VtepClient).EnsureVtepBinding(0x5c799896060, {0x1d1a8e1, 0xf}, {0x1d0cc60, 0xb}, {0x1d06c56, 0x9}, {0x1d06c5f?, 0x0?}, 0x78)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep.go:97 +0x276
            github.com/kubeovn/kube-ovn/pkg/ovs.(*OvnClientTestSuite).testSharedLogicalSwitchDeleteAndGC(0x5c7993f6180)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep_test.go:315 +0x76f
            github.com/kubeovn/kube-ovn/pkg/ovs.(*OvnClientTestSuite).Test_SharedLogicalSwitchDeleteAndGC(0x0?)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep_test.go:277 +0x13
            reflect.Value.call({0x5c7993f3a80?, 0x5c7994050a0?, 0x16cc9b7?}, {0x1cf78b1, 0x4}, {0x5c79a37ff28, 0x1, 0x1aa11e0?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/reflect/value.go:586 +0xf0c
            reflect.Value.Call({0x5c7993f3a80?, 0x5c7994050a0?, 0x2f0e4c0?}, {0x5c79a37ff28?, 0x1b87ac0?, 0x1?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/reflect/value.go:369 +0xb9
            github.com/stretchr/testify/suite.Run.func2(0x5c79a4fb448)
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:212 +0x479
            testing.tRunner(0x5c79a4fb448, 0x5c799348240)
            	/opt/hostedtoolcache/go/1.26.6/x64/src/testing/testing.go:2036 +0xea
            created by testing.(*T).Run in goroutine 27
            	/opt/hostedtoolcache/go/1.26.6/x64/src/testing/testing.go:2101 +0x4c5
    --- FAIL: TestOvnClientTestSuite/Test_RemoveVtepBindingSkipsReplacementSameLogicalSwitch (0.00s)
        panic.go:336: test panicked: runtime error: invalid memory address or nil pointer dereference
            goroutine 404 [running]:
            runtime/debug.Stack()
            	/opt/hostedtoolcache/go/1.26.6/x64/src/runtime/debug/stack.go:26 +0x5e
            github.com/stretchr/testify/suite.failOnPanic(0x5c79a5b1d48, {0x1a8e140, 0x2f84680})
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:93 +0x37
            github.com/stretchr/testify/suite.Run.func2.1()
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:200 +0x259
            panic({0x1a8e140?, 0x2f84680?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/runtime/panic.go:860 +0x13a
            github.com/kubeovn/kube-ovn/pkg/ovs.(*VtepClient).EnsureVtepBinding(0x5c799896060, {0x1d1d112, 0x10}, {0x1d0cc3f, 0xb}, {0x1d1f79d, 0x11}, {0x1d0cc4a?, 0x0?}, 0xa0)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep.go:97 +0x276
            github.com/kubeovn/kube-ovn/pkg/ovs.(*OvnClientTestSuite).testRemoveVtepBindingSkipsReplacementSameLogicalSwitch(0x5c7993f6180)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep_test.go:203 +0x76f
            github.com/kubeovn/kube-ovn/pkg/ovs.(*OvnClientTestSuite).Test_RemoveVtepBindingSkipsReplacementSameLogicalSwitch(0x0?)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep_test.go:165 +0x13
            reflect.Value.call({0x5c7993f3a80?, 0x5c799404fa8?, 0x13?}, {0x1cf78b1, 0x4}, {0x5c79a373f28, 0x1, 0x1?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/reflect/value.go:586 +0xf0c
            reflect.Value.Call({0x5c7993f3a80?, 0x5c799404fa8?, 0x2f0e4c0?}, {0x5c798eadf28?, 0x1b87ac0?, 0x1?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/reflect/value.go:369 +0xb9
            github.com/stretchr/testify/suite.Run.func2(0x5c79a5b1d48)
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:212 +0x479
            testing.tRunner(0x5c79a5b1d48, 0x5c799346fc0)
            	/opt/hostedtoolcache/go/1.26.6/x64/src/testing/testing.go:2036 +0xea
            created by testing.(*T).Run in goroutine 27
            	/opt/hostedtoolcache/go/1.26.6/x64/src/testing/testing.go:2101 +0x4c5
    --- FAIL: TestOvnClientTestSuite/Test_RemoveVtepBindingSkipsForeignMapping (0.00s)
        panic.go:336: test panicked: runtime error: invalid memory address or nil pointer dereference
            goroutine 403 [running]:
            runtime/debug.Stack()
            	/opt/hostedtoolcache/go/1.26.6/x64/src/runtime/debug/stack.go:26 +0x5e
            github.com/stretchr/testify/suite.failOnPanic(0x5c79a5b1b08, {0x1a8e140, 0x2f84680})
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:93 +0x37
            github.com/stretchr/testify/suite.Run.func2.1()
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:200 +0x259
            panic({0x1a8e140?, 0x2f84680?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/runtime/panic.go:860 +0x13a
            github.com/kubeovn/kube-ovn/pkg/ovs.(*VtepClient).EnsureVtepBinding(0x5c799896060, {0x1d13458, 0xd}, {0x1d0cc55, 0xb}, {0x1d0424c, 0x8}, {0x1d13465?, 0x0?}, 0xc8)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep.go:97 +0x276
            github.com/kubeovn/kube-ovn/pkg/ovs.(*OvnClientTestSuite).testRemoveVtepBindingSkipsForeignMapping(0x5c7993f6180)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep_test.go:255 +0x77e
            github.com/kubeovn/kube-ovn/pkg/ovs.(*OvnClientTestSuite).Test_RemoveVtepBindingSkipsForeignMapping(0x0?)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep_test.go:161 +0x13
            reflect.Value.call({0x5c7993f3a80?, 0x5c799404fa0?, 0x13?}, {0x1cf78b1, 0x4}, {0x5c79a372f28, 0x1, 0x1?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/reflect/value.go:586 +0xf0c
            reflect.Value.Call({0x5c7993f3a80?, 0x5c799404fa0?, 0x2f0e4c0?}, {0x5c798eadf28?, 0x1b87ac0?, 0x1?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/reflect/value.go:369 +0xb9
            github.com/stretchr/testify/suite.Run.func2(0x5c79a5b1b08)
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:212 +0x479
            testing.tRunner(0x5c79a5b1b08, 0x5c799346f30)
            	/opt/hostedtoolcache/go/1.26.6/x64/src/testing/testing.go:2036 +0xea
            created by testing.(*T).Run in goroutine 27
            	/opt/hostedtoolcache/go/1.26.6/x64/src/testing/testing.go:2101 +0x4c5
    --- FAIL: TestOvnClientTestSuite/Test_GCOrphanedVtepState (0.00s)
        panic.go:336: test panicked: runtime error: invalid memory address or nil pointer dereference
            goroutine 177 [running]:
            runtime/debug.Stack()
            	/opt/hostedtoolcache/go/1.26.6/x64/src/runtime/debug/stack.go:26 +0x5e
            github.com/stretchr/testify/suite.failOnPanic(0x5c79977e6c8, {0x1a8e140, 0x2f84680})
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:93 +0x37
            github.com/stretchr/testify/suite.Run.func2.1()
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:200 +0x259
            panic({0x1a8e140?, 0x2f84680?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/runtime/panic.go:860 +0x13a
            github.com/kubeovn/kube-ovn/pkg/ovs.(*VtepClient).EnsureVtepBinding(0x5c799896060, {0x1d0cc6b, 0xb}, {0x1d0cc76, 0xb}, {0x1cf909b, 0x5}, {0x1d09aa7?, 0x0?}, 0x82)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep.go:97 +0x276
            github.com/kubeovn/kube-ovn/pkg/ovs.(*OvnClientTestSuite).testGCOrphanedVtepState(0x5c7993f6180)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep_test.go:384 +0x770
            github.com/kubeovn/kube-ovn/pkg/ovs.(*OvnClientTestSuite).Test_GCOrphanedVtepState(0x0?)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep_test.go:346 +0x13
            reflect.Value.call({0x5c7993f3a80?, 0x5c799404bb8?, 0x13?}, {0x1cf78b1, 0x4}, {0x5c7999c3f28, 0x1, 0x1?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/reflect/value.go:586 +0xf0c
            reflect.Value.Call({0x5c7993f3a80?, 0x5c799404bb8?, 0x2f0e4c0?}, {0x5c798eac728?, 0x1b87ac0?, 0x5c798eac601?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/reflect/value.go:369 +0xb9
            github.com/stretchr/testify/suite.Run.func2(0x5c79977e6c8)
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:212 +0x479
            testing.tRunner(0x5c79977e6c8, 0x5c7993406c0)
            	/opt/hostedtoolcache/go/1.26.6/x64/src/testing/testing.go:2036 +0xea
            created by testing.(*T).Run in goroutine 27
            	/opt/hostedtoolcache/go/1.26.6/x64/src/testing/testing.go:2101 +0x4c5
    --- FAIL: TestOvnClientTestSuite/Test_EnsureVtepBinding (0.01s)
        panic.go:336: test panicked: runtime error: invalid memory address or nil pointer dereference
            goroutine 173 [running]:
            runtime/debug.Stack()
            	/opt/hostedtoolcache/go/1.26.6/x64/src/runtime/debug/stack.go:26 +0x5e
            github.com/stretchr/testify/suite.failOnPanic(0x5c799699d48, {0x1a8e140, 0x2f84680})
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:93 +0x37
            github.com/stretchr/testify/suite.Run.func2.1()
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:200 +0x259
            panic({0x1a8e140?, 0x2f84680?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/runtime/panic.go:860 +0x13a
            github.com/kubeovn/kube-ovn/pkg/ovs.(*VtepClient).EnsureVtepBinding(0x5c799896060, {0x1d0423c, 0x8}, {0x1d0cc1e, 0xb}, {0x1d10072, 0xc}, {0x1d09a89?, 0x0?}, 0x78)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep.go:97 +0x276
            github.com/kubeovn/kube-ovn/pkg/ovs.(*OvnClientTestSuite).testEnsureVtepBinding(0x5c7993f6180)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep_test.go:82 +0x772
            github.com/kubeovn/kube-ovn/pkg/ovs.(*OvnClientTestSuite).Test_EnsureVtepBinding(0x0?)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep_test.go:13 +0x13
            reflect.Value.call({0x5c7993f3a80?, 0x5c799404b98?, 0x13?}, {0x1cf78b1, 0x4}, {0x5c79999df28, 0x1, 0x1?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/reflect/value.go:586 +0xf0c
            reflect.Value.Call({0x5c7993f3a80?, 0x5c799404b98?, 0x2f0e4c0?}, {0x5c798eac728?, 0x1b87ac0?, 0x5c798eac601?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/reflect/value.go:369 +0xb9
            github.com/stretchr/testify/suite.Run.func2(0x5c799699d48)
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:212 +0x479
            testing.tRunner(0x5c799699d48, 0x5c799340480)
            	/opt/hostedtoolcache/go/1.26.6/x64/src/testing/testing.go:2036 +0xea
            created by testing.(*T).Run in goroutine 27
            	/opt/hostedtoolcache/go/1.26.6/x64/src/testing/testing.go:2101 +0x4c5
    --- FAIL: TestOvnClientTestSuite/Test_EnsureVtepBindingReplacesStaleVLANMapping (0.01s)
        panic.go:336: test panicked: runtime error: invalid memory address or nil pointer dereference
            goroutine 174 [running]:
            runtime/debug.Stack()
            	/opt/hostedtoolcache/go/1.26.6/x64/src/runtime/debug/stack.go:26 +0x5e
            github.com/stretchr/testify/suite.failOnPanic(0x5c79977e008, {0x1a8e140, 0x2f84680})
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:93 +0x37
            github.com/stretchr/testify/suite.Run.func2.1()
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:200 +0x259
            panic({0x1a8e140?, 0x2f84680?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/runtime/panic.go:860 +0x13a
            github.com/kubeovn/kube-ovn/pkg/ovs.(*VtepClient).EnsureVtepBinding(0x5c799896060, {0x1d1a8b4, 0xf}, {0x1d0cc29, 0xb}, {0x1d04244, 0x8}, {0x1d1344b?, 0x0?}, 0x8c)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep.go:97 +0x276
            github.com/kubeovn/kube-ovn/pkg/ovs.(*OvnClientTestSuite).testEnsureVtepBindingReplacesStaleVLANMapping(0x5c7993f6180)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep_test.go:150 +0x772
            github.com/kubeovn/kube-ovn/pkg/ovs.(*OvnClientTestSuite).Test_EnsureVtepBindingReplacesStaleVLANMapping(0x0?)
            	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/vtep_test.go:111 +0x13
            reflect.Value.call({0x5c7993f3a80?, 0x5c799404ba0?, 0x13?}, {0x1cf78b1, 0x4}, {0x5c79999ef28, 0x1, 0x1?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/reflect/value.go:586 +0xf0c
            reflect.Value.Call({0x5c7993f3a80?, 0x5c799404ba0?, 0x2f0e4c0?}, {0x5c798eac728?, 0x1b87ac0?, 0x5c798eac601?})
            	/opt/hostedtoolcache/go/1.26.6/x64/src/reflect/value.go:369 +0xb9
            github.com/stretchr/testify/suite.Run.func2(0x5c79977e008)
            	/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.12.1/suite/suite.go:212 +0x479
            testing.tRunner(0x5c79977e008, 0x5c799340510)
            	/opt/hostedtoolcache/go/1.26.6/x64/src/testing/testing.go:2036 +0xea
            created by testing.(*T).Run in goroutine 27
            	/opt/hostedtoolcache/go/1.26.6/x64/src/testing/testing.go:2101 +0x4c5

emamihe:
Today I learned that as of OVN version 26.03 (released on March 2026) there is a much more modern approach.
OVN logical switches can now directly participate in a VXLAN/EVPN fabric. OVN exposes MAC+IP/VTEP state through the Linux networking stack while FRR handles BGP EVPN Type 2 advertisements and convergence with the physical fabric.
This means we can extend an OVN logical L2 domain directly into an EVPN fabric without relying on the legacy OVSDB VTEP approach or requiring switches with OVSDB VTEP support.
Pretty much the architecture I was looking for.

PR #7237, opened 2026-08-14
feat(controller): support dynamic nat gateway lan IP

zbb88888:
# Pull Request

## What type of this PR

- Features
- Bug fixes
- Docs
- Tests

Fixes #(issue-number)

### DIFF
--- a/charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
+++ b/charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
@@ -4649,7 +4649,7 @@ spec:
     schema:
       openAPIV3Schema:
         description: |-
-          VpcNatGateway represents a NAT gateway for a VPC, implemented as a StatefulSet Pod.
+          VpcNatGateway represents a NAT gateway for a VPC, implemented by a StatefulSet in non-HA mode or a Deployment in HA mode.
 
           Architecture note:
           The NAT gateway Pod does NOT support hot updates. Any changes to Spec fields (ExternalSubnets,
@@ -4659,7 +4659,9 @@ spec:
            2. Runtime state (vpc_cidrs, init status) is managed by separate handlers and will be
               automatically restored after Pod recreation through the normal reconciliation flow
 
-          The only exception is QoSPolicy, which can be updated without Pod restart.
+          QoSPolicy updates are applied without restarting the Pod. Every other Spec change,
+          including the one-time persistence of a dynamically allocated LanIP, updates the
+          workload template and therefore recreates the Pod.
         properties:
           apiVersion:
             description: |-
@@ -5699,8 +5701,12 @@ spec:
                 type: array
               lanIp:
                 description: |-
-                  LAN IP address for the NAT gateway. This field is immutable after creation.
-                  Used only when Replicas = 1 (non-HA mode).
+                  LAN IP address for the NAT gateway. Used only when Replicas = 1 (non-HA mode).
+                  If empty, an address is allocated dynamically and persisted here after it is observed on the Pod.
+                  Once set, this field is immutable.
+                  In HA mode (Replicas > 1) this field is ignored and cannot be used to pin addresses: every
+                  replica allocates its own address dynamically, and the controller re-points the OVN policy
+                  routes and BFD sessions whenever a replica is replaced.
                 type: string
               namespace:
                 description: |-
@@ -6718,9 +6724,9 @@ spec:
                 type: array
               lanIp:
                 description: |-
-                  LAN IP address(es) for the NAT gateway.
-                  For non-HA, this is the single LanIP from spec.
-                  For HA, this is a comma-separated list of all IPs within the NAT gateway pods.
+                  LAN IP address(es) observed on the NAT gateway Pods.
+                  For non-HA, this is the single address persisted to spec.lanIp.
+                  For HA, this is a comma-separated, sorted list of all observed addresses.
                 type: string
               qosPolicy:
                 description: QoS policy applied to the NAT gateway
--- a/charts/kube-ovn-v2/templates/rbac/ovn-CR.yaml
+++ b/charts/kube-ovn-v2/templates/rbac/ovn-CR.yaml
@@ -13,6 +13,7 @@ rules:
       - vpcs/status
       - vpc-nat-gateways
       - vpc-nat-gateways/status
+      - vpc-nat-gateways/finalizers
       - vpc-egress-gateways
       - vpc-egress-gateways/status
       - subnets
@@ -155,6 +156,7 @@ rules:
       - watch
       - create
       - update
+      - patch
       - delete
   - apiGroups:
       - ""
@@ -202,9 +204,11 @@ rules:
     verbs:
       - get
       - list
+      - watch
       - create
       - delete
       - update
+      - patch
   - apiGroups:
       - ""
     resources:
--- a/charts/kube-ovn/templates/kube-ovn-crd.yaml
+++ b/charts/kube-ovn/templates/kube-ovn-crd.yaml
@@ -4674,7 +4674,7 @@ spec:
     schema:
       openAPIV3Schema:
         description: |-
-          VpcNatGateway represents a NAT gateway for a VPC, implemented as a StatefulSet Pod.
+          VpcNatGateway represents a NAT gateway for a VPC, implemented by a StatefulSet in non-HA mode or a Deployment in HA mode.
 
           Architecture note:
           The NAT gateway Pod does NOT support hot updates. Any changes to Spec fields (ExternalSubnets,
@@ -4684,7 +4684,9 @@ spec:
            2. Runtime state (vpc_cidrs, init status) is managed by separate handlers and will be
               automatically restored after Pod recreation through the normal reconciliation flow
 
-          The only exception is QoSPolicy, which can be updated without Pod restart.
+          QoSPolicy updates are applied without restarting the Pod. Every other Spec change,
+          including the one-time persistence of a dynamically allocated LanIP, updates the
+          workload template and therefore recreates the Pod.
         properties:
           apiVersion:
             description: |-
@@ -5724,8 +5726,12 @@ spec:
                 type: array
               lanIp:
                 description: |-
-                  LAN IP address for the NAT gateway. This field is immutable after creation.
-                  Used only when Replicas = 1 (non-HA mode).
+                  LAN IP address for the NAT gateway. Used only when Replicas = 1 (non-HA mode).
+                  If empty, an address is allocated dynamically and persisted here after it is observed on the Pod.
+                  Once set, this field is immutable.
+                  In HA mode (Replicas > 1) this field is ignored and cannot be used to pin addresses: every
+                  replica allocates its own address dynamically, and the controller re-points the OVN policy
+                  routes and BFD sessions whenever a replica is replaced.
                 type: string
               namespace:
                 description: |-
@@ -6743,9 +6749,9 @@ spec:
                 type: array
               lanIp:
                 description: |-
-                  LAN IP address(es) for the NAT gateway.
-                  For non-HA, this is the single LanIP from spec.
-                  For HA, this is a comma-separated list of all IPs within the NAT gateway pods.
+                  LAN IP address(es) observed on the NAT gateway Pods.
+                  For non-HA, this is the single address persisted to spec.lanIp.
+                  For HA, this is a comma-separated, sorted list of all observed addresses.
                 type: string
               qosPolicy:
                 description: QoS policy applied to the NAT gateway
--- a/charts/kube-ovn/templates/ovn-CR.yaml
+++ b/charts/kube-ovn/templates/ovn-CR.yaml
@@ -21,6 +21,7 @@ rules:
       - vpcs/status
       - vpc-nat-gateways
       - vpc-nat-gateways/status
+      - vpc-nat-gateways/finalizers
       - vpc-egress-gateways
       - vpc-egress-gateways/status
       - subnets
@@ -162,6 +163,7 @@ rules:
       - watch
       - create
       - update
+      - patch
       - delete
   - apiGroups:
       - ""
@@ -209,9 +211,11 @@ rules:
     verbs:
       - get
       - list
+      - watch
       - create
       - delete
       - update
+      - patch
   - apiGroups:
       - ""
     resources:
--- a/dist/images/install.sh
+++ b/dist/images/install.sh
@@ -5074,7 +5074,7 @@ spec:
     schema:
       openAPIV3Schema:
         description: |-
-          VpcNatGateway represents a NAT gateway for a VPC, implemented as a StatefulSet Pod.
+          VpcNatGateway represents a NAT gateway for a VPC, implemented by a StatefulSet in non-HA mode or a Deployment in HA mode.
 
           Architecture note:
           The NAT gateway Pod does NOT support hot updates. Any changes to Spec fields (ExternalSubnets,
@@ -5084,7 +5084,9 @@ spec:
            2. Runtime state (vpc_cidrs, init status) is managed by separate handlers and will be
               automatically restored after Pod recreation through the normal reconciliation flow
 
-          The only exception is QoSPolicy, which can be updated without Pod restart.
+          QoSPolicy updates are applied without restarting the Pod. Every other Spec change,
+          including the one-time persistence of a dynamically allocated LanIP, updates the
+          workload template and therefore recreates the Pod.
         properties:
           apiVersion:
             description: |-
@@ -6124,8 +6126,12 @@ spec:
                 type: array
               lanIp:
                 description: |-
-                  LAN IP address for the NAT gateway. This field is immutable after creation.
-                  Used only when Replicas = 1 (non-HA mode).
+                  LAN IP address for the NAT gateway. Used only when Replicas = 1 (non-HA mode).
+                  If empty, an address is allocated dynamically and persisted here after it is observed on the Pod.
+                  Once set, this field is immutable.
+                  In HA mode (Replicas > 1) this field is ignored and cannot be used to pin addresses: every
+                  replica allocates its own address dynamically, and the controller re-points the OVN policy
+                  routes and BFD sessions whenever a replica is replaced.
                 type: string
               namespace:
                 description: |-
@@ -7143,9 +7149,9 @@ spec:
                 type: array
               lanIp:
                 description: |-
-                  LAN IP address(es) for the NAT gateway.
-                  For non-HA, this is the single LanIP from spec.
-                  For HA, this is a comma-separated list of all IPs within the NAT gateway pods.
+                  LAN IP address(es) observed on the NAT gateway Pods.
+                  For non-HA, this is the single address persisted to spec.lanIp.
+                  For HA, this is a comma-separated, sorted list of all observed addresses.
                 type: string
               qosPolicy:
                 description: QoS policy applied to the NAT gateway
--- a/dist/images/install.sh
+++ b/dist/images/install.sh
@@ -7604,6 +7604,7 @@ rules:
       - vpcs/status
       - vpc-nat-gateways
       - vpc-nat-gateways/status
+      - vpc-nat-gateways/finalizers
       - vpc-egress-gateways
       - vpc-egress-gateways/status
       - subnets
@@ -7745,6 +7746,7 @@ rules:
       - watch
       - create
       - update
+      - patch
       - delete
   - apiGroups:
       - ""
@@ -7792,9 +7794,11 @@ rules:
     verbs:
       - get
       - list
+      - watch
       - create
       - delete
       - update
+      - patch
   - apiGroups:
       - ""
     resources:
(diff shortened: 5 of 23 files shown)


zbb88888:
# 关于设计


────────────────────────────────────────────────────────────────────────────────

 推荐的最终架构

 建议保持以下结构：

 ```text
   VpcNatGateway informer
       └── 直接 reconcile VpcNatGateway

   StatefulSet informer
       └── StatefulSet ownerReference
             └── VpcNatGateway queue

   Deployment informer
       └── Deployment ownerReference
             └── VpcNatGateway queue

   Pod informer
       └── FilteringResourceEventHandler
             └── NAT Gateway label filter
                   └── Pod owner chain resolver
                         ├── Pod -> StatefulSet -> VpcNatGateway
                         └── Pod -> ReplicaSet -> Deployment -> VpcNatGateway
 ```

 队列统一：

 ```text
   addOrUpdateVpcNatGatewayQueue.Add(gw.Name)
 ```

 这满足：

 - owner reference 三层链式关系；
 - 精准触发；
 - 不遍历全部资源；
 - Pod 状态变化可及时触发；
 - Workload 变化可及时触发；
 - 最小性能开销；
 - 最小代码侵入。

 结论

 你的设计方向正确，且当前代码已经实现了这个目标：

 │ VpcNatGateway CRD 作为领域根资源，StatefulSet/Deployment 和 Pod 通过 OwnerReference 建立链式归属，Controller 监听各层事件并沿 OwnerReference 精确回调 VpcNatGateway。

 需要明确的是：

 - OwnerReference 本身不会自动触发回调；
 - 回调必须由 Controller Watch + owner mapper 实现；
 - 当前 FilteringResourceEventHandler + vpcNatGatewayOwnerFromPod 正是这个机制；
 - 不应为消除普通 Pod 的一次 label filter 而引入独立 informer；
 - 不建议进一步把逻辑移动到 pod.go。

 当前方案可以作为最终实现保留。

changluyi:
Thanks for the PR. I did a deep pass over the controller/webhook changes; findings ordered by severity:

## Major

### 1. Webhook may still reject empty `lanIp` (would block the feature itself)
`pkg/webhook/vpc_nat_gateway.go` (hunk `@@ -392,11 +414,12`): this hunk only relaxes the "required" and `net.ParseIP` checks for `lanIp`, but the subsequent `CIDRContainIP(subnet.Spec.CIDRBlock, gw.Spec.LanIP)` check is unchanged — an empty string is not contained in any CIDR, so creating a dynamic (empty-`lanIp`) non-HA/HA VpcNatGateway may still be rejected. The CIDR check should only run when `lanIp` is non-empty. Note `TestValidateVpcNatGatewayAllowsDynamicLanIP` doesn't cover this later check, so this wouldn't be caught by the new tests.

### 2. Owner-reference migration can adopt workloads belonging to other gateways
`pkg/controller/vpc_nat_gateway_owner.go`, `vpcNatGatewayControllerReferencePatch` (called from `pkg/controller/vpc_nat_gateway.go` hunks `@@ -350` / `@@ -397`): when migrating legacy workload owner refs, only the *controller* owner is validated (`metav1.GetControllerOf(current)`). Owners set by older versions via `SetOwnerReference` are **not** controller refs, so `GetControllerOf` returns `nil` and the existing VpcNatGateway owner is replaced unconditionally — even if it points to a different VpcNatGateway with a different UID (e.g. after delete + recreate of a same-name CR, or leftover workloads). The current CR can then adopt a StatefulSet/Deployment it doesn't own, and deleting it will GC-delete that workload through the new controller owner ref.

### 3. Whole `ownerReferences` array replaced during migration
Same function: the patch replaces the entire `metadata.ownerReferences` with `desired.GetOwnerReferences()` instead of migrating only the VpcNatGateway owner entry. If a legacy workload carries other legitimate (non-controller) owner refs, they get dropped, changing its GC lifecycle. Please confirm `genNatGwStatefulSet`/`genNatGwDeployment` always preserves existing refs — this isn't evident from the diff.

### 4. `status.lanIp` updated before conflict detection (non-HA)
`pkg/controller/vpc_nat_gateway.go` (around diff lines 1172–1195 and 1275–1311): the observed pod IP is written to `status.lanIp` first, then `persistNatGwLanIP` checks for conflict with `spec.lanIp`. When the pod IP differs from the persisted `spec.lanIp`, the resource briefly exposes a wrong status before the conflict error returns and reconcile retries. Consider detecting the conflict before patching status and leaving status unchanged on conflict.

## Minor

5. **Test gaps** (`pkg/controller/vpc_nat_gateway_owner_test.go`): only HA dual-stack observation is covered; missing non-HA IPv6, non-HA dual-stack selection, and pod-rebuild scenarios. There's also no full-reconcile test for two running pods observing different LAN IPs; the conflict tests call the helpers directly without asserting the status/spec patch ordering.
6. **e2e coverage** (`test/e2e/iptables-vpc-nat-gw/e2e_test.go`): only IPv4 non-HA dynamic allocation is covered; IPv6, dual-stack, and LAN IP conflict paths remain untested.
7. **Silent `continue` on owner-chain resolution errors** (`pkg/controller/vpc_nat_gateway.go`, ~diff lines 1172–1175): errors are neither logged nor requeued. With informer caches not yet synced (or a transiently missing owner), this can silently skip LAN IP persistence for that round; it also conflicts with `CODE_STYLE.md` on not swallowing errors.

## Checked and found OK

- Both charts' CRD changes are in sync; no new `required`/`enum` fields, `lanIp` remains optional — old CRs stay compatible.
- RBAC additions (deployments/statefulsets patch, statefulsets watch, replicasets get/list/watch) match the new call paths; `dist/images/install.sh` is synced.
- The `/status/lanIp` persistence uses a JSON Patch `test` op against the observed value, so concurrent writers on the same CR are CAS-protected.
- Informer cache sync waiting and workqueue handler registration for sts/deploy/rs look correct; no custom finalizer is added on the delete path (cleanup relies on owner-ref GC) — no deadlock evidence found.

changluyi:
**P1: Persisted LAN IP is not propagated to the existing StatefulSet template**\n\nIn , the controller persists the dynamically allocated LAN IP into . However,  at  does not compare . As a result, the next reconcile can generate a StatefulSet template with the static LAN IP annotation, but  remains false and the existing StatefulSet is not updated.\n\nAfter the gateway Pod is deleted and recreated, it can receive a different LAN IP.  then sees the new IP conflicting with the persisted , so the gateway cannot recover. The newly added E2E case in  appears intended to cover exactly this restart scenario and should expose the issue.\n\nCould we update the StatefulSet when the persisted LAN IP changes, or compare the generated Pod template annotations/hash with the existing StatefulSet before deciding that no update is needed?

changluyi:
**P1: Persisted LAN IP is not propagated to the existing StatefulSet template**

In `pkg/controller/vpc_nat_gateway.go:250`, the controller persists the dynamically allocated LAN IP into `spec.lanIp`. However, `isVpcNatGwChanged` at `pkg/controller/vpc_nat_gateway.go:439` does not compare `LanIP`. As a result, the next reconcile can generate a StatefulSet template with the static LAN IP annotation, but `gwChanged` remains false and the existing StatefulSet is not updated.

After the gateway Pod is deleted and recreated, it can receive a different LAN IP. `patchNatGwStatus` then sees the new IP conflicting with the persisted `spec.lanIp`, so the gateway cannot recover. The newly added E2E case in `test/e2e/iptables-vpc-nat-gw/e2e_test.go:764-777` appears intended to cover exactly this restart scenario and should expose the issue.

Could we update the StatefulSet when the persisted LAN IP changes, or compare the generated Pod template annotations/hash with the existing StatefulSet before deciding that no update is needed?

zbb88888:
### 三处兼容路径

 1. vpcNatGatewayOwnerRef 的 all-owner 回退（pkg/controller/vpc_nat_gateway_owner.go）
 v1.18 之前用的是 SetOwnerReference，存量 workload 没有 controller ref，只能遍历全部 ownerReferences 匹配。

 ```go
   // TODO: drop the fallback once every supported upgrade path has passed through
   // v1.18, i.e. once no workload can still carry a plain (non-controller) owner
   // reference. vpcNatGatewayControllerReferencePatch below can then go as well,
   // together with the statefulsets/deployments `patch` RBAC verbs it requires.
 ```

 2. vpcNatGatewayControllerReferencePatch 迁移（同文件）
 把存量 workload 就地补成 controller ref，每个对象最多 patch 一次（迁移完成后 needed=false）。

 ```go
   // TODO: remove together with the fallback in vpcNatGatewayOwnerRef.
 ```

 3. HA 模式容忍存量脏 lanIp（pkg/webhook/vpc_nat_gateway.go）
 v1.18 前 lanIp 必填，HA 对象也可能带着无意义的值，现在不拒绝。

 ```go
   // TODO: reject a non-empty lanIp in HA mode once no pre-v1.18 object can
   // still carry one.
 ```

 ### 移除顺序（依赖关系）

 ```text
   1. 迁移逻辑(2)运行 ≥1 个发布周期，把存量对象全部补成 controller ref
   2. 之后可删回退(1)
   3. 再之后删迁移(2)本身 + statefulsets/deployments 的 patch RBAC
 ```

 注意 (1) 与 (2) 是「先迁移、后收紧」的关系，不能倒过来删：先删迁移会让存量对象永远停在旧格式，再删回退就直接失联。

 三处都锚定在 v1.18，可在一个 issue 里统一跟踪。RBAC 里 statefulsets/deployments 的 patch 动词也只服务于迁移，届时一并回收。

changluyi:
Additional findings on the current head aa416b7:

1. P1: full workload Updates omit metadata.resourceVersion

Both pkg/controller/vpc_nat_gateway.go:392-394 (HA Deployment) and :452-454 (StatefulSet) call Update with a newly generated object. genNatGwDeployment/genNatGwStatefulSet construct a fresh object and do not copy the existing workload resourceVersion. When the dynamic LAN IP changes template annotations, the StatefulSet path enters this Update and the real API server rejects it with "metadata.resourceVersion: must be specified for an update" (the fake client does not enforce this). The Deployment path has the same issue whenever hashChanged or parametersChanged is true. Please preserve oldDeploy/oldSts.ResourceVersion (and the existing metadata needed for an update), or update the fetched object in place.

2. P2: a later full Update can discard unrelated OwnerReferences

vpcNatGatewayControllerReferencePatch correctly preserves unrelated owner references while migrating the old workload. However, if the same reconcile then takes the full Update path above, newDeploy/newSts contain only the VpcNatGateway owner reference from util.SetControllerReference, so Update replaces the entire ownerReferences array and drops any other legitimate owners. This can change garbage-collection behavior. Please merge the generated spec/template into the fetched workload, or explicitly preserve oldDeploy/oldSts ownerReferences before Update.

zbb88888:
```bash


 ┌─────────────────┬──────────────────────────────────────────────────────────────────┬────────────────────────────────────────────────────────────────────────────────┐
 │ 维度            │ 改动前                                                           │ 改动后                                                                         │
 ├─────────────────┼──────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────┤
 │ LAN IP          │ spec.lanIp 必填；backfill 靠「STS 名字前缀」反推网关，无身份校验 │ 空则动态分配，观测→status→spec 同步固化；归属经 ownerRef + UID 校验            │
 ├─────────────────┼──────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────┤
 │ 模板同步        │ 固化后模板不更新，Pod 重建即 IP 漂移 → 永久 conflict             │ 检出 IP annotation 差异即更新，重建后复用固定地址                              │
 ├─────────────────┼──────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────┤
 │ 事件触发        │ Pod 事件 + workload 事件双路重复入队                             │ 仅 workload 事件，单一来源                                                     │
 ├─────────────────┼──────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────┤
 │ 全集群 Pod 事件 │ 每条都过 filter，NAT GW Pod 再做 maps.Equal + owner 链解析       │ 0                                                                              │
 ├─────────────────┼──────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────┤
 │ 工作负载状态    │ 每次 reconcile 一次 live API GET                                 │ lister 缓存读取                                                                │
 ├─────────────────┼──────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────┤
 │ 状态发布        │ 一次 reconcile 内两次 patchNatGwStatus                           │ 一次                                                                           │
 ├─────────────────┼──────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────┤
 │ 工作负载更新    │ 整体替换，抹掉 ownerReferences 与第三方元数据                    │ 原地更新，只覆盖 spec 与自有 annotation                                        │

```

changluyi:
## P1: Removing the Pod-to-NAT-gateway enqueue can leave dynamic allocation stuck

In the current head, `pkg/controller/pod.go` no longer enqueues the NAT gateway queues after `reconcileAllocateSubnets` successfully patches the Pod IP annotation. The controller now relies on StatefulSet/Deployment events to trigger `getNatGwObservedState()`.

There is no ordering guarantee that a StatefulSet/Deployment status event happens after the Pod IP annotation is written. If the workload event is processed first, the gateway reconcile sees no LAN IP, does not persist `spec.lanIp`, and does not enqueue the NAT gateway init flow. The subsequent Pod annotation update does not itself update StatefulSet/Deployment status, so there may be no later event to recover. This can leave both LAN IP persistence and NAT rule initialization stuck until an unrelated event occurs.

Please keep a filtered NAT-gateway Pod event/enqueue path, or enqueue the gateway and init queues after the successful NAT Pod annotation update. A regression test should cover the ordering where the workload event is observed before the Pod IP annotation update.

## P2: Owner-chain resolution errors are still dropped

`pkg/controller/vpc_nat_gateway.go:getNatGwObservedState` logs and skips Pods when `vpcNatGatewayOwnerFromPod` cannot resolve the owner chain. The caller then treats the incomplete observation as valid and does not requeue. With a transient informer-cache miss, the only running gateway Pod can be omitted and the status/init reconciliation can be missed permanently, especially now that Pod events no longer enqueue the gateway.

Please return a retryable error or explicitly requeue when owner-chain resolution fails.

zbb88888:
• ## 结论

  两个 commit 未发现功能性问题，代码、注释与测试相互一致；没有需要再“消融”掉的部分。工作区干净，因此不需要产生新的 commit——当前 PR 已经是两个 commit：
  11e2b0cd7（owner/controller ref）和 886e67310（动态 lanIp）。

  ## Commit 1 目的原则：用 Kubernetes 原生 ownership 收敛 NAT GW workload 生命周期

  - Workload 由 VpcNatGateway 以 controller reference 拥有，级联删除/GC 交给 k8s，不再靠名称猜测或人工维护。
  - owner reference 不产生事件，所以单独 watch 带 vpc-nat-gw=true 的 StatefulSet 和所有 Deployment，把 add/update/delete 事件通过 owner 映射回 gateway queue；通知只入
    队，实际收敛仍在 handleAddOrUpdateVpcNatGw 中按最新状态执行。

  - 对旧 workload（plain owner ref / 无 ref）保留一次性迁移 patch 和旧 StatefulSet 显式删除 fallback，并补齐 RBAC：statefulset watch/patch、deployment patch、vpc-nat-
    gateways/finalizers（controller reference 的 blockOwnerDeletion 需要）。

  - workload 更新改为“原地更新 spec/template”，不再整体替换对象，第三方 owner references 和 metadata 不再被冲掉。

  这个 commit 里没有多余改动：Deployment 的 delete 事件、tombstone 处理、UID 校验、迁移验证都是生命周期闭环所必需的。

  ## Commit 2 目的原则：lanIp 只由 NAT GW 控制器观察、发布、幂等回填

  - pod.go 现在只保留一个真正必要的触发：地址分配完成后入队一次 initVpcNatGateway；不再为每个 NAT GW Pod update 都入队 add/update。
  - 回填不再由 pod.go 直接写 spec。handleInitVpcNatGw 分成两个阶段：
      1. patchNatGwStatus：从 Pod annotation（不等 Running、不等 status.podIP）读取地址，先发布 status.lanIp，再在 spec.lanIp 为空且 status 匹配时用 JSON Patch
         test+replace 幂等写回；

      2. 原来的 exec pod 内 init，等 Pod running 后重试完成。

  - spec 回填后不需要控制器再“自己入队一次”：spec patch 产生的 CR update 事件天然进入 add/update queue，由它把地址写进 StatefulSet template，保证 Pod 重建仍拿同一个地
    址。

  - 因此你问的三点答案分别是：pod.go 只需一次 init 触发；vpc-nat-gw 在 init 阶段内直接回填；回填基于 pod annotation 而非 pod status IP。

  幂等性由三层保证：JSON Patch 的 test 条件、status 先于 spec 发布、webhook 只允许从 status 回填且 spec 一旦设置即不可变。

  ## 生命周期事件链的源码核对

  - StatefulSet controller 注册了 Pod add/update/delete handler 并 enqueue 对应 StatefulSet；状态在 updateStatefulSetStatus 中仅在 inconsistentStatus 时写入（pkg/
    controller/statefulset/stateful_set.go、stateful_set_control.go）。Pod 创建/删除/Ready 变化都会改变 Replicas/ReadyReplicas 等计数，产生 update 事件。

  - Deployment 链路上，ReplicaSet controller 对 Pod 事件执行“always updates status as pods come up or die”，计数不同才写；Deployment controller watch ReplicaSet 和
    Pod delete 事件，status 不同才写（pkg/controller/replicaset/replica_set_utils.go、pkg/controller/deployment/sync.go）。

  - 容器重启不会改变 workload 计数，所以 pod.go 里的 enqueueVpcNatGatewayRestart 是必要的保留触发，不能消融。
  - CR 的 add/update/delete 由 kubeovn informer 驱动，delete tombstone 已有专门处理；CR 删除时 add/update 路径直接进入 handleDelVpcNatGw，先清理再移除 finalizer。

  由于 workload 状态事件可能先于 pod informer cache 更新，listNatGwPods 改为直接读 API server，是解决这个竞态的必要手段，代价只限于每个 NAT GW reconcile 的一次小范围
  List。

  ## 验证结果

  - go test ./pkg/controller ./pkg/webhook ./pkg/util：全部通过
  - E2E 包 test/e2e/iptables-vpc-nat-gw：编译通过
  - golangci-lint 三个改动相关包：0 issues
  - make verify-crd：通过，CRD 与 install manifest 同步
  - git status：clean，本地 HEAD 与 origin/nat-gw-own-ref 一致

  API 兼容性方面：旧的非 HA lanIp 必填约束被放宽为空（允许动态分配），已有静态 lanIp 语义不变；旧 workload 的 plain owner ref 会一次性迁移；没有删除或弱化已有 RBAC。

changluyi (commented):
Thanks for the update. I re-reviewed the current head (`886e673`). The four findings from my earlier pass are all addressed — empty `lanIp` is now accepted by the webhook, the StatefulSet template is re-synced via `templateChanged`, `updatedSts`/`updatedDeploy` are `DeepCopy`-ed so `resourceVersion` is preserved, and the Pod handler still triggers init/restart through `enqueueVpcNatGatewayInit` / `enqueueVpcNatGatewayRestart`. The owner-chain attribution with UID validation and the two-stage init split read well.

A few smaller things worth confirming before merge (none blocking):

**1. Dual-stack non-HA pins only the v4 address.** `selectNatGwLanIP` returns the v4 IP for `ProtocolDual`, so `spec.lanIp` only ever carries v4 and the v6 address is re-allocated on Pod recreation. This matches the pre-existing single-string `lanIp` constraint (the webhook rejects `"v4,v6"`), so it isn't a regression — but it's worth stating explicitly in the API/field doc so dual-stack users don't assume both families are pinned.

**2. `persistNatGwLanIP` tests `/status/lanIp` through the main-resource endpoint.** The JSON patch carries `{"op":"test","path":"/status/lanIp",...}` but is issued to `VpcNatGateways().Patch(...)` without the `status` subresource. This works because the `test` op reads the same underlying object and is read-only, but it's subtle enough that a follow-up author may "fix" it by adding the `status` subresource and break the `replace /spec/lanIp`. A one-line comment explaining the intent would help.

**3. The `vpc-nat-gateways/finalizers` RBAC entry looks unused.** I couldn't find any finalizer logic added in this PR — the persist path patches `/spec/lanIp` on the main resource. If it's forward-looking that's fine, but a permission nobody uses isn't ideal for least-privilege; worth confirming.

Minor: `handleInitVpcNatGw` and `handleAddOrUpdateVpcNatGw` now live-list Pods on every reconcile; fine at small scale, worth keeping in mind for controllers managing many gateways.

zbb88888:
1. 双栈非 HA 引脚仅绑定 v4 地址。 后续考虑继续支持。
2. persistNatGwLanIP 通过主资源端点对 /status/lanIp 进行了测试。JSON 补丁中包含了 {"op":"test","path":"/status/lanIp",...} 的数据，但数据被发送到 VpcNatGateways().Patch(...) ，而无需使用 status 子资源。这样做是因为 test 操作只是读取同一个底层对象，而且该操作是只读的。不过，这种差异可能很容易被后续作者“修复”，他们可能会添加 status 子资源，从而破坏 replace /spec/lanIp 的功能。附上一条解释此行为的注释会有所帮助。 ----- 看起来是一个测试的问题，后续我再分析下这点。

3. vpc-nat-gateways/finalizers 这个可以作为前瞻性保留
4. handleInitVpcNatGw 和 handleAddOrUpdateVpcNatGw 现在在每次同步时都会显示实时状态；在小规模应用中效果不错，对于管理多个网关的控制器来说，这一点值得注意。 后续考虑优化开销

PR #7243, opened 2026-08-16
Add VpcIPsecGateway CRD for VPC site-to-site IPsec

emamihe:
## Summary
- Add cluster-scoped `VpcIPsecGateway` CRD for VPC site-to-site IPsec (PSK via Secret), modeled after `VpcNatGateway`.
- Controller provisions a privileged strongSwan StatefulSet, swanctl ConfigMap, and OVN static routes so VPC traffic to `remoteCIDRs` is routed via the gateway `lanIp`.
- Wire compatibility pieces: client/codegen, webhooks, RBAC, Helm/install/cleanup/GC, feature ConfigMaps (disabled by default), and gateway image under `dist/images/vpcipsecgateway/`.

## Test plan
- [ ] `make gen-crd` / `make verify-crd` and `make lint` pass
- [ ] Enable `ovn-vpc-ipsec-gw-config` (`enable-vpc-ipsec-gw=true`) and set image in `ovn-vpc-ipsec-config`
- [ ] Create PSK Secret + `VpcIPsecGateway` (see `yamls/vpc-ipsec-gateway.yaml`)
- [ ] Confirm StatefulSet/pod starts, init annotation is set, and OVN static routes for `remoteCIDRs` point at `lanIp`
- [ ] Delete the CR and confirm STS/ConfigMap cleanup and route removal


zhangzujian (changes requested):
Request changes: the PR has several correctness and release-blocking issues.

### P1 findings

1. **Existing v2 Helm upgrades can prevent the controller from starting.** The new CRD is added only under `charts/kube-ovn-v2/crds/kube-ovn-crd.yaml`. Helm installs `crds/` files only on install and does not create newly added CRDs during upgrade. However, `pkg/controller/controller.go:452,836-869` unconditionally starts and waits for the `VpcIPsecGateway` informer. On an existing v2 release, the list/watch returns 404 and `WaitForCacheSync` never completes. Add an upgrade path for the CRD or make this informer optional until the CRD exists.

2. **The arm64 release is currently unbuildable.** `dist/images/vpcipsecgateway/Dockerfile:14` installs `strongswan-swanctl`, but the arm64 build reports `strongswan-swanctl (no such package)`: [Build arm64](https://github.com/kubeovn/kube-ovn/actions/runs/32467329844/job/100296264070). Since `release-arm` builds this image, the release job fails.

3. **The webhook hard-codes `kube-system` while the controller and charts support a configurable namespace.** `pkg/webhook/vpc_ipsec_gateway.go:59-85,138-144` looks up both feature ConfigMaps and defaults the PSK Secret to `kube-system`; the controller uses `c.config.PodNamespace` and Helm renders `.Values.namespace`. Any non-`kube-system` installation will reject an otherwise valid gateway because its configuration is not found.

4. **The feature gate fails open for malformed configuration.** `pkg/controller/vpc_ipsec_gateway.go:76` disables the feature only when the value is exactly `"false"`. A missing or misspelled `enable-vpc-ipsec-gw` value therefore enables reconciliation and can create privileged strongSwan workloads. Treat every value other than `"true"` as disabled.

5. **A controller restart can leave an existing gateway permanently uninitialized.** `initResourceOnce` calls `initVpcIPsecGw` (`pkg/controller/controller.go:1711`) while `vpcIPsecEnabled` may still be `unknown`; the function then returns without enqueueing init. The later resync (`pkg/controller/vpc_ipsec_gateway.go:93-102`) only enqueues add/update, not init. An existing running pod without the init annotation can remain `Ready=false` indefinitely.

### P2 findings

6. `pkg/controller/gc.go:161-175` only scans StatefulSets in the controller namespace, so `spec.namespace` gateways and their workloads are not garbage-collected. The same GC path does not remove orphan gateway ConfigMaps or IPsec routes.

7. `dist/images/cleanup.sh:80-81` does not delete the new `ovn-vpc-ipsec-config` and `ovn-vpc-ipsec-gw-config` ConfigMaps during uninstall.

8. The PR test plan remains unchecked for StatefulSet/Pod initialization, route programming, and deletion cleanup. The added tests only cover swanctl rendering and utility naming, so the lifecycle regressions above are not covered.

Local verification: `go test ./pkg/controller ./pkg/webhook ./pkg/util` and `make verify-crd` pass. The review applies to head `d14ba57efd5c89bc370b47c0114b1978a1de60dd`.

PR #7251, opened 2026-08-18
fix: preserve VEG policies without dangling address sets

### LINKED ISSUE #7247: [BUG]  VpcEgressGateway (subnet-scoped, no selector) leaves a dangling logical_router_policy referencing a non-existent port-group address set → ovn-controller "expecting address set name" on every recompute
### Kube-OVN Version

v1.16.2

### Kubernetes Version

Server Version: v1.36.2

Talos Linux v1.13.7 — kernel 6.18.39-talos (amd64) — containerd 2.2.6

### Description

A `VpcEgressGateway` that uses **only subnet-based SNAT policies** (`spec.policies[].subnets`) with **no namespace/pod selector** reaches `phase: Completed` / `ready: true` and SNAT works, but `ovn-contr
oller` logs a recurring flow-parse error on every recompute:

```
lflow|WARN|error parsing match "ip4.src == $VEG.895d4d253ef1_ip4":
  Syntax error at `$VEG.895d4d253ef1_ip4' expecting address set name.
```

Root cause (pkg/controller/vpc_egress_gateway.go, the `matches := set.New(...)` block in `reconcileVpcEgressGatewayOVNRoutes`): the controller creates **two** `logical_router_policy` reroute rules per ad
dress family on the VPC router:

```go
matches := set.New(
    fmt.Sprintf("ip%d.src == $%s_ip%d", af, pgName, af),  // port-group form  -> $VEG.<hash>_ip4
    fmt.Sprintf("ip%d.src == $%s",     af, asName),        // address-set form -> $VEG.<hash>.ipv4
)
```

`asName = "VEG.<hash>.ipv4"` (a standalone address set that is created and populated) resolves fine. `pgName = "VEG.<hash>"` is an OVN **port group** whose auto-generated address set is referenced as `$V
EG.<hash>_ip4`. The port group is created **empty** (it is populated only with client-pod LSPs selected by the VEG's `spec.selectors`, and this VEG has none), and is then removed (GC'd / not persisted),
while the port-group-based policy remains. `ovn-controller` can then no longer resolve `$VEG.<hash>_ip4` and fails to parse that logical flow on every recompute.

Observed on the live cluster:
- NB `logical_router_policy` on the VPC router `tenant-a`: **two** reroute policies at the egress-gateway priority (29100), both `external_ids: {vendor=kube-ovn, vpc-egress-gateway="vpc-gateways/tenant-a
-egress"}`, nexthop `192.168.50.2`:
  - `match: "ip4.src == $VEG.895d4d253ef1.ipv4"`  → resolves (address set exists)
  - `match: "ip4.src == $VEG.895d4d253ef1_ip4"`   → **dangling** (no such port group / address set)
- NB/SB `address_set`: `VEG.895d4d253ef1.ipv4` exists; **no** `VEG.895d4d253ef1_ip4`.
- NB `port_group`: **no** `VEG.895d4d253ef1`.
- Restarting `kube-ovn-controller` does **not** recreate the port group. Only an explicit VEG **update event** re-runs `reconcileVpcEgressGatewayOVNRoutes`, which (re)creates the empty port group — after
 which the warning clears until the empty port group is GC'd again.

The same two-form `matches` block is present in v1.14.0, v1.15.0 and v1.16.2, so this is not specific to a single patch release.

### Steps To Reproduce

1. Create a custom VPC (`tenant-a`) with an overlay subnet (`tenant-a-default`, e.g. 192.168.50.0/24).
2. Create a `VpcEgressGateway` with **only** a subnet-based SNAT policy and **no** `selectors`:
   ```yaml
   apiVersion: kubeovn.io/v1
   kind: VpcEgressGateway
   metadata:
     name: tenant-a-egress
     namespace: vpc-gateways
   spec:
     vpc: tenant-a
     externalSubnet: public-subnet
     internalIPs: ["192.168.50.2", "192.168.50.3"]
     replicas: 1
     bfd:
       enabled: false
     policies:
       - snat: true
         subnets: [tenant-a-default]
   ```
3. Wait for the VEG to reach `ready: true` / `phase: Completed`.
4. Tail any node's `ovs-ovn` (openvswitch / ovn-controller) logs, or run `ovn-appctl -t ovn-controller inc-engine/recompute`.

### Current Behavior

- `ovn-controller` logs on every recompute (the controller force-recomputes on its own every ~10 min):
  ```
  lflow|WARN|error parsing match "ip4.src == $VEG.<hash>_ip4":
    Syntax error at `$VEG.<hash>_ip4' expecting address set name.
  ```
- `port_group VEG.<hash>` does not exist; the `$VEG.<hash>_ip4` `logical_router_policy` is dangling and never programs a flow.
- SNAT still works via the parallel `$VEG.<hash>.ipv4` policy, so it is functionally benign, but it produces continuous WARN-level log noise and one permanently unprogrammable logical flow per recompute.

## Expected Behavior
No dangling policy and no parse error. Either:
- the `VEG.<hash>` port group is created and **kept for the lifetime of the VEG even when empty**, so `$VEG.<hash>_ip4` always resolves; or
- the port-group-based policy (`$VEG.<hash>_ip4`) is **not added when the VEG has no pod/namespace selector** (subnet-only), leaving only the address-set policy (`$VEG.<hash>.ipv4`).

### Expected Behavior

No dangling policy and no parse error. Either:
- the VEG.<hash> port group is created and kept for the lifetime of the VEG even when empty, so $VEG.<hash>_ip4 always resolves; or
- the port-group-based policy ($VEG.<hash>_ip4) is not added when the VEG has no pod/namespace selector (subnet-only), leaving only the address-set policy ($VEG.<hash>.ipv4).

---
I can't submit it for you (no GitHub auth in this session). Two ways to file it:

1. Open the form link above and paste each block into its field, or
2. If you want, tell me and I'll build a pre-filled URL (I have the field IDs: kube-ovn-version, kubernetes-version, os-info, description, repro, current-behavior, expected-behavior) so the form opens already populated — you'd just review and click Submit.

zhangzujian:
# Pull Request

## What type of this PR

- Bug fixes
- Tests

For a VpcEgressGateway without selectors, keep the existing two-policy contract without leaving an unresolvable port-group address-set reference. Subnet-only gateways now remove the unused Port Group and create an empty standalone compatibility address set named `VEG.<hash>_ip4` or `VEG.<hash>_ip6`. The existing policy therefore remains parseable and matches no traffic, while subnet traffic continues to use the populated `VEG.<hash>.ipv4` or `VEG.<hash>.ipv6` address set.

Gateways with selectors continue to use the Port Group auto-generated address sets. Reconciliation removes the standalone compatibility address set before using the Port Group, so the name has exactly one owner. Gateway deletion cleans up both address-set forms.

Regression tests cover subnet-only and selector resource ownership, regular reroute and BFD drop policies, IPv6 local-policy matches, stale-policy cleanup, deletion cleanup, and the two-policy E2E assertion.

Fixes #7247

## Testing

- `gofmt` completed successfully.
- `git diff --check` completed successfully.
- `go test policy_nexthops.go policy_nexthops_test.go -count=1` passed under the required 512 MiB / 0.5 CPU systemd scope.
- The targeted `pkg/controller` test was attempted with `-p 1` and compiler concurrency `-c=1`, but the Go compiler still exceeded the required 512 MiB cgroup limit before tests could run. It was not retried without the limit; GitHub Actions provides the compiled controller test result.

### DIFF
--- a/pkg/controller/vpc_egress_gateway.go
+++ b/pkg/controller/vpc_egress_gateway.go
@@ -849,6 +849,26 @@ func setVpcEgressGatewayBandwidthAnnotations(annotations map[string]string, band
 	return nil
 }
 
+func vpcEgressGatewayPolicyMatches(af int, pgName, asName string, includePortGroup bool) set.Set[string] {
+	matches := set.New[string](fmt.Sprintf("ip%d.src == $%s", af, asName))
+	if includePortGroup {
+		matches.Insert(fmt.Sprintf("ip%d.src == $%s_ip%d", af, pgName, af))
+	}
+	return matches
+}
+
+func vpcEgressGatewayLocalPolicyMatches(af int, localPgName, pgName, asName string, includePortGroup bool) set.Set[string] {
+	matches := set.New[string](fmt.Sprintf(
+		"ip%d.src == $%s_ip%d && ip%d.src == $%s", af, localPgName, af, af, asName,
+	))
+	if includePortGroup {
+		matches.Insert(fmt.Sprintf(
+			"ip%d.src == $%s_ip%d && ip%d.src == $%s_ip%d", af, localPgName, af, af, pgName, af,
+		))
+	}
+	return matches
+}
+
 func (c *Controller) reconcileVpcEgressGatewayOVNRoutes(gw *kubeovnv1.VpcEgressGateway, af int, lrName, lrpName, bfdIP string, nodeNexthops map[string]set.Set[string], sources set.Set[string]) error {
 	nextHops := flattenVpcEgressGatewayNexthops(nodeNexthops)
 
@@ -905,6 +925,7 @@ func (c *Controller) reconcileVpcEgressGatewayOVNRoutes(gw *kubeovnv1.VpcEgressG
 	}
 	key := cache.MetaObjectToName(gw).String()
 	pgName := vegPortGroupName(key)
+	includePortGroup := len(gw.Spec.Selectors) > 0
 	if err = c.OVNNbClient.CreatePortGroup(pgName, externalIDs); err != nil {
 		err = fmt.Errorf("failed to create port group %s: %w", pgName, err)
 		klog.Error(err)
@@ -963,8 +984,9 @@ func (c *Controller) reconcileVpcEgressGatewayOVNRoutes(gw *kubeovnv1.VpcEgressG
 				return err
 			}
 			localPgName := strings.ReplaceAll(portName, "-", ".")
-			rules[fmt.Sprintf("ip%d.src == $%s_ip%d && ip%d.src == $%s_ip%d", af, localPgName, af, af, pgName, af)] = nodeNextHops
-			rules[fmt.Sprintf("ip%d.src == $%s_ip%d && ip%d.src == $%s", af, localPgName, af, af, asName)] = nodeNextHops
+			for _, match := range vpcEgressGatewayLocalPolicyMatches(af, localPgName, pgName, asName, includePortGroup).UnsortedList() {
+				rules[match] = nodeNextHops
+			}
 		}
 		policies, err := c.OVNNbClient.ListLogicalRouterPolicies(lrName, util.EgressGatewayLocalPolicyPriority, externalIDs, false)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


zhangzujian:
Temporarily closing to retrigger trusted automatic x86 E2E dispatch after the isolated-ref 404 hotfix landed on master in #7253.

zhangzujian:
Retrigger automatic x86 E2E after isolated-ref name check landed on master in #7254.

PR #7256, opened 2026-08-19
feat: add PrivateLink-style VPC endpoints for overlapping CIDRs

emamihe:
## Summary

Kube-OVN already lets each tenant run in its own VPC, so two tenants can both use `192.168.10.0/24` without colliding. Full VPC peering still cannot connect those networks, because overlapping subnets cannot be routed to each other.

This change adds a **service-style connection** between VPCs, similar to AWS PrivateLink:

- The **provider** publishes an existing Kubernetes Service from its VPC.
- The **consumer** gets a local IP **inside its own subnet**.
- Applications in the consumer VPC talk to that local IP as if the service lived next to them.
- Behind the scenes, traffic is translated over a unique **transit network** so overlapping tenant CIDRs never mix.

This is **not** a replacement for VPC peering. Peering remains the right tool when CIDRs do not overlap and you need full L3 reachability. Use this when you only need selected services across overlapping VPCs.

## What users get

**Provider tenant**
- Publishes a Service with a `VpcEndpointService` resource.
- Can optionally restrict which consumer VPCs are allowed.

**Consumer tenant**
- Creates a `VpcEndpoint` that points at that published service.
- Receives a local VIP in the chosen consumer subnet.
- Apps keep using a private IP in their own range; they do not need to know the provider CIDR or transit addressing.

**Cloud admin**
- Enables the feature explicitly (`--enable-vpc-endpoint=true`; **off by default**).
- Configures the transit CIDR used for this fabric (must not overlap join, pod, or service CIDRs).
- Default transit CIDR is `100.65.0.0/16` because the join subnet already uses `100.64.0.0/16`.

Helm / install knobs:
- `--enable-vpc-endpoint` / `ENABLE_VPC_ENDPOINT` / Helm `func.ENABLE_VPC_ENDPOINT` / `features.ENABLE_VPC_ENDPOINT` (default **false**)
- `--vpc-endpoint-transit-cidr` / `--vpc-endpoint-transit-switch`
- Helm v1: `networking.VPC_ENDPOINT_TRANSIT_CIDR`
- Helm v2: `networking.vpcEndpointTransit.cidr`

Requires `--enable-lb` and Multus (second NIC on stitcher pods, same pattern as `vpc-nat-gw`).

## How it works (short)

1. Consumer pod sends traffic to the **local VIP** in its own subnet.
2. A dual-NIC **consumer stitcher** pod DNATs that flow to the provider **TransitVIP** and SNATs onto the transit fabric.
3. A dual-NIC **provider stitcher** pod DNATs TransitVIP to Service backends (iptables `statistic nth` across EndpointSlice replicas) and MASQUERADEs return traffic on the VPC leg.

NAT/LB therefore runs **in privileged stitcher pods via iptables**, not on OVN logical-router NAT/LB. That avoids a shared gateway-chassis SPOF for this path.

## Example

```yaml
# Provider publishes a database Service from its VPC
apiVersion: kubeovn.io/v1
kind: VpcEndpointService
metadata:
  name: tenant-b-db
spec:
  vpc: tenant-b
  namespace: tenant-b-ns
  service: db
  allowedVpcs:
    - tenant-a

---
# Consumer attaches and gets a local IP in its own subnet
apiVersion: kubeovn.io/v1
kind: VpcEndpoint
metadata:
  name: tenant-a-db
spec:
  vpc: tenant-a
  subnet: tenant-a-subnet
  endpointService: tenant-b-db
```

Consumer apps then connect to `status.localVIP` on the `VpcEndpoint`.

## Out of scope / limitations

- Not full mesh connectivity between overlapping VPCs (no arbitrary pod-to-pod routing).
- Provider backends see the **provider stitcher VPC-leg IP** (iptables MASQUERADE), not the original overlapping consumer pod IP and not a transit SNAT IP.
- Provider load-balancing is iptables `nth` (per-packet), not OVN LB.
- Existing `VpcPeering` is unchanged and still requires non-overlapping CIDRs.
- Feature is disabled unless `--enable-vpc-endpoint=true`.

## Test plan

- [x] Unit: feature gate off skips init/workers; nil EndpointSlice Port falls back to TargetPort/Port; AllowedVpcs deny cleans consumer; transit ACL sync; GC orphans
- [x] E2E (Multus + `--enable-vpc-endpoint=true`, e.g. `kind-install-vpc-nat-gw` / `kind-install-vpc-endpoint`): overlapping CIDR path works; LocalVIP ∈ consumer subnet; denied `allowedVpcs` stays NotReady; deleting VEP removes consumer stitcher
- [ ] Manual: transit CIDR overlapping join/pod/service is rejected at controller start when the feature is enabled
- [ ] Manual: uninstall via `cleanup.sh` removes tenant stitcher ConfigMaps/Deployments and CRDs


zhangzujian:
Please rebase and fix the build failure.

emamihe:
Rebased onto latest master and fixed the Build kube-ovn failure (golangci-lint modernize: ptr.To(x) → &x in pkg/controller/vpc_endpoint.go).

pkg/ovs/ovn-nb-nat.go
```diff
@@ -130,6 +131,81 @@ func (c *OVNNbClient) CreateNats(lrName string, nats ...*ovnnb.NAT) error {
...
+		return err
+	}
+	for _, nat := range nats {
+		if nat.ExternalIP != externalIP {
+			continue
+		}
+		if nat.Match == match {
+			return nil
+		}
+		nat.Match = match
```
zhangzujian: This updates the existing SNAT row whenever the same externalIP/logicalIP pair is found. All endpoints in one VPC reuse that pair (the logical IP is 0.0.0.0/0 or ::/0), so creating endpoint B changes endpoint A's match to B's TransitVIP and stops A from being SNATed. Please use a per-endpoint NAT identity/rule, or otherwise preserve all destination matches.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,777 @@
...
+			return err
+		}
+		if controllerutil.ContainsFinalizer(eps, util.KubeOVNControllerFinalizer) {
+			controllerutil.RemoveFinalizer(eps, util.KubeOVNControllerFinalizer)
+			if _, err := c.config.KubeOvnClient.KubeovnV1().VpcEndpointServices().Update(context.Background(), eps, metav1.UpdateOptions{}); err != nil {
+				klog.Errorf("failed to remove finalizer from VpcEndpointService %s: %v", key, err)
+				return err
+			}
+		}
+		c.enqueueVpcEndpointsForService(eps.Name)
```
zhangzujian, line 331: This only requeues consumers when the VpcEndpointService itself is deleted. Deleting the provider Kubernetes Service (and its EndpointSlice objects) does not call this path: the EndpointSlice informer has no DeleteFunc and the Service delete handler only handles the legacy LB queue. The VES and all consumer LB/SNAT/VIP resources can therefore remain stale after the provider Service is deleted. Please add deletion-driven reconciliation and cleanup.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,777 @@
...
+		return fmt.Errorf("get subnet %s: %w", ep.Spec.Subnet, err)
+	}
+	if subnet.Spec.Vpc != ep.Spec.Vpc {
+		return fmt.Errorf("subnet %s belongs to vpc %s, not %s", ep.Spec.Subnet, subnet.Spec.Vpc, ep.Spec.Vpc)
+	}
+	eps, err := c.vpcEndpointServiceLister.Get(ep.Spec.EndpointService)
+	if err != nil {
+		return fmt.Errorf("get VpcEndpointService %s: %w", ep.Spec.EndpointService, err)
+	}
+	if !vpcEndpointServiceAllowed(eps, ep.Spec.Vpc) {
```
zhangzujian, line 832: The allow-list is checked only at CR reconciliation time. Every VPC with any endpoint gets a logical router port on the same transit switch, and every provider TransitVIP LSP is attached to that switch. That shared L2 fabric can expose a restricted service's TransitVIP to an already-attached but unauthorized VPC without creating this endpoint. Please enforce isolation in the dataplane (or use per-service isolation) rather than relying only on this control-plane check.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,777 @@
...
+		protocols[port.Protocol] = struct{}{}
+		lbName := vpcEndpointServiceLBName(eps.Name, string(port.Protocol))
+		if err := c.OVNNbClient.CreateLoadBalancer(lbName, strings.ToLower(string(port.Protocol))); err != nil {
+			return err
+		}
+		if err := c.OVNNbClient.LogicalRouterUpdateLoadBalancers(eps.Spec.Vpc, ovsdb.MutateOperationInsert, lbName); err != nil {
+			return err
+		}
+		vip := util.JoinHostPort(transitVIP, port.Port)
+		backends := c.getEndpointBackend(endpointSlices, port, transitVIP)
```
zhangzujian: This calls getEndpointBackend for the provider Service, but that helper only matches EndpointSlice ports when port.Name is non-nil and equal to servicePort.Name. For an ordinary unnamed Service port, EndpointSlice port.Name is nil, so no backends are installed and the endpoint VIP cannot reach the Service. Please handle unnamed ports (and nil Port defensively).

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,777 @@
...
+	v4, v6, mac, err := c.ipam.GetRandomAddress(ipamName, lspName, nil, c.config.VpcEndpointTransitSwitch, "", nil, true)
+	if err != nil {
+		return "", "", err
+	}
+	return vpcEndpointPreferIP(v4, v6), mac, nil
+}
+
+func (c *Controller) ensureVpcEndpointServiceLSP(name, ip, mac string) error {
+	lspName := vpcEndpointServiceLSPName(name)
+	if err := c.OVNNbClient.CreateLogicalSwitchPort(c.config.VpcEndpointTransitSwitch, lspName, ip, mac, name, metav1.NamespaceSystem, false, "", "", false, nil, ""); err != nil {
```
zhangzujian: Please verify this MAC choice against the existing SwitchLB VIP datapath. Here the TransitVIP LSP gets the IPAM-generated MAC, while the LB is attached to the provider logical router. The existing VIP implementation explicitly replaces the VIP LSP MAC with the subnet router-port MAC before enabling arp_proxy; without the equivalent router-facing MAC/route behavior, traffic may terminate on the transit switch instead of entering the provider router LB. An end-to-end overlapping-CIDR test should prove this path.

zhangzujian (changes requested):
Request changes.

I found several correctness and lifecycle issues that should be addressed before merging:

* Multiple endpoints in the same consumer VPC reuse one SNAT external IP, but the current NAT helper rewrites the existing rule's match. Creating a second endpoint can therefore break the first endpoint's SNAT.
* The provider Service/EndpointSlice deletion path does not enqueue the endpoint-service reconciliation, so provider-side and consumer-side OVN resources can remain stale after deletion.
* `allowedVpcs` is enforced only while reconciling the CR. All endpoint VPCs share one transit switch, so an attached VPC may be able to reach another service's transit VIP without an endpoint.
* The endpoint backend lookup does not support unnamed Service ports.
* The provider transit LSP uses an IPAM-generated MAC while the load balancer is attached to the provider router; this needs to follow the existing SwitchLB VIP datapath or be covered by an end-to-end test.
* The new CRDs are absent from `dist/images/cleanup.sh` and endpoint-specific GC, and the PR test plan is not covered by integration/E2E tests.

The current PR checks also show `x86-e2e / required-gate` as ACTION_REQUIRED and `Iptables VPC NAT Gateway E2E` as FAILURE.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,823 @@
...
+		return fmt.Errorf("get provider vpc %s: %w", eps.Spec.Vpc, err)
+	}
+
+	snatIP, err := c.ensureVpcTransitAttachment(eps.Spec.Vpc)
+	if err != nil {
+		return err
+	}
+	klog.V(5).Infof("provider vpc %s attached to transit with snat ip %s", eps.Spec.Vpc, snatIP)
+
+	svc, err := c.servicesLister.Services(eps.Spec.Namespace).Get(eps.Spec.Service)
```
zhangzujian, line 369: When the provider Service is missing, this branch only removes the provider-side protocol LBs and ACLs, then returns an error. The VpcEndpointService handler therefore exits before enqueueing its consumers, and reconcileVpcEndpoint also fails on the missing Service, so existing consumer LBs, SNAT rules, local Vip, and the VPC transit attachment are never cleaned up. This violates the deletion behavior described in the PR test plan. Please enqueue/clean all dependent VpcEndpoints and release the transit resources when the backing Service disappears.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,823 @@
...
+	return len(eps.Spec.AllowedVpcs) == 0 || slices.Contains(eps.Spec.AllowedVpcs, vpc)
+}
+
+func (c *Controller) enqueueAddVpcEndpointService(obj any) {
+	key := cache.MetaObjectToName(obj.(*kubeovnv1.VpcEndpointService)).String()
+	klog.Infof("enqueue add VpcEndpointService %s", key)
+	c.addOrUpdateVpcEndpointServiceQueue.Add(key)
+}
+
+func (c *Controller) enqueueUpdateVpcEndpointService(oldObj, newObj any) {
```
zhangzujian, line 73: VpcEndpointService spec fields are mutable, but the update handler discards oldEps and reconciles only the new spec. Changing vpc/namespace/service leaves the old provider router attached to its old LB/LSP/VIP resources; deleting the object later cleans only the current spec.vpc. Either make these identity fields immutable or explicitly tear down the old state before applying the new one.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,823 @@
...
+	c.addOrUpdateVpcEndpointServiceQueue.Add(key)
+}
+
+func (c *Controller) enqueueAddVpcEndpoint(obj any) {
+	key := cache.MetaObjectToName(obj.(*kubeovnv1.VpcEndpoint)).String()
+	klog.Infof("enqueue add VpcEndpoint %s", key)
+	c.addOrUpdateVpcEndpointQueue.Add(key)
+}
+
+func (c *Controller) enqueueUpdateVpcEndpoint(oldObj, newObj any) {
```
zhangzujian, line 109: VpcEndpoint spec is also mutable, while this update path retains no previous VPC/subnet/endpointService state. A change of VPC, subnet, or endpoint service creates/attaches resources for the new values but does not remove the old router LB, SNAT match, local Vip allocation, or transit attachment; cleanup later uses only the new spec. Please enforce immutability or reconcile the old state as part of updates.

pkg/controller/vpc_endpoint.go
zhangzujian, line 369: The fetched Service is not checked against eps.Spec.Vpc. A Service with no ovn.kubernetes.io/vpc annotation belongs to the cluster router, and an annotated Service may belong to another VPC, but this code publishes either one through the router named by eps.Spec.Vpc. That can install a provider LB on the wrong router and make the endpoint unusable, or expose a service from an unintended VPC. Please compare the Service effective VPC, with the existing cluster-router fallback, to eps.Spec.Vpc and reject mismatches.

dist/images/cleanup.sh
```diff
@@ -130,6 +130,8 @@ kubectl delete --ignore-not-found crd \
   vips.kubeovn.io \
   switch-lb-rules.kubeovn.io \
   vpc-dnses.kubeovn.io \
+  vpc-endpoint-services.kubeovn.io \
```
zhangzujian, line 134: The new CRDs are deleted below, but the finalizer-removal loop above still covers only subnets, vpcs, and ips. VpcEndpointService and VpcEndpoint instances carry kubeovn.io/kube-ovn-controller finalizers, so an uninstall with instances present can leave the CRDs stuck terminating and block cleanup. Add both resource types to the finalizer loop, and handle the empty-finalizers case.

pkg/controller/gc.go
```diff
@@ -1471,3 +1472,88 @@ func logicalRouterPortFilter(exceptPeerPorts *strset.Set) func(lrp *ovnnb.Logica
...
+		klog.Errorf("failed to list VpcEndpointServices for gc: %v", err)
+		return err
+	}
+	endpoints, err := c.vpcEndpointLister.List(labels.Everything())
+	if err != nil {
+		klog.Errorf("failed to list VpcEndpoints for gc: %v", err)
+		return err
+	}
+
+	expectedLBs := map[string]struct{}{}
```
zhangzujian, line 1498: This GC only tracks load balancers and provider transit LSPs. It does not detect orphan consumer/provider transit LRPs/LSPs, SNAT NAT rows, transit IPAM allocations, or local Vip CRs after a force-deleted object, failed finalizer path, or controller restart. Those resources can continue routing traffic and consuming transit addresses indefinitely. Please extend the endpoint GC to reconcile the complete resource set, or document and enforce a recovery path that makes these resources impossible to orphan.

zhangzujian (changes requested):
Request changes.

The previous review items for SNAT matching, delete event handlers, transit ACLs, unnamed ports, and provider MAC selection are addressed, but the following lifecycle and isolation issues remain.

emamihe:
## Update: datapath redesign

I changed the `VpcEndpoint` / `VpcEndpointService` implementation away from OVN logical-router NAT + load balancers (and the shared gateway-chassis dependency that came with it).

### New approach

Dual-NIC **stitcher** pods, same pattern as `vpc-nat-gw`:

- **Provider stitcher:** VPC NIC + transit NIC; iptables DNAT from TransitVIP → Service backends (nth load-balancing across EndpointSlice replicas)
- **Consumer stitcher:** VPC NIC + transit NIC; iptables DNAT from LocalVIP → TransitVIP, plus SNAT on the transit path
- Shared transit subnet/VPC as the unique fabric between overlapping tenant CIDRs

### Why

SNAT/DNAT on OVN logical routers still needs a gateway chassis, which is a central SPOF. In-pod iptables keeps NAT on the stitcher pods and avoids that.

### Ops notes

- Multus is required for the second NIC (same as `vpc-nat-gw`)
- The controller embeds and manages the `vpc-endpoint-stitcher` ConfigMap — no manual script install needed

### Verification

Smoke-tested on a multi-node OrbStack cluster (overlapping `10.210.0.0/24` provider/consumer VPCs, multi-replica provider Service, curl via LocalVIP → `200` with even backend distribution).

pkg/ovs/ovn-nb-nat.go
metsoft68: Addressed by stitcher redesign (see summary comment).

metsoft68: Addressed by the stitcher redesign: VpcEndpoint no longer uses OVN SNAT/NAT rows. Consumer SNAT is applied inside the dual-NIC stitcher pod via iptables, so the shared externalIP/logicalIP OVN NAT collision no longer applies.

pkg/controller/vpc_endpoint.go
metsoft68: Addressed: Service delete/update and EndpointSlice add/update/delete enqueue matching VpcEndpointServices (`enqueueVpcEndpointServiceForK8sService`). On VES reconcile failure (including missing Service), we deactivate the provider stitcher, patch Ready=false, and enqueue consumer VpcEndpoints so they tear down.

pkg/controller/vpc_endpoint.go
metsoft68: Fixed in c24da186: restricted services install transit-switch from-lport ACLs that drop traffic to TransitVIP except from authorized consumer stitcher LSPs. Open services (empty allowedVpcs) leave the VIP reachable. Consumer reconcile requeues the parent VES so ACL allow-lists stay in sync as stitchers come and go.

pkg/controller/vpc_endpoint.go
metsoft68: Addressed: `endpointSlicePortMatchesServicePort` matches unnamed Service ports to nil/empty EndpointSlice port names, and requires a non-nil Port. Provider stitcher mappings use this via `getEndpointBackend`.

pkg/controller/vpc_endpoint.go
metsoft68: Addressed by the stitcher redesign: there is no longer a TransitVIP LSP + provider router LB/arp_proxy path. The transit VIP is owned by the provider stitcher Multus interface; overlapping-CIDR coverage is in the Multus-gated e2e suite.

pkg/controller/vpc_endpoint.go
metsoft68: Addressed: missing provider Service deactivates the VES (deletes stitcher, clears TransitVIP/Ready) and the handler enqueues dependent VpcEndpoints. Consumer reconcile also cleans up when the Service or VES is missing/not ready.

pkg/controller/vpc_endpoint.go
metsoft68: Addressed: `vpc`/`namespace`/`service` are CRD-immutable via `XValidation: self == oldSelf`, with a controller-side label check as defense in depth.

pkg/controller/vpc_endpoint.go
metsoft68: Addressed: `vpc`/`subnet`/`endpointService`/`ip` are CRD-immutable via `XValidation`, with controller label checks for vpc/endpointService.

pkg/controller/vpc_endpoint.go
metsoft68: Addressed: reconcile compares `vpcEndpointEffectiveServiceVpc` (annotation / cluster-router fallback) to `eps.Spec.Vpc` and deactivates the VES on mismatch.

dist/images/cleanup.sh
metsoft68: Addressed: cleanup.sh patches finalizers for `vpc-endpoint-services.kubeovn.io` and `vpc-endpoints.kubeovn.io`, including the empty/null finalizers case, before deleting the CRDs.

pkg/controller/gc.go
metsoft68: Addressed and extended in c24da186: GC covers legacy LBs/LSPs/LRP attachments/SNAT/VIP CRs, plus orphan transit ACLs and stitcher Deployments for the current datapath.

metsoft68:
## Review feedback follow-up

Replied on each open review thread. Summary:

| Topic | Status |
| --- | --- |
| OVN SNAT match collision | Obsolete after stitcher redesign (iptables SNAT in-pod) |
| Service / EndpointSlice deletion | Addressed (enqueue VES + consumer cleanup) |
| AllowedVpcs dataplane isolation | **Fixed in `c24da186`** — transit ACLs on restricted services |
| Unnamed Service ports | Addressed (`endpointSlicePortMatchesServicePort`) |
| TransitVIP MAC / arp_proxy | Obsolete after stitcher redesign |
| Missing Service consumer cleanup | Addressed |
| VES / VEP identity immutability | Addressed (CRD `XValidation` + controller checks) |
| Service VPC mismatch | Addressed |
| cleanup.sh finalizers | Addressed |
| Endpoint GC completeness | Addressed + extended (orphan ACLs + stitcher Deployments) |

@zhangzujian please take another look when convenient; several threads can be resolved/outdated after the stitcher move.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,1155 @@
...
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	kubeovnv1 "github.com/kubeovn/kube-ovn/pkg/apis/kubeovn/v1"
+	"github.com/kubeovn/kube-ovn/pkg/ovs"
+	"github.com/kubeovn/kube-ovn/pkg/ovsdb/ovnnb"
+	"github.com/kubeovn/kube-ovn/pkg/util"
+)
+
+//go:embed vpc-endpoint-stitcher.sh
```
zhangzujian, line 34: The PR description still says "No extra proxy pods. The path stays in OVN." This embed, plus `genVpcEndpointStitcherDeployment`, creates privileged dual-NIC pods that DNAT/SNAT in iptables. That is a different product from the stated OVN NAT/LB path.

Please either restore an in-OVN datapath or rewrite the PR contract (including the "provider backends see a transit SNAT IP" limitation). If stitchers stay, they also need an explicit feature gate so this is off by default.

pkg/controller/config.go
```diff
@@ -294,6 +298,9 @@ func ParseFlags() (*Configuration, error) {
 		argNonPrimaryCNI = pflag.Bool("non-primary-cni-mode", false, "Use Kube-OVN in non primary cni mode. When true, Kube-OVN will only manage the network for network attachment definitions")
 
 		argSkipConntrackDstCidrs = pflag.String("skip-conntrack-dst-cidrs", "", "Comma-separated list of destination IP CIDRs that should skip conntrack processing")
+
+		argVpcEndpointTransitSwitch = pflag.String("vpc-endpoint-transit-switch", util.DefaultVpcEndpointTransitSwitch, "Logical switch used as the unique transit fabric for VPC endpoint services")
```
zhangzujian, line 311: `CODE_STYLE.md` CRD rule 1: new features should be disabled. These flags default to a real transit switch/CIDR, and `initVpcEndpointTransit` always runs when `EnableLb` is true (`controller.go`), which creates a transit VPC/subnet/NAD and starts workers.

Please add something like `--enable-vpc-endpoint` (default false), similar to `--enable-anp` / `--enable-lb-svc`, and do not create transit resources until it is on.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,1155 @@
...
+		return err
+	}
+	if err := c.ensureVpcEndpointStitcherConfigMap(); err != nil {
+		klog.Warningf("stitcher configmap ensure: %v", err)
+	}
+
+	svc, err := c.servicesLister.Services(eps.Spec.Namespace).Get(eps.Spec.Service)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			_ = c.deactivateVpcEndpointService(eps)
```
zhangzujian: `deactivateVpcEndpointService` errors are discarded here, and the same `_ = c.cleanupVpcEndpoint(...)` pattern appears on the consumer path (`:905`, `:914`, `:921`). If stitcher/OVN teardown fails, reconcile still returns the original NotFound/not-ready error and dependents can stay stale.

`CODE_STYLE.md` requires unexpected errors to be logged and returned. Please log and return the cleanup error instead of dropping it.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,1155 @@
...
+						ImagePullPolicy: corev1.PullIfNotPresent,
+						Lifecycle: &corev1.Lifecycle{
+							PostStart: &corev1.LifecycleHandler{
+								Exec: &corev1.ExecAction{
+									Command: []string{"sh", "-c", "sysctl -w net.ipv4.ip_forward=1"},
+								},
+							},
+						},
+						SecurityContext: &corev1.SecurityContext{
+							Privileged:               &privileged,
```
zhangzujian, line 642: The stitcher runs privileged in the provider/consumer namespace and `ensureVpcEndpointStitcherConfigMapIn` copies `vpc-endpoint-stitcher` into those tenant namespaces with no ownerRef.

`cleanup.sh` only deletes the kube-system ConfigMap. After uninstall, tenant-namespace ConfigMaps and any leftover stitcher pods can remain. Please set ownerRefs (or delete those ConfigMaps on deactivate/cleanup/GC/uninstall).

dist/images/vpcnatgateway/vpc-endpoint-stitcher.sh
```diff
@@ -0,0 +1,177 @@
...
+					-j DNAT --to-destination "$dest"
+			else
+				exec_cmd $iptables_cmd -t nat -A VPC_EP_PREROUTING -d "$transit_vip" -p "$proto" --dport "$port" \
+					-j DNAT --to-destination "$dest"
+			fi
+			i=$((i + 1))
+		done
+	done
+
+	# Return path: backend sees stitcher VPC-leg IP as source.
```
zhangzujian: Two mismatches with the PR contract:

1. Load-balancing above is iptables `statistic nth` (per-packet, connection-agnostic), not OVN LB.
2. This `MASQUERADE` on the VPC leg means provider backends see the stitcher VPC-leg IP, not the transit SNAT IP described in the PR ("Provider backends see the consumer as a transit SNAT IP").

Please align the datapath with the documented behavior, or update the limitation text and add an overlapping-CIDR e2e that asserts the observed source IP.

pkg/controller/endpoint_slice.go
```diff
@@ -739,6 +771,23 @@ func (c *Controller) getEndpointBackend(endpointSlices []*discoveryv1.EndpointSl
 	return backends
 }
 
+// endpointSlicePortMatchesServicePort matches EndpointSlice ports to a Service
+// port by protocol and name. Unnamed Service ports (empty name) match
+// EndpointSlice ports whose Name is nil or empty; Port must be non-nil.
+func endpointSlicePortMatchesServicePort(port discoveryv1.EndpointPort, servicePort v1.ServicePort) bool {
+	if port.Port == nil {
```
zhangzujian: The previous review asked unnamed Service ports and nil EndpointSlice `Port` to work. This still returns false when `port.Port` is nil, so `getEndpointBackend` installs no backends.

If `Port` is nil, fall back to `servicePort.TargetPort` / `servicePort.Port` instead of dropping the slice. Please add a unit test for an unnamed Service port whose EndpointSlice port has a nil `Port`.

test/e2e/kube-ovn/vpc-endpoint/vpc_endpoint.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+		if providerNS != "" {
+			nsClient.Delete(providerNS)
+		}
+		if consumerNS != "" {
+			nsClient.Delete(consumerNS)
+		}
+	})
+
+	framework.ConformanceIt("should expose a provider Service to a consumer VPC with an overlapping CIDR", func() {
```
zhangzujian: This is the only e2e, and it skips when Multus is unavailable (`:48`). It does not cover the PR test plan:

- a consumer VPC **not** in `allowedVpcs` cannot attach
- deleting the endpoint/service tears down unused transit attachments
- controller rejects a transit CIDR that overlaps join/pod/service CIDRs
- `status.localVIP` is allocated **from the consumer subnet** (this only checks non-empty at `:154`)

Please add those cases, or run this suite in a Multus-enabled job and assert the subnet of `LocalVIP`.

zhangzujian (changes requested):
Request changes.

The stitcher redesign addresses several of the earlier OVN NAT/LB lifecycle comments, but head `02a77effb` still has contract, default-on, cleanup, and coverage issues.

* The PR body still says "No extra proxy pods. The path stays in OVN." The implementation now deploys privileged dual-NIC stitcher pods and does DNAT/SNAT/load-balancing in iptables.
* `CODE_STYLE.md` requires new CRD features to be disabled. VPC endpoint transit is created whenever `--enable-lb` is true; there is no `enable-vpc-endpoint` gate.
* Cleanup paths discard stitcher/OVN errors with `_ =`, and tenant-namespace stitcher ConfigMaps have no ownerRef. `cleanup.sh` only deletes the kube-system ConfigMap.
* The test plan is not covered: denied `allowedVpcs`, unused transit teardown, transit CIDR overlap reject, and LocalVIP-in-consumer-subnet. The only e2e skips without Multus.
* Nil EndpointSlice `Port` still yields no backends. Provider `MASQUERADE` makes backends see the stitcher VPC-leg IP, not a transit SNAT IP.

Please update the PR contract or restore an in-OVN datapath, gate the feature off by default, make cleanup fail-closed, and add the missing tests.

pkg/controller/vpc_endpoint.go
metsoft68: Keeping the stitcher/iptables datapath. PR body rewritten in 1e58497d: dual-NIC privileged stitchers, Multus, iptables DNAT/SNAT/`nth`, and backends see the provider stitcher VPC-leg IP (MASQUERADE). Feature is gated by `--enable-vpc-endpoint` (default false).

pkg/controller/config.go
metsoft68: Fixed in 1e58497d: added `--enable-vpc-endpoint` (default false). Transit VPC/subnet/NAD, informers/handlers/workers, and GC only run when the flag is on (and `--enable-lb`). Charts/install default the flag off.

pkg/controller/vpc_endpoint.go
metsoft68: Fixed in 1e58497d: deactivate/cleanup errors are logged and returned on the VES/VEP reconcile paths instead of being discarded with `_ =`.

pkg/controller/vpc_endpoint.go
metsoft68: Fixed in 1e58497d: tenant-namespace stitcher ConfigMaps get an ownerRef to the owning VES/VEP, are deleted when unused on deactivate/cleanup, and `cleanup.sh` removes all `vpc-endpoint-stitcher` ConfigMaps plus leftover stitcher Deployments cluster-wide.

dist/images/vpcnatgateway/vpc-endpoint-stitcher.sh
metsoft68: Documented in 1e58497d: provider LB is iptables `nth`; backends see the stitcher VPC-leg IP via MASQUERADE (not a transit SNAT IP). PR limitations and stitcher script comments match that behavior.

pkg/controller/endpoint_slice.go
metsoft68: Fixed in 1e58497d: nil EndpointSlice `Port` falls back to Service `TargetPort` (int) or `Port`, with unit coverage for unnamed Service + nil slice Port.

test/e2e/kube-ovn/vpc-endpoint/vpc_endpoint.go
metsoft68: Expanded in 1e58497d: e2e asserts LocalVIP ∈ consumer CIDR, denied `allowedVpcs` stays NotReady, and VEP delete removes the consumer stitcher. Suite skips unless Multus is present and `--enable-vpc-endpoint=true` (enabled by `kind-install-vpc-nat-gw` / `kind-install-vpc-endpoint`). Transit CIDR overlap remains enforced by controller startup when the feature is on.

metsoft68:
## Follow-up for @zhangzujian review on `02a77eff`

Pushed `1e58497d` addressing the stitcher-contract review:

| Ask | Change |
| --- | --- |
| `--enable-vpc-endpoint` default false | Added; gates init/handlers/workers/GC; charts + install.sh default off |
| Fail-closed cleanup | deactivate/cleanup errors returned on VES/VEP paths |
| Tenant stitcher ConfigMaps | ownerRef + delete-when-unused + `cleanup.sh` cluster-wide CM/Deploy cleanup |
| MASQUERADE / nth vs docs | Documented: backends see stitcher VPC-leg IP; LB is iptables `nth` |
| Nil EndpointSlice `Port` | Falls back to TargetPort/Port + unit test |
| E2E gaps | LocalVIP ∈ consumer CIDR; denied `allowedVpcs`; VEP delete removes stitcher; skips unless Multus + flag on (`kind-install-vpc-nat-gw` / `kind-install-vpc-endpoint`) |

### Contract rewrite (for PR description)

NAT/LB runs in privileged dual-NIC stitcher pods via iptables (Multus required), not OVN LR NAT/LB. Feature is **off by default** (`--enable-vpc-endpoint=false`). Provider backends see the **provider stitcher VPC-leg IP** (MASQUERADE), not a transit SNAT IP. Provider LB is iptables `statistic nth`.

dist/images/vpcnatgateway/vpc-endpoint-stitcher.sh
```diff
@@ -0,0 +1,178 @@
...
+}
+
+init() {
+	local interfaces="${1:-}"
+	if [[ -n "$interfaces" ]]; then
+		VPC_INTERFACE=$(echo "$interfaces" | cut -d',' -f1 | tr -d ' ')
+		TRANSIT_INTERFACE=$(echo "$interfaces" | cut -d',' -f2 | tr -d ' ')
+	fi
+	save_env
+	exec_cmd sysctl -w net.ipv4.ip_forward=1
```
zhangzujian, line 51: This datapath only enables `net.ipv4.ip_forward` and later uses `iptables` (not `ip6tables`). `install.sh` still sets IPv6/dual transit CIDRs (`fd00:100:65::/112` and `100.65.0.0/16,fd00:100:65::/112`). Dual-stack/IPv6 clusters would get a transit network this stitcher cannot program.

Please either make the stitcher dual-stack, or reject IPv6/dual transit CIDRs and document IPv4-only.

dist/images/install.sh
```diff
@@ -150,6 +154,7 @@ if [ "$DUAL_STACK" = "true" ]; then
   POD_GATEWAY="10.16.0.1,fd00:10:16::1"
   SVC_CIDR="10.96.0.0/12,fd00:10:96::/108"               # Do NOT overlap with NODE/POD/JOIN CIDR
   JOIN_CIDR="100.64.0.0/16,fd00:100:64::/112"            # Do NOT overlap with NODE/POD/SVC CIDR
+  VPC_ENDPOINT_TRANSIT_CIDR="100.65.0.0/16,fd00:100:65::/112"
```
zhangzujian: Dual-stack install still assigns an IPv6 transit CIDR here (and IPv6-only at the ipv6 branch above). The stitcher only programs IPv4 iptables, so this CIDR cannot carry the published service path.

Please stop advertising an IPv6 transit network until the datapath supports it.

pkg/apis/kubeovn/v1/vpc-endpoint.go
```diff
@@ -0,0 +1,149 @@
...
+	Namespace string `json:"namespace"`
+	// Name of the provider Kubernetes Service.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="service is immutable"
+	Service string `json:"service"`
+	// Consumer VPCs allowed to attach. Empty means any VPC may consume the service.
+	AllowedVpcs []string `json:"allowedVpcs,omitempty"`
+}
+
+type VpcEndpointServiceStatus struct {
+	// Unique IP allocated from the transit subnet and used as the provider OVN LB VIP.
```
zhangzujian: This still documents TransitVIP as "the provider OVN LB VIP". The current design runs NAT/LB in privileged stitcher pods via iptables, not OVN logical-router LB.

Please update the CRD comments (and generated CRD descriptions) to match the stitcher contract.

pkg/apis/kubeovn/v1/vpc-endpoint.go
```diff
@@ -0,0 +1,149 @@
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="ip is immutable"
+	IP string `json:"ip,omitempty"`
+}
+
+type VpcEndpointStatus struct {
+	// IP in the consumer subnet that applications dial.
+	LocalVIP string `json:"localVIP,omitempty"`
+	// Provider transit VIP this endpoint maps to.
+	TransitVIP string `json:"transitVIP,omitempty"`
+	// Shared SNAT IP of the consumer VPC on the transit subnet.
```
zhangzujian: The PR contract says provider backends see the stitcher VPC-leg IP via MASQUERADE, not a transit SNAT IP. This field is filled from the consumer stitcher's transit-leg address, not a shared consumer-VPC SNAT IP.

Please rename or rewrite the comment so users are not told this is the source IP backends observe.

pkg/controller/config.go
```diff
@@ -421,6 +435,9 @@ func ParseFlags() (*Configuration, error) {
 	if config.EnableLbSvc && !config.EnableLb {
 		klog.Warning("--enable-lb-svc requires --enable-lb, the loadbalancer service feature will not work")
 	}
+	if config.EnableVpcEndpoint && !config.EnableLb {
```
zhangzujian: The feature requires `--enable-lb`, but this only logs a warning and continues. Controllers can still start with `--enable-vpc-endpoint=true` and then fail later in confusing ways.

Please fail closed when the feature is on and LB is off.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,1234 @@
...
+	nad := &nadv1.NetworkAttachmentDefinition{
+		Name:      subnetName,
+		Namespace: c.config.PodNamespace,
+		Spec:      nadv1.NetworkAttachmentDefinitionSpec{Config: string(buf)},
+	}
+	_, err = c.config.AttachNetClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(c.config.PodNamespace).Get(context.Background(), subnetName, metav1.GetOptions{})
+	if k8serrors.IsNotFound(err) {
+		if _, err = c.config.AttachNetClient.K8sCniCncfIoV1().NetworkAttachmentDefinitions(c.config.PodNamespace).Create(context.Background(), nad, metav1.CreateOptions{}); err != nil {
+			if k8serrors.IsForbidden(err) || k8serrors.IsAlreadyExists(err) {
+				klog.Warningf("transit NAD %s/%s create skipped: %v (create it manually if missing)", c.config.PodNamespace, subnetName, err)
```
zhangzujian: Skipping `Forbidden` (and continuing) means a cluster without Multus/NAD permission can still mark the feature as initialized. Stitcher pods then have no second NIC.

Please return this error so reconcile stays NotReady until the NAD exists.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,1234 @@
...
+	if err := validateVpcEndpointServiceImmutability(eps); err != nil {
+		return err
+	}
+	if _, err := c.vpcsLister.Get(eps.Spec.Vpc); err != nil {
+		return fmt.Errorf("get provider vpc %s: %w", eps.Spec.Vpc, err)
+	}
+	if err := c.ensureVpcEndpointTransitNetwork(); err != nil {
+		return err
+	}
+	if err := c.ensureVpcEndpointStitcherConfigMap(); err != nil {
```
zhangzujian, line 373: `CODE_STYLE.md` requires unexpected errors to be logged and returned. A warning here lets reconcile proceed without the stitcher script ConfigMap, so later exec fails with a worse error.

Please return the ensure error (same at the VEP path).

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,1234 @@
...
+		return fmt.Errorf("exec stitcher on %s/%s: %w, stdout=%s stderr=%s", pod.Namespace, pod.Name, err, stdOut, errOut)
+	}
+	klog.V(3).Infof("stitcher %s/%s: %s %s", pod.Namespace, pod.Name, stdOut, errOut)
+	return nil
+}
+
+func (c *Controller) cleanupLegacyVpcEndpointServiceOVN(eps *kubeovnv1.VpcEndpointService) {
+	for _, protocol := range []string{"tcp", "udp", "sctp"} {
+		lbName := vpcEndpointServiceLBName(eps.Name, protocol)
+		_ = c.OVNNbClient.LogicalRouterUpdateLoadBalancers(eps.Spec.Vpc, ovsdb.MutateOperationDelete, lbName)
```
zhangzujian: These `_ =` calls drop OVN cleanup errors. `CODE_STYLE.md` says unexpected errors must be logged and returned, not skipped silently.

If this is best-effort leftover cleanup from the old LB datapath, log the failures; also delete leftover SNATs if GC still treats them as expected objects.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,1234 @@
...
+	}
+	return c.patchVpcEndpointStatus(ep, true, "")
+}
+
+func validateVpcEndpointImmutability(ep *kubeovnv1.VpcEndpoint) error {
+	if ep.Labels == nil {
+		return nil
+	}
+	if vpc := ep.Labels[util.VpcEndpointVpcLabel]; vpc != "" && vpc != ep.Spec.Vpc {
+		return fmt.Errorf("vpc is immutable after creation (was %s)", vpc)
```
zhangzujian, line 1076: CRD CEL marks `spec.subnet` (and `spec.ip`) immutable, but this controller check only compares vpc/endpointService labels. Please also enforce subnet (and ip, if kept) here so a CEL-bypass or older apiserver cannot move an endpoint to another subnet.

pkg/controller/gc.go
```diff
@@ -1471,3 +1473,235 @@ func logicalRouterPortFilter(exceptPeerPorts *strset.Set) func(lrp *ovnnb.Logica
...
```
zhangzujian: `gcVpcEndpoint` is ~231 lines (`CODE_STYLE.md` limit is 100) and still builds `expectedLBs` / `expectedSnat` for the pre-stitcher OVN LB/NAT datapath.

Please split this function and stop treating OVN LBs/SNATs as live objects unless they are actually created on this path.

pkg/ovs/ovn-nb-nat.go
```diff
@@ -130,6 +131,80 @@ func (c *OVNNbClient) CreateNats(lrName string, nats ...*ovnnb.NAT) error {
 	return nil
 }
 
+// AddSnatWithMatch creates an SNAT rule that only matches traffic selected by
+// match (for example `ip4.dst == <transitVIP>`). Multiple rules may share the
+// same externalIP/logicalIP pair when their matches differ; an existing rule is
+// never overwritten. Construction bypasses newNat, which treats
+// (externalIP, logicalIP) as a unique SNAT key without considering Match.
+func (c *OVNNbClient) AddSnatWithMatch(lrName, externalIP, logicalIP, match string) error {
```
zhangzujian: This helper (and `SetLoadBalancerNeighborResponder`) looks leftover from the OVN LR NAT/LB design. The current feature uses in-pod iptables and should not need new NAT uniqueness semantics.

Please drop the unused helpers, or show a remaining caller on the stitcher path.

zhangzujian (changes requested):
Request changes.

The stitcher redesign is closer to the documented PrivateLink-style path, but several contract and lifecycle issues remain:

* IPv6/dual: `install.sh` still provisions IPv6/dual transit CIDRs, but the stitcher only programs `iptables` / `net.ipv4`. Dual-stack clusters would get a transit network the datapath cannot use.
* API/docs still describe TransitVIP as an OVN LB VIP and `status.snatIP` as a shared consumer SNAT IP. After the stitcher change, NAT/LB is in-pod iptables; backends see the provider stitcher VPC-leg IP; `snatIP` is the per-pod transit address.
* `--enable-lb` and Multus are required, but a missing LB is only a warning and a Forbidden NAD create is skipped. That is not fail-closed.
* Legacy OVN LB/NAT helpers (`AddSnatWithMatch`, `SetLoadBalancerNeighborResponder`) and GC expected-LB/SNAT maps look leftover from the previous datapath. `cleanupLegacy*` also swallows OVN errors with `_ =`.
* `CODE_STYLE.md`: `gcVpcEndpoint` is ~231 lines and `reconcileVpcEndpoint` is over 100. Unexpected errors must be logged and returned, not dropped.
* Controller immutability checks omit `spec.subnet` even though CRD CEL marks it immutable.

Please also look at the trusted x86 `required-gate` failures on this SHA (Conformance dual/ipv6, IC ipv4, Submariner, MetalLB).

dist/images/vpcnatgateway/vpc-endpoint-stitcher.sh
metsoft68: Addressed in 6b5f5d77: kept IPv4-only stitcher (iptables). Controller now rejects IPv6/dual `--vpc-endpoint-transit-cidr`; install.sh no longer advertises IPv6 transit CIDRs; docs/CRD comments call out IPv4-only.

dist/images/install.sh
metsoft68: Addressed in 6b5f5d77: `install.sh` keeps an IPv4 transit CIDR for ipv6/dual installs and documents that VPC endpoints are IPv4-only until a dual-stack stitcher exists.

pkg/apis/kubeovn/v1/vpc-endpoint.go
metsoft68: Addressed in 6b5f5d77: CRD comments regenerated — TransitVIP is the provider stitcher Multus transit address used by in-pod iptables DNAT, not an OVN LB VIP. `Mac` marked deprecated/unused.

pkg/apis/kubeovn/v1/vpc-endpoint.go
metsoft68: Addressed in 6b5f5d77: `status.snatIP` comment rewritten — it is the consumer stitcher transit-leg IP; backends see the provider stitcher VPC-leg IP after MASQUERADE.

pkg/controller/config.go
metsoft68: Addressed in 6b5f5d77: `--enable-vpc-endpoint=true` without `--enable-lb` now fails config parse (hard error).

pkg/controller/vpc_endpoint.go
metsoft68: Addressed in 6b5f5d77: Forbidden NAD create is returned as an error so init/reconcile stay fail-closed until Multus/RBAC allow the transit NAD.

pkg/controller/vpc_endpoint.go
metsoft68: Addressed in 6b5f5d77: stitcher ConfigMap ensure errors are logged and returned on both VES and VEP reconcile paths.

pkg/controller/vpc_endpoint.go
metsoft68: Addressed in 6b5f5d77: `cleanupLegacy*` now logs and returns OVN cleanup errors instead of `_ =`.

pkg/controller/vpc_endpoint.go
metsoft68: Addressed in 6b5f5d77: controller immutability now also enforces `spec.subnet` and `spec.ip` via labels (alongside CEL).

pkg/controller/gc.go
metsoft68: Addressed in 6b5f5d77: `gcVpcEndpoint` split into smaller helpers; OVN LBs/SNATs/Vip CRs are treated only as leftover upgrade garbage (no expected live maps for the stitcher path).

pkg/ovs/ovn-nb-nat.go
metsoft68: Addressed in 6b5f5d77: removed unused `AddSnatWithMatch` and `SetLoadBalancerNeighborResponder`. Kept `DeleteSnatWithMatch` for leftover SNAT GC only.

metsoft68:
## Follow-up for @zhangzujian review on `a1f53e56`

Pushed `6b5f5d77` addressing the Sept 10 stitcher contract / lifecycle review:

| Ask | Change |
| --- | --- |
| IPv6/dual transit CIDR | **Rejected**: IPv4-only stitcher; controller fails if transit CIDR is IPv6/dual; `install.sh` keeps IPv4 transit on ipv6/dual installs |
| CRD TransitVIP / snatIP docs | Updated + `make gen-crd` (stitcher Multus VIP; snatIP ≠ backend-observed source) |
| Fail-closed `--enable-lb` | Hard error when `--enable-vpc-endpoint` without `--enable-lb` |
| Forbidden NAD / stitcher CM | Errors returned (no warning-and-continue) |
| Legacy `_ =` OVN cleanup | Logged + returned |
| `spec.subnet` / `spec.ip` immutability | Enforced in controller via labels |
| Split `gcVpcEndpoint` / drop live LB+SNAT maps | Split helpers; OVN LB/SNAT/Vip treated as leftover upgrade GC only |
| Unused `AddSnatWithMatch` / `SetLoadBalancerNeighborResponder` | Removed (kept `DeleteSnatWithMatch` for leftover GC) |

Also reject IPv6-only consumer subnets / Spec.IP and filter non-IPv4 provider backends.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,1329 @@
...
+}
+
+// validateVpcEndpointIPv4 rejects IPv6-only consumer subnets and IPv6 Spec.IP.
+// The stitcher datapath programs iptables only.
+func validateVpcEndpointIPv4(ep *kubeovnv1.VpcEndpoint, subnet *kubeovnv1.Subnet) error {
+	proto := subnet.Spec.Protocol
+	if proto == "" {
+		proto = util.CheckProtocol(subnet.Spec.CIDRBlock)
+	}
+	if proto == kubeovnv1.ProtocolIPv6 {
```
zhangzujian, line 1004: [P1] Reject or select IPv4 for dual-stack consumer subnets. `validateVpcEndpointIPv4` rejects only `ProtocolIPv6`, so a `ProtocolDual` subnet is accepted. However, dynamically allocated pod IP annotations use the `v4,v6` form; `vpcEndpointStitcherIPs` only strips CIDR masks and returns the whole string, and the later `CheckProtocol(...) != ProtocolIPv4` check then rejects it. Every VpcEndpoint without an explicit IP on a dual-stack consumer subnet therefore remains NotReady. Either reject `ProtocolDual` here or split/select the IPv4 address before programming the stitcher.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,1329 @@
...
+	client := c.config.KubeClient.CoreV1().ConfigMaps(namespace)
+	existing, err := client.Get(context.Background(), vpcEndpointStitcherCMName, metav1.GetOptions{})
+	if k8serrors.IsNotFound(err) {
+		cm := &corev1.ConfigMap{
+			Name:      vpcEndpointStitcherCMName,
+			Namespace: namespace,
+			Data:      src.Data,
+		}
+		if owner != nil {
+			if err := util.SetOwnerReference(owner, cm); err != nil {
```
zhangzujian: [P2] Do not rely on this ownerRef for tenant ConfigMap cleanup. VpcEndpoint/VpcEndpointService are cluster-scoped CRs, while this dependent ConfigMap is namespaced; Kubernetes garbage collection does not honor a namespaced dependent's ownerRef to a cluster-scoped owner. If a CR is force-deleted or its finalizer is bypassed, `gcVpcEndpointStitcherDeployments` can remove the orphan Deployment but never removes this ConfigMap, leaving the tenant copy behind. Remove tenant ConfigMaps explicitly from the GC path (or use a valid namespaced lifecycle owner).

zhangzujian (changes requested):
重新审查当前 HEAD 后仍有两个生命周期/功能问题，详见内联意见。

pkg/controller/vpc_endpoint.go
metsoft68: Addressed in `950547c0`: dual-stack consumer subnets stay allowed; `vpcEndpointSelectIPv4` / `vpcEndpointStitcherIPs` now pick the IPv4 address from `v4,v6` annotations (and Multus network-status) before programming the stitcher.

pkg/controller/vpc_endpoint.go
metsoft68: Addressed in `950547c0`: stopped setting ownerRef from cluster-scoped VES/VEP onto namespaced ConfigMaps, and `gcVpcEndpointStitcherConfigMaps` now deletes orphaned tenant `vpc-endpoint-stitcher` ConfigMaps when no live VES/VEP references that namespace.

metsoft68:
## Follow-up for @zhangzujian review on `6b5f5d77`

Pushed `950547c0`:

| Ask | Change |
| --- | --- |
| Dual-stack consumer LocalVIP | Select IPv4 from `v4,v6` pod/Multus annotations before stitcher programming |
| Tenant ConfigMap ownerRef/GC | No cluster→namespaced ownerRef; explicit `gcVpcEndpointStitcherConfigMaps` |

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,1340 @@
...
+func vpcEndpointServiceDeployName(name string) string {
+	return "vpc-eps-" + name
+}
+
+func vpcEndpointDeployName(name string) string {
+	return "vpc-ep-" + name
+}
+
+func vpcEndpointTransitProvider() string {
+	return fmt.Sprintf("%s.%s.ovn", util.DefaultVpcEndpointTransitSwitch, metav1.NamespaceSystem)
```
zhangzujian: [P1] Derive this provider from the configured transit NAD name. `--vpc-endpoint-transit-switch` is advertised as configurable, and the NAD is created with that name, but this helper always returns `vpc-endpoint-transit.kube-system.ovn`. The pod controller derives an OVN attachment provider as `<nad>.<namespace>.ovn`; with `--vpc-endpoint-transit-switch=custom`, it therefore looks for `custom.kube-system.ovn`, while this code creates the Subnet and annotations under the default provider. The stitcher Pod then fails with `provider custom.kube-system.ovn is not bound to any subnet`. Please use `c.config.VpcEndpointTransitSwitch` and the actual NAD namespace consistently, and cover a non-default switch.

pkg/controller/gc.go
```diff
@@ -1471,3 +1473,314 @@ func logicalRouterPortFilter(exceptPeerPorts *strset.Set) func(lrp *ovnnb.Logica
...
+		ns, err := c.vpcEndpointConsumerNamespace(ep.Spec.Vpc)
+		if err != nil {
+			continue
+		}
+		expectedNS[ns] = struct{}{}
+	}
+
+	var cms []*corev1.ConfigMap
+	var err error
+	if c.configMapsLister != nil {
```
zhangzujian: [P1] This branch prevents tenant ConfigMap GC in production. `configMapsLister` is always wired from `cmInformerFactory`, which is created with `WithNamespace(config.PodNamespace)` in `controller.go`; its `List` can only return controller-namespace ConfigMaps. Consequently no tenant `vpc-endpoint-stitcher` ConfigMap ever reaches the deletion loop. The new test leaves `configMapsLister` nil and exercises only the direct-client fallback, so it misses the real path. Use a cluster-wide list (or a separate all-namespace informer) here and add a test with the production-style lister scope.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,1340 @@
...
+func (c *Controller) vpcEndpointProviderMappings(svc *corev1.Service) ([]string, error) {
+	endpointSlices, err := c.endpointSlicesLister.EndpointSlices(svc.Namespace).List(labels.Set{
+		discoveryv1.LabelServiceName: svc.Name,
+	}.AsSelector())
+	if err != nil {
+		return nil, err
+	}
+	// getEndpointBackend filters addresses by protocol of serviceIP; use ClusterIP
+	// (or a v4 placeholder) so empty string does not drop all backends.
+	serviceIP := svc.Spec.ClusterIP
```
zhangzujian: [P1] Select IPv4 before asking `getEndpointBackend` to filter EndpointSlices. For an IPv6-first dual-stack Service, `Spec.ClusterIP` is IPv6, so `getEndpointBackend` returns only IPv6 addresses; the loop below then removes all of them because this datapath is IPv4-only. The VES stays NotReady with `no ready backends` even though the Service has IPv4 backends. Use an IPv4 ClusterIP from `Spec.ClusterIPs` (or an IPv4 placeholder, since only the address family is used) and add IPv6-first dual-stack coverage.

pkg/controller/vpc_endpoint.go
```diff
@@ -0,0 +1,1340 @@
...
+	if c.config.VpcEndpointTransitSwitch != "" {
+		if err := c.OVNNbClient.UpdateVpcEndpointServiceACLs(c.config.VpcEndpointTransitSwitch, eps.Name, "", nil); err != nil {
+			return fmt.Errorf("clear transit ACLs for %s: %w", eps.Name, err)
+		}
+	}
+	if err := c.deleteVpcEndpointStitcherConfigMapIfUnused(eps.Spec.Namespace); err != nil {
+		return err
+	}
+	eps.Status.TransitVIP = ""
+	eps.Status.Mac = ""
```
zhangzujian, line 900: [P1] The current trusted `Build kube-ovn` job fails here with SA1019 because `VpcEndpointServiceStatus.Mac` is now documented with the Go `Deprecated:` directive. The same diagnostic fires in two unit-test references. Please make the compatibility cleanup/test lint-approved (or express API deprecation without marking internal required references as forbidden) so the required build passes: https://github.com/kubeovn/kube-ovn/actions/runs/34707956474/job/103591367884

zhangzujian (changes requested):
Re-reviewed the current HEAD. The IPv4 selection for dual-stack pod annotations is fixed and its focused tests pass, but the tenant ConfigMap GC does not run against tenant namespaces in production. I also found two IPv4/configuration paths that remain broken, and the current trusted build fails on the new deprecation diagnostics.

pkg/controller/vpc_endpoint.go
metsoft68: Fixed. `vpcEndpointTransitProvider()` now derives `<switch>.<PodNamespace>.ovn` from `--vpc-endpoint-transit-switch` and the controller Pod namespace (with defaults), matching the pod-controller Multus/OVN provider naming. Added coverage for a non-default switch/namespace.

pkg/controller/gc.go
metsoft68: Fixed. GC always lists cluster-wide via the API client with `metadata.name=vpc-endpoint-stitcher` and no longer uses the PodNamespace-scoped `configMapsLister`. The unit test now wires a production-style namespaced lister and still asserts the orphan tenant CM is deleted through the client path.

pkg/controller/vpc_endpoint.go
metsoft68: Fixed. `vpcEndpointProviderMappings` now picks an IPv4 address via `vpcEndpointServiceIPv4` (prefer `Spec.ClusterIPs`, then IPv4 `Spec.ClusterIP`, else `0.0.0.0`) before calling `getEndpointBackend`. Added IPv6-first dual-stack Service/EndpointSlice coverage.

pkg/controller/vpc_endpoint.go
metsoft68: Fixed. Removed the Go `Deprecated:` directive from `VpcEndpointServiceStatus.Mac` (kept a plain compatibility note) so the trusted Build kube-ovn job no longer fails SA1019 on required references.

metsoft68:
Addressed the four P1s from the Sep 13 review in `cbe3abde`:

1. Transit Multus/OVN provider now follows `--vpc-endpoint-transit-switch` + PodNamespace.
2. Tenant stitcher ConfigMap GC always lists cluster-wide (ignores namespaced lister).
3. Provider mappings select an IPv4 ClusterIP before EndpointSlice filtering.
4. Dropped `Deprecated:` on status.Mac to clear SA1019.

Compile-checked with `GOOS=linux CGO_ENABLED=0 go build ./cmd/controller` and `go test -c ./pkg/controller`.

pkg/controller/vpc_endpoint_test.go
```diff
@@ -0,0 +1,1285 @@
...
+		ClusterIP:  "fd00:10:96::a",
+		ClusterIPs: []string{"fd00:10:96::a"},
+	}}))
+}
+
+func TestVpcEndpointProviderMappingsIPv6FirstDualStack(t *testing.T) {
+	tcp := corev1.ProtocolTCP
+	portNum := int32(80)
+	slice := &discoveryv1.EndpointSlice{
+		Name:      "svc-a",
```
zhangzujian: [P1] Run gofumpt on this new test literal

The trusted Build kube-ovn job fails here with File is not properly formatted (gofumpt). Locally, gofumpt -d pkg/controller/vpc_endpoint_test.go aligns Name, Namespace, and Labels with the longer AddressType field. Please apply that formatting so the required build can pass.

zhangzujian (changes requested):
Re-reviewed the current HEAD. The four P1 findings from the previous review are addressed, and the focused VPC endpoint controller tests pass locally. One reproducible standards/CI blocker remains in the new regression test.

pkg/controller/vpc_endpoint_test.go
metsoft68: Fixed in `62794240`. Ran gofumpt on `TestVpcEndpointProviderMappingsIPv6FirstDualStack` so Name/Namespace/Labels align with AddressType.

metsoft68:
Addressed the Sep 20 P1 (gofumpt on the IPv6-first dual-stack regression test) in `62794240`.

zhangzujian (changes requested):
Request changes.

Re-reviewed current HEAD `627942402e0ddfe7a2651114d85950390e93140e`. The focused controller tests pass locally, but the following issues remain:

1. **Provider subnet ownership is not validated.** `reconcileVpcEndpointService` validates the Service VPC annotation, then `vpcEndpointProviderSubnet` returns the subnet named by the Service logical-switch annotation without checking `subnet.Spec.Vpc == eps.Spec.Vpc`. A malformed or malicious Service annotation can attach the provider stitcher to a subnet in another VPC, breaking tenant isolation. See `pkg/controller/vpc_endpoint.go:400` and `:462-466`.

2. **Stale provider iptables rules survive backend exhaustion.** When `vpcEndpointProviderMappings` returns `no ready backends`, reconciliation exits before `provider-sync`. The error path only patches `Ready=false`; it does not deactivate the provider stitcher or flush its rules. The old TransitVIP can therefore continue forwarding to deleted or stale EndpointSlice backends. See `pkg/controller/vpc_endpoint.go:362-366` and `:437-442`.

3. **The install script hard-codes the transit configuration.** `dist/images/install.sh` assigns `VPC_ENDPOINT_TRANSIT_SWITCH` and `VPC_ENDPOINT_TRANSIT_CIDR` as constants instead of honoring environment overrides. The install path therefore cannot configure the transit switch/CIDR promised by the PR contract. See `dist/images/install.sh:138-141`.

4. **Transit CIDR changes are silently ignored for an existing subnet.** `ensureVpcEndpointTransitNetwork` only repairs `Vpc` and `Provider` when the transit subnet already exists; it does not validate or reconcile `CIDRBlock`, gateway, or protocol. Changing `--vpc-endpoint-transit-cidr` leaves the actual fabric on the old CIDR. See `pkg/controller/vpc_endpoint.go:244-272`.

5. **The required Build kube-ovn check is still failing.** The trusted build fails modernize on `pkg/controller/vpc_endpoint.go:759-760` (`strings.Split` should use `strings.Cut`, and the range should use `strings.SplitSeq`). Coveralls also fails because coverage drops by 0.003%.

Please address these issues and rerun the required checks before merge.

zhangzujian:
One more thing, please resolve the conflicts.

PR #7281, opened 2026-08-20
ci: skip the x86 E2E gate when the PR is already merged

zhangzujian:
The unprivileged `pull_request` Build x86 Image for #7280 completed after the PR was merged:

https://github.com/kubeovn/kube-ovn/actions/runs/32360797816/job/96399743984

`workflow_run.pull_requests` was empty, the fallback listed `state=open` PRs for `zhangzujian:fix/e2e-pr-check-run-input`, and jq failed with:

```
workflow HEAD must resolve to exactly one open pull request
```

Zero matching open PRs means the source PR is already gone; skip the gate. Still fail when more than one open PR matches the same HEAD.

Signed-off-by: Zujian Zhang <zhangzujian.7@gmail.com>

### DIFF
--- a/.github/workflows/build-x86-image.yaml
+++ b/.github/workflows/build-x86-image.yaml
@@ -24,13 +24,6 @@ defaults:
       shutil.rmtree(shadowDir); sys.exit(result.returncode)" {0}
 
 on:
-  pull_request:
-    branches:
-    - master
-    - release-*
-    paths-ignore:
-    - 'docs/**'
-    - '**.md'
   push:
     branches:
     - master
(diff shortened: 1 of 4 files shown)


zhangzujian:
Confirmed the duplicate entry point: `Build x86 Image` was triggered by both `pull_request` and the trusted automatic `workflow_dispatch` from `x86-e2e-dispatcher` (for example, runs `32361530045` and `32361560546` for the same PR HEAD). The dispatch executor is required for private Kind images and isolated PR-HEAD E2E execution, so this change removes only the native `pull_request` trigger from `build-x86-image.yaml`; `push` and `workflow_dispatch` remain. Added a contract test covering the trigger set.

Validation: 95 control-module tests passed; `git diff --check` passed. Commit: `776190df48925edaea8038240fc4ca859af4d3ec`.

PR #7291, opened 2026-08-21
fix: delete vip mapping from VPC load balancers on SwitchLBRule deletion

### LINKED ISSUE #6392: [BUG] ls-lb exist even SLR deleted
### Kube-OVN Version

v1.15.4

### Kubernetes Version

v1.33.5

4.18.0-513.5.1.el8_9.x86_64

### Description

When I delete slr, the vip and loabalancer health check is deleted but ls-lb still exist.

```
apiVersion: kubeovn.io/v1
kind: SwitchLBRule
metadata:
  labels:
    ovn.kubernetes.io/vpc-dns: "true"
  name: test-vpc3-dns
spec:
  namespace: cno-vpc-test-vpc3
  ports:
  - name: dns
    port: 53
    protocol: UDP
  - name: dns-tcp
    port: 53
    protocol: TCP
  - name: metrics
    port: 9153
    protocol: TCP
  selector:
  - k8s-app:vpc-dns-test-vpc3
  vip: 172.19.0.10
```

```
k ko nbctl list load_balancer
...
_uuid               : f060d04a-7d31-4817-ae1d-48fe0882647f
external_ids        : {vendor=kube-ovn}
health_check        : []
ip_port_mappings    : {}
name                : vpc-test-vpc3-udp-load
options             : {prefer_local_backend="false"}
protocol            : udp
selection_fields    : []
vips                : {"172.19.0.10:53"="172.22.0.2:53,172.22.0.3:53"}

_uuid               : 2c6526fc-4187-4834-af58-a35196aa498c
external_ids        : {vendor=kube-ovn}
health_check        : []
ip_port_mappings    : {"172.22.0.2"="vpc-dns-test-vpc3-868957d955-g6nhv.cno-vpc-test-vpc3:172.22.0.4", "172.22.0.3"="vpc-dns-test-vpc3-868957d955-jrdp6.cno-vpc-test-vpc3:172.22.0.4"}
name                : vpc-test-vpc3-tcp-load
options             : {prefer_local_backend="false"}
protocol            : tcp
selection_fields    : []
vips                : {"172.19.0.10:53"="172.22.0.2:53,172.22.0.3:53", "172.19.0.10:9153"="172.22.0.2:9153,172.22.0.3:9153"}

# k ko nbctl ls-lb-list test-vpc3default
UUID                                    LB                  PROTO      VIP                 IPs
2c6526fc-4187-4834-af58-a35196aa498c    vpc-test-vpc3-tc    tcp        172.19.0.10:53      172.22.0.2:53,172.22.0.3:53
                                                            tcp        172.19.0.10:9153    172.22.0.2:9153,172.22.0.3:9153
f060d04a-7d31-4817-ae1d-48fe0882647f    vpc-test-vpc3-ud    udp        172.19.0.10:53      172.22.0.2:53,172.22.0.3:53
```

### Steps To Reproduce

1. create vpc and subnets
2. create slr
3. create deploy
4. delete slr

### Current Behavior

only the vip and lb health check deleted

### Expected Behavior

vpc loadbalancer should be deleted

pujitha24:
# Pull Request

## What type of this PR

- Bug fixes

Fixes #(issue-number)

## Details

When a `SwitchLBRule` (SLR) is deleted, `handleDelSwitchLBRule` cleans up the
associated `LoadBalancerHealthCheck`, IP-port mappings, and the Kubernetes-side
VIP custom resource, but it never removed the actual `vip -> backends` entry
from the SLR's VPC OVN load balancers (TCP/UDP/SCTP and their session-affinity
variants). That entry was previously only removed indirectly and
asynchronously, once the SLR's underlying headless Service object was deleted
and picked up by the separate `handleDeleteService` path in
`pkg/controller/service.go`.

This adds a direct, synchronous `LoadBalancerDeleteVip` call for each of the
SLR's own VPC load balancers and each of the SLR's vips, scoped by the
already-computed `vpcLBNames` set so it cannot touch another VPC's load
balancer. The call is idempotent (`LoadBalancerDeleteVip` is a no-op if the
vip isn't present on that particular load balancer), so it is additive and
does not change behavior when the vip has already been cleaned up via the
existing service-delete path.

Note: `make lint` mirrors this repo's CI-pinned golangci-lint (v2.13.1); the
x86 E2E gate workflow is currently red on `master` for unrelated reasons
(`Update x86 E2E gate`), independent of this change.

CI on this repo requires a maintainer to add `ok-to-test` — happy to address
anything it surfaces.

Fixes #6392

### DIFF
--- a/pkg/controller/switch_lb_rule.go
+++ b/pkg/controller/switch_lb_rule.go
@@ -378,6 +378,21 @@ func (c *Controller) handleDelSwitchLBRule(info *SwitchLBRuleInfo) error {
 		}
 	}
 
+	// The LBHC-driven cleanup above only removes health-check-related state; the
+	// actual vip->backends mapping on the VPC load balancers is otherwise only
+	// removed asynchronously once the backing service delete event reaches
+	// handleDeleteService. Delete it here too so the load balancer mapping is
+	// gone as soon as the SLR is deleted, rather than depending on that separate
+	// event to arrive and be processed.
+	for lbName := range vpcLBNames {
+		for _, vip := range info.Vips {
+			if err = c.OVNNbClient.LoadBalancerDeleteVip(lbName, vip, true); err != nil && !k8serrors.IsNotFound(err) {
+				klog.Errorf("failed to delete vip %s from load balancer %s, err: %v", vip, lbName, err)
+				return err
+			}
+		}
+	}
+
 	return nil
 }
 
--- a/pkg/controller/switch_lb_rule.go
+++ b/pkg/controller/switch_lb_rule.go
@@ -381,9 +381,11 @@ func (c *Controller) handleDelSwitchLBRule(info *SwitchLBRuleInfo) error {
 	// The LBHC-driven cleanup above only removes health-check-related state; the
 	// actual vip->backends mapping on the VPC load balancers is otherwise only
 	// removed asynchronously once the backing service delete event reaches
-	// handleDeleteService. Delete it here too so the load balancer mapping is
-	// gone as soon as the SLR is deleted, rather than depending on that separate
-	// event to arrive and be processed.
+	// handleDeleteService. When the VPC is known (vpcLBNames != nil), delete the
+	// mapping here too so it is gone as soon as the SLR is deleted, rather than
+	// depending on that separate event to arrive and be processed. If the VPC is
+	// unknown, vpcLBNames is nil and this loop is a no-op, falling back to the
+	// asynchronous handleDeleteService cleanup.
 	for lbName := range vpcLBNames {
 		for _, vip := range info.Vips {
 			if err = c.OVNNbClient.LoadBalancerDeleteVip(lbName, vip, true); err != nil && !k8serrors.IsNotFound(err) {
--- a/pkg/controller/switch_lb_rule_test.go
+++ b/pkg/controller/switch_lb_rule_test.go
@@ -244,6 +244,8 @@ func Test_handleDelSwitchLBRule(t *testing.T) {
 		fc.mockOvnClient.EXPECT().DeleteLoadBalancerHealthChecks(gomock.Any()).Return(nil)
 		// Second call: after LBHC deletion, no more LBHCs for this subnet
 		fc.mockOvnClient.EXPECT().ListLoadBalancerHealthChecks(gomock.Any()).Return([]ovnnb.LoadBalancerHealthCheck{}, nil)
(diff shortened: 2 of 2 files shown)


pujitha24:
Good catch, the comment did read as if this always ran synchronously. Updated it to spell out that this only happens when `vpcLBNames` is non-nil (VPC known), and that when it's nil the loop is a no-op and we fall back to the async `handleDeleteService` cleanup as before. Didn't add an explicit `vpcLBNames != nil` guard since `vpcLBNames` is a `set.Set[string]` (a map under the hood), so ranging over nil is already a safe no-op — the guard would just be redundant.

zhangzujian:
Status update after comparing the current branches:

- `master` (HEAD `53d3fceb3b1`) no longer has the original stale VIP-mapping path. `handleDelSwitchLBRule` calls `deleteLegacyVpcVIPs(vpcForSlr, info.Vips)`, which removes each VIP from the VPC legacy load balancers. For the newer service-scoped load balancers, it calls `deleteServiceScopedLoadBalancers(ownerSvc)`, which removes the SLR-owned load balancers and their mappings. See [switch_lb_rule.go](https://github.com/kubeovn/kube-ovn/blob/53d3fceb3b135b8aa8965f46cee4f1ce85d1a42d/pkg/controller/switch_lb_rule.go#L266-L268) and [service_scoped_lb.go](https://github.com/kubeovn/kube-ovn/blob/53d3fceb3b135b8aa8965f46cee4f1ce85d1a42d/pkg/controller/service_scoped_lb.go#L788-L817).

- `release-1.16` (HEAD `7038dd682cc3`) still has the older implementation. It scopes LBHC handling with `vpcLBNames`, but does not call `LoadBalancerDeleteVip`, so the original stale `vip -> backends` mapping can still remain after SLR deletion. See [switch_lb_rule.go](https://github.com/kubeovn/kube-ovn/blob/7038dd682cc373a72c53bc9b11e1111ce621e312/pkg/controller/switch_lb_rule.go#L238-L259).

Therefore, the original issue is covered on current `master`, while an equivalent fix/backport is still needed for `release-1.16`.

pujitha24:
Thanks for tracking that down. Confirmed on my end too — deleteLegacyVpcVIPs (added in #7378) already calls LoadBalancerDeleteVip for the SLR's vips in handleDelSwitchLBRule, so master is covered and this PR is redundant there. Closing it; a release-1.16 backport would need its own PR since that branch is still on the older path.

PR #7294, opened 2026-08-21
feat: add routed subnet mode with /32 addressing and gateway hairpin

emamihe:
## Summary

Adds a new Subnet option `spec.routed` that switches the subnet from a shared L2 broadcast domain to a routed (/32|/128) model.

When enabled (CIDR can still be something like `10.0.0.0/16`):
- Each pod gets a host address (`/32` IPv4 or `/128` IPv6), not the subnet prefix.
- The pod netns is configured with an on-link route to the gateway, then a default route via that gateway:
  ```
  ip addr add 10.0.0.5/32 dev eth0
  ip route add 10.0.0.1/32 dev eth0
  ip route add default via 10.0.0.1
  ```
- Same-subnet east-west traffic is forced through the OVN logical router (gateway hairpin), not direct L2 between pods.
- OVN ACLs allow ARP/ND only for the gateway and IP frames only to/from the logical router port MAC, so pods cannot talk peer-to-peer on the logical switch.

This improves scalability by removing the host-visible broadcast domain and making the gateway the only L2 neighbor for pods.

## Scope / constraints

- Supported on overlay subnets, and underlay only when an OVN router port exists (`logicalGateway` or `u2oInterconnection`).
- Rejected for pure underlay with a physical gateway, non-OVN providers, mac-only/BYO-DHCP, DHCP, IPv6 RA, and multicast snoop.
- IPAM is unchanged: addresses are still allocated from `spec.cidrBlock`.
- Existing pods must be recreated after enabling or disabling `routed` (CNI configures addressing at ADD time).
- Distinct from the existing pod annotation `ovn.kubernetes.io/routed` (that only means “OVN routes are ready”).

## How to try

```yaml
apiVersion: kubeovn.io/v1
kind: Subnet
metadata:
  name: routed-subnet
spec:
  cidrBlock: 10.0.0.0/16
  gateway: 10.0.0.1
  protocol: IPv4
  provider: ovn
  routed: true
```

Then create pods on that subnet and verify:
1. Pod iface address is `/32` (or `/128`).
2. On-link route to the gateway and default via gateway exist.
3. Pod-to-pod on the same subnet still works, but via the gateway (not direct L2).

## Test plan

- [ ] Unit: validation matrix for `spec.routed`
- [ ] Unit: host-mask helpers (`/32`/`/128`, dual-stack)
- [ ] Unit: OVN ACL builder for routed mode
- [ ] Manual: create overlay subnet with `routed: true`, schedule two pods, confirm `/32` + gateway routes and same-subnet connectivity via gateway
- [ ] Manual: confirm underlay without logical gateway/U2O is rejected
- [ ] Manual: recreate pods after toggling `routed` and confirm addressing updates

### DIFF
--- a/charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
+++ b/charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
@@ -2807,6 +2807,14 @@ spec:
               routeTable:
                 description: Route table associated with the subnet.
                 type: string
+              routed:
+                description: |-
+                  Routed configures pods with /32 (IPv4) or /128 (IPv6) addresses and
+                  on-link routes to the gateway. Same-subnet traffic is forced through the
+                  OVN logical router instead of direct L2. Requires an OVN logical router
+                  port (overlay, or underlay with logicalGateway / u2oInterconnection).
+                  Existing pods must be recreated after enabling or disabling this field.
+                type: boolean
               u2oFeatures:
                 description: Feature configurations for underlay to overlay interconnection.
                 properties:
--- a/charts/kube-ovn/templates/kube-ovn-crd.yaml
+++ b/charts/kube-ovn/templates/kube-ovn-crd.yaml
@@ -2826,6 +2826,14 @@ spec:
               routeTable:
                 description: Route table associated with the subnet.
                 type: string
+              routed:
+                description: |-
+                  Routed configures pods with /32 (IPv4) or /128 (IPv6) addresses and
+                  on-link routes to the gateway. Same-subnet traffic is forced through the
+                  OVN logical router instead of direct L2. Requires an OVN logical router
+                  port (overlay, or underlay with logicalGateway / u2oInterconnection).
+                  Existing pods must be recreated after enabling or disabling this field.
+                type: boolean
               u2oFeatures:
                 description: Feature configurations for underlay to overlay interconnection.
                 properties:
--- a/dist/images/install.sh
+++ b/dist/images/install.sh
@@ -3219,6 +3219,14 @@ spec:
               routeTable:
                 description: Route table associated with the subnet.
                 type: string
+              routed:
+                description: |-
+                  Routed configures pods with /32 (IPv4) or /128 (IPv6) addresses and
+                  on-link routes to the gateway. Same-subnet traffic is forced through the
+                  OVN logical router instead of direct L2. Requires an OVN logical router
+                  port (overlay, or underlay with logicalGateway / u2oInterconnection).
+                  Existing pods must be recreated after enabling or disabling this field.
+                type: boolean
               u2oFeatures:
                 description: Feature configurations for underlay to overlay interconnection.
                 properties:
--- a/docs/routed-subnet.md
+++ b/docs/routed-subnet.md
@@ -15,7 +15,8 @@ ip route add default via 10.0.0.1
 Same-subnet east-west traffic is hairpinned through the OVN logical router. OVN ACLs use an allow-list / default-deny policy:
 
 - Allow ARP/ND only for the gateway
-- Allow IP frames whose Ethernet source or destination is the logical router port MAC (both ACL directions), so pod→router and router→pod hairpin traffic is permitted while direct L2 pod→pod is denied
+- Allow IP frames to the logical router port MAC (`eth.dst`) on both ACL directions (pod→router)
+- Allow IP frames from the logical router port MAC (`eth.src`) on `to-lport`, and on `from-lport` only when `inport` is the router LSP (so pods cannot spoof the LRP MAC; port security is off by default)
 - Drop all other IP/ARP/ND (both directions)
 
 When `private: true`, ingress via the router is further limited to hairpin (own CIDR), node join CIDR, and `allowSubnets`.
--- a/docs/routed-subnet.md
+++ b/docs/routed-subnet.md
@@ -1,0 +1,27 @@
+# Routed subnet mode
+
+Set `spec.routed: true` on a Subnet so pods get host routes (`/32` IPv4 or `/128` IPv6) and send all traffic through the subnet gateway instead of sharing an L2 broadcast domain.
+
+## Behavior
+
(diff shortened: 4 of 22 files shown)


zhangzujian (dismissed):
Blocking findings:\n\n1. **Routed ACLs do not enforce the advertised gateway-only policy.**  allows IP traffic to/from the router-port MAC, but it only drops IP packets whose source and destination are both inside the subnet CIDR, and only drops ARP/ND discovery targeting that CIDR. OVN ACLs default to allow when no ACL matches, so packets to other IP ranges, arbitrary L2 destinations, and ARP/ND targets outside the CIDR are still permitted. This does not implement the PR's stated contract that pods may communicate only with the gateway and that IP frames must be only to/from the logical-router-port MAC. Please add an explicit default-deny/allow-list policy for routed mode and cover both IPv4 and IPv6.\n\n2. **Private routed subnets lose their isolation.** The generic  ingress allow is priority , while the private default drop is priority . Any IP packet entering from the router port therefore matches the 1002 allow before the private drop, regardless of . Traffic from other subnets can bypass the private routed policy. The router-port allow must be constrained by the private allow-list, or the priorities/policy composition must be redesigned.\n\n3. **User-provided  can bypass routed isolation.**  calls  after installing the routed ACLs and does not reject or constrain custom ACL priorities. A custom allow ACL with a higher priority can re-enable peer/L2 traffic, defeating the feature's advertised mandatory gateway/LRP-MAC restriction. Routed mode needs to reject conflicting custom ACLs or reserve an unbypassable policy tier/priority.\n\n4. **The DPDK CNI path does not configure routed addressing or routes.** In , the DPDK branch still calls , which only creates the vhost-user port and bandwidth settings. It never applies the new /32 or /128 address, on-link gateway route, or default route handled by . The PR does not exclude DPDK providers, so routed mode is incomplete for that supported path.\n\nThe new  implementation is also about 186 lines, exceeding this repository's documented 100-line function limit, and the core controller/CNI paths lack regression tests for hairpin, private, dual-stack, and DPDK behavior.

zhangzujian (changes requested):
Blocking findings:

1. **Routed ACLs do not enforce the advertised gateway-only policy.** `SetLogicalSwitchRouted` allows IP traffic to/from the router-port MAC, but it only drops IP packets whose source and destination are both inside the subnet CIDR, and only drops ARP/ND discovery targeting that CIDR. OVN ACLs default to allow when no ACL matches, so packets to other IP ranges, arbitrary L2 destinations, and ARP/ND targets outside the CIDR are still permitted. This does not implement the PR's stated contract that pods may communicate only with the gateway and that IP frames must be only to/from the logical-router-port MAC. Please add an explicit default-deny/allow-list policy for routed mode and cover both IPv4 and IPv6.

2. **Private routed subnets lose their isolation.** The generic `eth.src == gatewayMAC` ingress allow is priority `1002`, while the private default drop is priority `1000`. Any IP packet entering from the router port therefore matches the 1002 allow before the private drop, regardless of `allowSubnets`. Traffic from other subnets can bypass the private routed policy. The router-port allow must be constrained by the private allow-list, or the priorities/policy composition must be redesigned.

3. **User-provided `spec.acls` can bypass routed isolation.** `handleAddOrUpdateSubnet` calls `UpdateLogicalSwitchACL` after installing the routed ACLs and does not reject or constrain custom ACL priorities. A custom allow ACL with a higher priority can re-enable peer/L2 traffic, defeating the feature's advertised mandatory gateway/LRP-MAC restriction. Routed mode needs to reject conflicting custom ACLs or reserve an unbypassable policy tier/priority.

4. **The DPDK CNI path does not configure routed addressing or routes.** In `handleAdd`, the DPDK branch still calls `configureDpdkNic`, which only creates the vhost-user port and bandwidth settings. It never applies the new /32 or /128 address, on-link gateway route, or default route handled by `configureContainerNic`. The PR does not exclude DPDK providers, so routed mode is incomplete for that supported path.

The new `SetLogicalSwitchRouted` implementation is also about 186 lines, exceeding this repository's documented 100-line function limit, and the core controller/CNI paths lack regression tests for hairpin, private, dual-stack, and DPDK behavior.

zhangzujian (changes requested):
Follow-up blocking finding about Pod route annotations in routed mode:

The annotation path is only partially compatible with `routed: true`. For normal non-DPDK interfaces, `handler.go` parses the provider-specific routes annotation and `configureContainerNic` attempts to install it. However:

- The routed setup adds an on-link host route only for the subnet gateway (`ovs_linux.go:527-544`). A route annotation using another gateway has no connected/on-link route for that next hop when the Pod address is /32 or /128, so the kernel can reject it.
- Errors from parsing or installing an individual annotation route are only logged and then ignored (`ovs_linux.go:560-585`). CNI ADD can therefore succeed while the annotated route is absent.
- The DPDK branch calls `configureDpdkNic` and then discards `routes` (`handler.go:487-492`), so route annotations are not applied at all for DPDK.

Please either support/validate annotation gateways (including the required on-link behavior) and return route installation failures, or explicitly reject/document unsupported combinations. DPDK should not silently discard routes in a feature that claims routed subnet support.

emamihe:
proof of concept, hope it helps :) 

image

image

image

image

image

image

image

zhangzujian (changes requested):
Blocking finding: the new loop adds `ip && eth.src == <LRP-MAC>` as an allow rule for `from-lport` (pkg/ovs/ovn-nb-acl.go:890-898). `from-lport` is evaluated when a packet enters the logical switch from a pod, so a pod can forge the router-port source MAC and match this priority-2100 allow for arbitrary IP traffic. Kube-OVN creates logical switch ports with port security disabled by default (`pkg/controller/pod.go:748-751`), therefore port security cannot be relied on to prevent this spoofing. This bypasses the routed default-deny policy and the private source-CIDR restrictions. Please only allow the LRP MAC in the direction/field combinations that represent actual router hairpin traffic (or otherwise constrain the source/inport and enforce anti-spoofing), and add a regression test for a pod-originated packet with `eth.src == <LRP-MAC>`.

zhangzujian (approved):

Issue #7325, opened 2026-08-26
[Enhancement]Support imagePullSecrets for VPC NAT Gateway StatefulSet


rrajendran17:
The VPC NAT Gateway controller creates a StatefulSet using the image configured in ovn-vpc-nat-config configmap, but there is currently no way to configure imagePullSecrets for the generated StatefulSet/Pod. This prevents the vpc-nat-gateway image from being pulled from private OCI registries that require authentication.

Add support for configuring imagePullSecrets in ovn-vpc-nat-config configmap and propagate the configured secret(s) to spec.template.spec.imagePullSecrets of the VPC NAT Gateway StatefulSet. The configuration should remain optional so existing deployments using public registries continue to work.

PR #7326, opened 2026-08-26
support image pull secrets for vpc nat gw pod

rrajendran17:
## What type of this PR

Enhancement - support imagepullsecret for vpcnatgw pod 

https://github.com/kubeovn/kube-ovn/issues/7325

Fixes #(issue-number)

Testplan:
1.Create a secret in kube-system namespace
2,Provide the secret name as imagePullSecret in ovn-vpc-nat-config  configmap

```
kubectl get configmap -n kube-system ovn-vpc-nat-config -o yaml
apiVersion: v1
data:
  image: renukaraaj17/vpc-nat-gateway:v1.16.1
  imagePullSecret: pull-secret
kind: ConfigMap
metadata:
  annotations:
    kubernetes.io/description: |
      kube-ovn vpc-nat common config
  creationTimestamp: "2026-05-21T00:15:04Z"
  name: ovn-vpc-nat-config
  namespace: kube-system
  ownerReferences:
  - apiVersion: kubeovn.io/v1
    blockOwnerDeletion: true
    controller: true
    kind: Configuration
    name: kubeovn
    uid: 205693f4-9cad-4884-b316-9508cc38ecb0
  resourceVersion: "113682777"
  uid: 1a58e840-7bbc-49ff-88e3-c5e542dc75d5

```
3. Create the vpc nat gw config 
```
 kubectl get vpc-nat-gw gw1 -o yaml
apiVersion: kubeovn.io/v1
kind: VpcNatGateway
metadata:
  annotations:
    k8s.v1.cni.cncf.io/networks: default/vswitchinternal
  creationTimestamp: "2026-08-26T22:21:25Z"
  generation: 1
  name: gw1
  resourceVersion: "113683227"
  uid: 0be81eeb-f539-4032-869d-5cb802ecf310
spec:
  externalSubnets:
  - subnetexternal
  lanIp: 172.20.10.1
  subnet: subnetinternal
  vpc: commonvpc
status:
  affinity: {}
  externalSubnets:
  - subnetexternal
  qosPolicy: ""

```

4.Verify the vpc nat gw statefulset and vpc nat gw pod in kube-system namespace reflecting the imagepullsecret and vpc nat gw pod is created successfully and running.

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -73,6 +73,7 @@ type fakeControllerInformers struct {
 	subnetInformer    kubeovninformer.SubnetInformer
 	ipInformer        kubeovninformer.IPInformer
 	vlanInformer      kubeovninformer.VlanInformer
+	configMapInformer coreinformers.ConfigMapInformer
 	serviceInformer   coreinformers.ServiceInformer
 	namespaceInformer coreinformers.NamespaceInformer
 	nodeInformer      coreinformers.NodeInformer
@@ -133,7 +134,7 @@ func newFakeControllerWithOptions(t *testing.T, opts *FakeControllerOptions) (*f
 		}}
 	}
 
-	// Create fake Kubernetes client with namespaces, pods, nodes, services, and workloads.
+	// Create fake Kubernetes client with namespaces, pods, nodes, services, workloads and config maps.
 	kubeObjects := make([]runtime.Object, 0, len(namespaces)+len(opts.Pods)+len(opts.Nodes)+len(opts.Services)+
 		len(opts.StatefulSets)+len(opts.Deployments)+len(opts.ConfigMaps))
 	for _, ns := range namespaces {
@@ -331,6 +332,7 @@ func newFakeControllerWithOptions(t *testing.T, opts *FakeControllerOptions) (*f
 	ovnSnatRuleInformer := kubeovnInformerFactory.Kubeovn().V1().OvnSnatRules()
 	qosPolicyInformer := kubeovnInformerFactory.Kubeovn().V1().QoSPolicies()
 	iptablesEipInformer := kubeovnInformerFactory.Kubeovn().V1().IptablesEIPs()
+	configMapInformer := kubeInformerFactory.Core().V1().ConfigMaps()
 
 	fakeInformers := &fakeControllerInformers{
 		vpcInformer:       vpcInformer,
@@ -342,6 +344,7 @@ func newFakeControllerWithOptions(t *testing.T, opts *FakeControllerOptions) (*f
 		namespaceInformer: namespaceInformer,
 		nodeInformer:      nodeInformer,
 		podInformer:       podInformer,
+		configMapInformer: configMapInformer,
 	}
 
 	// Create mock OVN clients
@@ -398,6 +401,9 @@ func newFakeControllerWithOptions(t *testing.T, opts *FakeControllerOptions) (*f
 		updateSubnetStatusQueue:       newTypedRateLimitingQueue[string]("UpdateSubnetStatus", nil),
 		addOrUpdateVpcNatGatewayQueue: newTypedRateLimitingQueue[string]("AddOrUpdateVpcNatGateway", nil),
 		initVpcNatGatewayQueue:        newTypedRateLimitingQueue[string]("InitVpcNatGateway", nil),
+		configMapsLister:              configMapInformer.Lister(),
+		configMapsSynced:              alwaysReady,
+		serviceCIDRStore:              util.NewServiceCIDRStore("10.96.0.0/12"),
 	}
 
 	ctrl.config = &Configuration{
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -332,7 +332,6 @@ func newFakeControllerWithOptions(t *testing.T, opts *FakeControllerOptions) (*f
 	ovnSnatRuleInformer := kubeovnInformerFactory.Kubeovn().V1().OvnSnatRules()
 	qosPolicyInformer := kubeovnInformerFactory.Kubeovn().V1().QoSPolicies()
 	iptablesEipInformer := kubeovnInformerFactory.Kubeovn().V1().IptablesEIPs()
-	configMapInformer := kubeInformerFactory.Core().V1().ConfigMaps()
 
 	fakeInformers := &fakeControllerInformers{
 		vpcInformer:       vpcInformer,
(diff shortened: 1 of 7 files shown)


rrajendran17:
@zbb88888 can you please check this ? We have a requirement where we maintain the kubeovn images in private registry and we observed vpc nat gw pod creation was failing due to authentication, so it requires a need for imagepullsecret in the vpcnatgw pod which is taken care by this PR.

zbb88888:
@rrajendran17 Hi, please fix the lint err:

image

zbb88888:
The code seems simple and ok, But you can add some UT if you like!

  ———

  Please add unit tests before merging. The PR currently has no test coverage, and the two behaviors this feature depends on (empty vs. configured imagePullSecret) should be locked down:

  1. pkg/util/vpc_nat_gateway_test.go — TestGetImagePullSecrets
      - empty string returns nil;
      - non-empty string returns exactly []corev1.LocalObjectReference{{Name: "<secret>"}};
      - decide and assert what happens with whitespace/empty-after-trim.

  2. pkg/controller/vpc_nat_gateway_test.go — generator coverage for both workload types:
      - genNatGwStatefulSet (default, replicas=1) and genNatGwDeployment (HA, replicas>1) must both be covered;
      - when vpcNatImagePullSecret == "", generated Spec.Template.Spec.ImagePullSecrets is nil;
      - when vpcNatImagePullSecret == "pull-secret", generated Spec.Template.Spec.ImagePullSecrets contains exactly one reference with that name;
      - this proves the field is correctly injected in both STS and Deployment paths, not just one.

  3. pkg/controller/vpc_nat.go — resyncVpcNatConfig coverage:
      - ConfigMap contains imagePullSecret → global vpcNatImagePullSecret is set to that value;
      - ConfigMap does not contain the key → it is reset to "";
      - ConfigMap is missing → no panic and previous behavior remains safe.

  4. Recommended: handleAddOrUpdateVpcNatGw creation-path test:
      - with a configured secret, the controller creates the STS/Deployment with ImagePullSecrets populated;
      - without one, it creates the workload with ImagePullSecrets unset.

  5. If runtime config update is intended (not just creation), also add a test proving that updating ovn-vpc-nat-config enqueues all VpcNatGateways and that a re-reconcile updates an existing
     STS/Deployment template. Note: with the current code this test would fail for the StatefulSet path, because the STS update condition only checks isVpcNatGwChanged/restart recovery and
     does not compare the regenerated template, so either the behavior should be explicitly limited to newly created gateways or the update path needs to be fixed first.

  An E2E is not strictly required if the unit tests above cover the field propagation; if one is added, creating a kubernetes.io/dockerconfigjson Secret, setting imagePullSecret in
  ovn-vpc-nat-config, creating a VpcNatGateway, and asserting both the STS/Deployment template and the running Pod would be sufficient.

rrajendran17:
> The code seems simple and ok, But you can add some UT if you like!
> 
> ———
> 
> Please add unit tests before merging. The PR currently has no test coverage, and the two behaviors this feature depends on (empty vs. configured imagePullSecret) should be locked down:
> 
>     1. pkg/util/vpc_nat_gateway_test.go — TestGetImagePullSecrets
>        
>        * empty string returns nil;
>        * non-empty string returns exactly []corev1.LocalObjectReference{{Name: ""}};
>        * decide and assert what happens with whitespace/empty-after-trim.
> 
>     2. pkg/controller/vpc_nat_gateway_test.go — generator coverage for both workload types:
>        
>        * genNatGwStatefulSet (default, replicas=1) and genNatGwDeployment (HA, replicas>1) must both be covered;
>        * when vpcNatImagePullSecret == "", generated Spec.Template.Spec.ImagePullSecrets is nil;
>        * when vpcNatImagePullSecret == "pull-secret", generated Spec.Template.Spec.ImagePullSecrets contains exactly one reference with that name;
>        * this proves the field is correctly injected in both STS and Deployment paths, not just one.
> 
>     3. pkg/controller/vpc_nat.go — resyncVpcNatConfig coverage:
>        
>        * ConfigMap contains imagePullSecret → global vpcNatImagePullSecret is set to that value;
>        * ConfigMap does not contain the key → it is reset to "";
>        * ConfigMap is missing → no panic and previous behavior remains safe.
> 
>     4. Recommended: handleAddOrUpdateVpcNatGw creation-path test:
>        
>        * with a configured secret, the controller creates the STS/Deployment with ImagePullSecrets populated;
>        * without one, it creates the workload with ImagePullSecrets unset.
> 
>     5. If runtime config update is intended (not just creation), also add a test proving that updating ovn-vpc-nat-config enqueues all VpcNatGateways and that a re-reconcile updates an existing
>        STS/Deployment template. Note: with the current code this test would fail for the StatefulSet path, because the STS update condition only checks isVpcNatGwChanged/restart recovery and
>        does not compare the regenerated template, so either the behavior should be explicitly limited to newly created gateways or the update path needs to be fixed first.
> 
> 
> An E2E is not strictly required if the unit tests above cover the field propagation; if one is added, creating a kubernetes.io/dockerconfigjson Secret, setting imagePullSecret in ovn-vpc-nat-config, creating a VpcNatGateway, and asserting both the STS/Deployment template and the running Pod would be sufficient.

@zbb88888 I have added UT cases for 1 to 4 as you mentioned above.
For 5) since configmap of vpc nat gw contains the build and imagepullsecret, and its possible user can update it, it is important we handle the case, but any change to image or imagepullsecrets requires the vpc nat gw pods to be re spawned again, so can we add a limitation to doc saying in case if the vpc nat configmap is updated, the vpc nat gw custom resource must be deleted , config map must be updated and then vpc nat gw custom resource must be created again ?
This will make sure the new vpc nat gw pods use the updated image or image pull secrets.
Do we need a webhook change for restricting this ?

Also I see some of the e2e tests are failing, looks not related to the code, can you please confirm if any action needed by me ?

zbb88888:
@rrajendran17 Hi, Thanks for the PR!

`Do we need a webhook change for restricting this ?` 

Using webhook to restrict imagepullsecret updating is not needed, becuase, if changed, pod always need restart.

webhook can not restrict it, user can  delete it and create it.

And, we will split the gw pod image and the gw script. if vpc nat gw has some new feature, you can keep the pod running, just update the gw scripts.

Restarting gw pod to update and use new feature is not always need after the splitting.

zbb88888 (approved):

zbb88888:
@rrajendran17 Please fix the UT


image

rrajendran17:
> @rrajendran17 Please fix the UT
> image

@zbb88888 Fixed the UT case errors.

zbb88888:
@rrajendran17 Hi, Please resolve the conflicts.

rrajendran17:
> @rrajendran17 Hi, Please resolve the conflicts.

@zbb88888 done, thanks

zbb88888 (approved):

rrajendran17:
@zbb88888 We would like to have this fix in v1.16.x, can this be included in your next release ? And when can I expect v1.17.x ?

I'm not sure about v1.17.x, maybe, will not that fast as your expect.

I tried, but conflicts, so make a new PR.

Could you please review it?

PR #7342, opened 2026-08-28
fix: add DNS resolution fallback in CheckProtocol() and export resolved env vars in start-db.sh

### LINKED ISSUE #7341: [BUG] CheckProtocol() fails on DNS hostnames, breaking ovn-central raft cluster with headless Service
### Kube-OVN Version

v1.17.0

### Kubernetes Version

v1.35.x

Ubuntu 22.04 / Linux 5.15+

### Description

When `ovn-central` is deployed as a StatefulSet with `POD_IP=$(DB_CLUSTER_ADDR)` where `DB_CLUSTER_ADDR` resolves to a hostname like `ovn-central-0.ovn-central.kube-system.svc`, the `kube-ovn-leader-checker` binary crashes because `CheckProtocol()` in `pkg/util/net.go` only uses `net.ParseIP()` which cannot resolve DNS hostnames.

This is particularly important for the HCP (Hosted Control Plane) deployment mode where ovn-central uses headless Service DNS names for raft clustering instead of pod IPs.

### Steps To Reproduce

1. Deploy ovn-central StatefulSet with `POD_IP=$(DB_CLUSTER_ADDR)` where `DB_CLUSTER_ADDR=$(POD_NAME).ovn-central.$(POD_NAMESPACE).svc`
2. The leader checker logs: `failed to parse ip ovn-central-0.ovn-central.kube-system.svc`
3. ovn-central pods crash-loop

### Current Behavior

`CheckProtocol()` in `pkg/util/net.go:213-222` does:

```go
ip := net.ParseIP(address)
if ip == nil {
    return "", fmt.Errorf("failed to parse ip %s", address)
}
```

`net.ParseIP()` only handles IP literals (e.g. `10.0.0.1`), not DNS hostnames. When the pod IP is set to a headless Service hostname via `$(DB_CLUSTER_ADDR)`, the function returns an error and the leader checker fails to start.

The shell script `start-db.sh` resolves the hostname via `normalize_raft_addrs`/`expand_k8s_svc_addr` but does **not** export the resolved value, so the Go binary reads the unresolved hostname from `os.Getenv("POD_IP")`.

### Expected Behavior

`CheckProtocol()` should resolve DNS hostnames via `net.LookupIP()` when `net.ParseIP()` returns nil:

```go
ip := net.ParseIP(address)
if ip == nil {
    ips, err := net.LookupIP(address)
    if err != nil || len(ips) == 0 {
        return "", fmt.Errorf("failed to resolve ip %s: %v", address, err)
    }
    ip = ips[0]
}
```

I can submit a PR for this fix.

cleman95:
## Description

In HCP mode, ovn-central intentionally supplies a headless-Service hostname as `POD_IP` (e.g. `ovn-central-0.ovn-central.kube-system.svc`), while the leader checker needs the address family to pick the EndpointSlice address type (#7341).

Per review feedback, this PR keeps the generic `CheckProtocol` validator **literal-only** and introduces an explicit, testable resolver used only by the HCP leader-checker path:

- `util.resolveProtocol` (injectable lookup) + `util.ResolveProtocol` (wrapper over `net.DefaultResolver`) in `pkg/util/net.go`
- Dual-stack behavior is defined explicitly: when a hostname resolves to both A and AAAA records, **IPv4 is preferred** — deterministic, since resolver ordering is not and callers only distinguish IPv4/IPv6
- The leader checker falls back to IPv4 with a warning when resolution fails, preserving its previous defaulting behavior
- `CheckProtocol` is unchanged: no ambient DNS lookups from IP/CIDR validation paths
- `start-db.sh`: export the shell-resolved `POD_IP`/`NODE_IPS`/`DB_CLUSTER_ADDR` after `normalize_raft_addrs`

## Changes

- `pkg/util/net.go`: add `resolveProtocol` (injectable lookup) + `ResolveProtocol`; `CheckProtocol` untouched
- `pkg/ovn_leader_checker/ovn.go`: classify `POD_IP` via `ResolveProtocol`, warn + default to IPv4 on failure
- `pkg/util/net_test.go`: `TestResolveProtocol` with fake lookup covering literal v4/v6, hostname→v4/v6, hostname→both families (asserts IPv4 determinism), lookup failure, empty result, empty input
- `dist/images/start-db.sh`: export resolved env values

## Review notes

- The container capabilities change initially bundled here was split into a separate PR (to be linked).

## Test Plan

- `go test ./pkg/util/ -run "TestCheckProtocol|TestResolveProtocol" -v` — all pass, including the previously failing invalid-input cases (`not-an-ip`, `1.2.3`)
- `go test ./pkg/ovn_leader_checker/` — all pass

Fixes #7341

### DIFF
--- a/pkg/util/net.go
+++ b/pkg/util/net.go
@@ -1,6 +1,7 @@
 package util
 
 import (
+	"context"
 	"crypto/rand"
 	"errors"
 	"fmt"
@@ -226,6 +227,63 @@ func CheckProtocol(address string) string {
 	err := fmt.Errorf("invalid address %q", address)
 	klog.Error(err)
 	return ""
+}
+
+// resolveProtocol classifies the given address into kubeovnv1.ProtocolIPv4 or
+// kubeovnv1.ProtocolIPv6. The address may be a literal IP address or a
+// hostname, in which case lookup is used to resolve it. lookup is injectable
+// so tests can provide deterministic results.
+//
+// When a hostname resolves to addresses of both address families, IPv4 is
+// preferred: the choice must be deterministic because resolver ordering is
+// not, and callers (e.g. the OVN leader checker) only distinguish IPv4 from
+// IPv6.
+func resolveProtocol(ctx context.Context, lookup func(ctx context.Context, host string) ([]net.IP, error), address string) (string, error) {
+	address = strings.Split(address, "/")[0]
+	if ip := net.ParseIP(address); ip != nil {
+		if ip.To4() != nil {
+			return kubeovnv1.ProtocolIPv4, nil
+		}
+		return kubeovnv1.ProtocolIPv6, nil
+	}
+
+	addrs, err := lookup(ctx, address)
+	if err != nil || len(addrs) == 0 {
+		return "", fmt.Errorf("failed to resolve address %q: %w", address, err)
+	}
+	var v4, v6 bool
+	for _, addr := range addrs {
+		if addr.To4() != nil {
+			v4 = true
+		} else if addr.To16() != nil {
+			v6 = true
+		}
+	}
+	switch {
+	case v4:
+		return kubeovnv1.ProtocolIPv4, nil
+	case v6:
+		return kubeovnv1.ProtocolIPv6, nil
+	default:
+		return "", fmt.Errorf("failed to resolve address %q to a valid IP", address)
+	}
+}
+
+// ResolveProtocol returns the address family of the given IP address or
+// hostname. Unlike CheckProtocol, which only accepts literal IP addresses,
+// hostnames are resolved via DNS.
+func ResolveProtocol(address string) (string, error) {
+	return resolveProtocol(context.Background(), func(ctx context.Context, host string) ([]net.IP, error) {
+		addrs, err := net.DefaultResolver.LookupIPAddr(ctx, host)
+		if err != nil {
+			return nil, err
+		}
+		ips := make([]net.IP, 0, len(addrs))
+		for _, addr := range addrs {
+			ips = append(ips, addr.IP)
+		}
+		return ips, nil
+	}, address)
 }
 
 func AddressCountBigInt(network *net.IPNet) internal.BigInt {
--- a/dist/images/start-db.sh
+++ b/dist/images/start-db.sh
@@ -255,6 +255,7 @@ function set_nb_version_compatibility() {
 
 if [[ -n "${NODE_IPS:-}" ]]; then
     normalize_raft_addrs
+    export POD_IP NODE_IPS DB_CLUSTER_ADDR
 fi
 
 function archive_recovery_file {
--- a/pkg/ovn_leader_checker/ovn.go
+++ b/pkg/ovn_leader_checker/ovn.go
@@ -566,10 +566,17 @@ func doOvnLeaderCheck(cfg *Configuration, podName, podNamespace string) {
 		util.LogFatalAndExit(nil, "preValidChkCfg: invalid cfg")
 	}
 
-	// Determine the expected AddressType based on pod IP protocol
+	// Determine the expected AddressType based on pod IP protocol.
+	// In HCP mode POD_IP is a headless-Service hostname (e.g.
+	// ovn-central-0.ovn-central.kube-system.svc) rather than a literal IP,
+	// so it must be resolved to determine the address family.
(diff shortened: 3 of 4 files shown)


pkg/util/net.go
```diff
@@ -213,8 +213,12 @@ func CheckProtocol(address string) string {
 	address = strings.Split(address, "/")[0]
 	ip := net.ParseIP(address)
 	if ip == nil {
-		klog.Errorf("failed to parse address %q", address)
-		return ""
+		addrs, err := net.LookupIP(address)
```
zhangzujian, line 216: [P1] Please keep the generic CheckProtocol validator literal-only, or move hostname resolution behind an explicit resolver used by the HCP leader-checker path. CheckProtocol is called throughout IP/CIDR validation, so this fallback makes arbitrary invalid strings trigger blocking ambient DNS lookups and changes them into valid protocols when they happen to resolve. In the PR worktree, go test ./pkg/util already fails because the existing invalid-input cases `not-an-ip` and `1.2.3` resolve to IPv4. Please add controlled regression tests for hostname success/failure without making all validation dependent on cluster DNS.

pkg/util/net.go
```diff
@@ -213,8 +213,12 @@ func CheckProtocol(address string) string {
...
+		if err != nil || len(addrs) == 0 {
+			klog.Errorf("failed to parse or resolve address %q: %v", address, err)
+			return ""
+		}
+		ip = addrs[0]
```
zhangzujian, line 221: [P1] `addrs[0]` is not a deterministic protocol choice for a hostname with both A and AAAA records. Resolver ordering can vary, so the same HCP pod may be classified as IPv4 or IPv6 across environments. Please define the dual-stack behavior explicitly (or use an address-family-aware lookup) and add a regression test covering a hostname that returns both families.

charts/kube-ovn-v2/templates/central/central-deployment.yaml
```diff
@@ -85,6 +85,8 @@ spec:
               {{- if .Values.central.hcp.enabled }}
               add:
                 - CHOWN
+                - DAC_OVERRIDE
```
zhangzujian, line 88: [P2] This capability change is unrelated to #7341's DNS/leader-checker fix and changes the HCP container privilege model. It also updates only kube-ovn-v2; the equivalent v1 HCP template still runs the same `chown -R` with only CHOWN while dropping all other capabilities. Please split this into a separate PR, or update both supported charts with focused rendering tests and explain why it belongs here.

zhangzujian (changes requested):
Request changes. Issue #7341 is valid: HCP ovn-central intentionally supplies a headless-Service hostname as POD_IP, while the leader checker needs the address family. The two DNS-related changes are directionally relevant, but the current implementation introduces a broad validator regression and lacks deterministic hostname coverage. The capability change is unrelated to this issue and should not be bundled without covering both chart variants.

pkg/util/net.go
cleman95: Done. CheckProtocol is back to literal-only, and hostname resolution now lives in an explicit resolver:

- `util.resolveProtocol(ctx, lookup, address)`  injectable lookup function so regression tests are deterministic and independent of cluster DNS
- `util.ResolveProtocol` exported wrapper over `net.DefaultResolver`

The leader checker is the only caller. Regression tests added in `TestResolveProtocol` cover hostname success/failure; `TestCheckProtocol` keeps its invalid-input cases (`not-an-ip`, `1.2.3`) as literal-only.

pkg/util/net.go
cleman95: Done. Dual-stack behavior is now defined explicitly: when a hostname resolves to both A and AAAA records, **IPv4 is preferred**. The choice must be deterministic because resolver ordering is not, and callers (EndpointSlice address type) only distinguish IPv4 from IPv6. Both orderings (v6-first and v4-first resolver output) are asserted in `TestResolveProtocol/hostnameResolvesBothFamilies*`.

charts/kube-ovn-v2/templates/central/central-deployment.yaml
cleman95: Agreed this change was unrelated to #7341 and only touched one chart. It is removed from this PR and will be submitted separately, covering both charts (kube-ovn and kube-ovn-v2)

cleman95:
The capabilities change is now in a standalone PR covering both charts: #7386.

Issue #7339, opened 2026-08-28
Split Helm chart into control-plane and data-plane + fix upgrade hook for dataPlaneOnly mode


cleman95:
## Context

We recently added `installMode` (`full`/`controlPlaneOnly`/`dataPlaneOnly`) and `externalOvnCentral` to the `kube-ovn-v2` chart to support split-cluster deployments where ovn-central runs on a management cluster and the data-plane agents connect back remotely.

While this works, we found that a monolithic chart with an `installMode` switch creates coupling issues. We propose two alternative approaches.

### 1. Chart coupling

- Hooks (`upgrade-ovs-ovn.yaml`, `post-delete-hook.yaml`) are rendered in both modes but only apply to the control-plane
- `nodeCount` helper does master node lookup even in `dataPlaneOnly` mode, causing `helm install` to fail if `masterNodesLabels` labels do not exist on any node
- CRDs are installed in both modes but only the data-plane needs them (control-plane has no controller)
- Impossible to version/deploy control-plane and data-plane independently

### 2. Upgrade hook broken in `dataPlaneOnly`

In `charts/kube-ovn-v2/templates/hooks/upgrade-ovs-ovn.yaml`:

- Lines 154-160: `OVN_NB_ADDR` is only set when `central.hcp.enabled` is true. The `dataPlaneOnly` mode is ignored, so the hook has no NB address to connect to.
- The underlying `upgrade-ovs.sh` script runs `kubectl rollout status deploy ovn-central --timeout=120s`, which fails with `NotFound` because `ovn-central` does not exist in `dataPlaneOnly` mode.

Current workaround: set `ovsOvn.upgrade.enabled: false`, which disables upgrade logic entirely.

### Option A: Keep `installMode`, fix the coupling issues

Keep the current monolithic chart but fix the conditionals:
- Guard hooks with proper installMode checks so they only render in the appropriate mode
- Fix `nodeCount` to not require master node lookup in `dataPlaneOnly`
- Fix upgrade hook to set `OVN_NB_ADDR` from `externalOvnCentral.nbEndpoint` in `dataPlaneOnly` mode
- Skip `kubectl rollout status deploy ovn-central` when ovn-central does not exist

Simpler change, backward compatible, but does not solve independent versioning.

### Option B: Split into umbrella chart with subcharts

Split `kube-ovn-v2` into an umbrella chart with two subcharts:

[...]

PR #7365, opened 2026-09-02
fix(chart): avoid master node lookup in dataPlaneOnly mode

zhangzujian:
## Summary

Addresses the legacy `kube-ovn` chart portion of #7339.

- Make `kubeovn.nodeCount` use Kubernetes node count in `dataPlaneOnly` mode.
- Avoid falling back to `kubeovn.nodeIPs`, which requires `MASTER_NODES_LABEL`; use an explicit `MASTER_NODES` fallback for offline rendering and default to one replica when no node data is available.
- Render the pre/post OVS upgrade hooks in `dataPlaneOnly` mode, inject the external OVN NB/SB addresses, and skip the local `ovn-central` rollout check that does not exist on tenant clusters.
- Add Helm unit coverage for CRD and hook boundaries in `controlPlaneOnly`/`dataPlaneOnly`, the remote upgrade-hook environment, and the no-master-label replica case.

The remaining `kube-ovn-v2` port will follow after #7366 lands so it can use the release-scoped resource names without template conflicts.

## Validation

- `helm unittest ./charts/kube-ovn` (41 tests)
- `helm lint ./charts/kube-ovn`
- `make verify-crd`
- `make lint`
- `bash -n dist/images/upgrade-ovs.sh`
- Script behavior checks for full and `dataPlaneOnly` modes

### DIFF
--- a/charts/kube-ovn/templates/_helpers.tpl
+++ b/charts/kube-ovn/templates/_helpers.tpl
@@ -57,10 +57,15 @@ nodeAffinity:
 {{- end -}}
 
 {{/*
-Number of master nodes
+Number of nodes used for component replicas. Data-plane-only installs count
+Kubernetes nodes directly because they do not require master node labels.
 */}}
 {{- define "kubeovn.nodeCount" -}}
+  {{- if eq .Values.installMode "dataPlaneOnly" -}}
+    {{- include "kubeovn.k8sNodeCount" . -}}
+  {{- else -}}
   {{- len (split "," (.Values.MASTER_NODES | default (include "kubeovn.nodeIPs" .))) }}
+  {{- end -}}
 {{- end -}}
 
 {{/*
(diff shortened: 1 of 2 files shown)


zhangzujian:
The `lint-test (ipv4)` failure is unrelated to this chart change. The job fails while creating the Kind cluster: `kindest/node:v1.37.0` rejects the repository's existing `kubeadm.k8s.io/v1beta3` config and requests v1.36. The Helm lint/install steps are not reached. Local chart validation passes (`helm unittest` 41/41, `helm lint`, `make verify-crd`, and `make lint`).

cleman95:
Thanks for this fix! However, this only ports `dataPlaneOnly` to the legacy `kube-ovn` chart. The `installMode` / `dataPlaneOnly` mode is still missing from `kube-ovn-v2`, which is the chart new deployments will use.

We've tested the split-cluster pattern (ovn-central on management cluster, data-plane on tenant cluster) with Kamaji and confirmed the `hcp` configuration + `dataPlaneOnly` works on kube-ovn-v2, but `installMode` needs to be ported there as well.

Since PR #7366 is refactoring the kube-ovn-v2 templates with dynamic release-scoped names, it would make sense to port `installMode` to kube-ovn-v2 **after** #7366 is merged, to avoid template conflicts. Would that be a reasonable sequencing?

PR #7375, opened 2026-09-03
feat: enhance tcpdump for kubevirt / fix: add default container to tcpdump / feat: enhance namespace selection for tcpdump

dgsponer:
# Pull Request

## What type of this PR
- Features
- Fix
- Docs

## Changes
- For troubleshooting and analysis, you can run kubectl ko tcpdump namespace/vm. This creates a packet capture session for the virtual machine, providing functionality equivalent to tcpdump. this cases are covered (pod / vm / virt-launcher)
- The default container has been added to kubectl exec. This prevents messages such as: ```Defaulted container "cni-server" out of: cni-server, hostpath-init (init), install-cni (init)``` from being displayed during execution.
- Namespace handling has been enhanced. If no namespace is provided as an input parameter, the tool first attempts to use the namespace of the current Kubernetes context. If the current context does not define a namespace, the tool automatically falls back to the default namespace.

### DIFF
--- a/dist/images/kubectl-ko
+++ b/dist/images/kubectl-ko
@@ -171,15 +171,15 @@ kubevirt-tcpdump(){
 
   if [ "$hostNetwork" = "true" ]; then
     set -x
-    kubectl exec "$ovsPod" -n $KUBE_OVN_NS -- tcpdump -nn "$@"
+    kubectl exec "$ovsPod" -n $KUBE_OVN_NS -c openvswitch -- tcpdump -nn "$@"
   else
-    nicName=$(kubectl exec "$ovsPod" -n $KUBE_OVN_NS -- ovs-vsctl --data=bare --no-heading --columns=name find interface external-ids:iface-id="$vmName"."$namespace" | tr -d '\r')
+    nicName=$(kubectl exec "$ovsPod" -n $KUBE_OVN_NS -c openvswitch -- ovs-vsctl --data=bare --no-heading --columns=name find interface external-ids:iface-id="$vmName"."$namespace" | tr -d '\r')
     if [ -z "$nicName" ]; then
       echo "nic doesn't exist on node $nodeName"
       exit 1
     fi
     podNicType=$(kubectl get pod "$podName" -n "$namespace" -o jsonpath={.metadata.annotations.ovn\\.kubernetes\\.io/pod_nic_type})
-    podNetNs=$(kubectl exec "$ovsPod" -n $KUBE_OVN_NS -- ovs-vsctl --data=bare --no-heading get interface "$nicName" external-ids:pod_netns | tr -d '\r' | sed -e 's/^"//' -e 's/"$//')
+    podNetNs=$(kubectl exec "$ovsPod" -n $KUBE_OVN_NS -c openvswitch -- ovs-vsctl --data=bare --no-heading get interface "$nicName" external-ids:pod_netns | tr -d '\r' | sed -e 's/^"//' -e 's/"$//')
     ovnCni=$(kubectl get pod -n $KUBE_OVN_NS -l app=kube-ovn-cni -o 'jsonpath={.items[?(@.spec.nodeName=="'$nodeName'")].metadata.name}')
     if [ -z "$ovnCni" ]; then
       echo "kube-ovn-cni not exist on node $nodeName"
@@ -187,9 +187,9 @@ kubevirt-tcpdump(){
     fi
     set -x
     if [ "$podNicType" = "internal-port" ]; then
-      kubectl exec "$ovnCni" -n $KUBE_OVN_NS -- nsenter --net="$podNetNs" tcpdump -nn -i "$nicName" "$@"
+      kubectl exec "$ovnCni" -n $KUBE_OVN_NS -c cni-server -- nsenter --net="$podNetNs" tcpdump -nn -i "$nicName" "$@"
     else
-      kubectl exec "$ovnCni" -n $KUBE_OVN_NS -- nsenter --net="$podNetNs" tcpdump -nn -i k6t-eth0 "$@"
+      kubectl exec "$ovnCni" -n $KUBE_OVN_NS -c cni-server -- nsenter --net="$podNetNs" tcpdump -nn -i k6t-eth0 "$@"
     fi
   fi
 }
@@ -218,15 +218,15 @@ tcpdump(){
 
   if [ "$hostNetwork" = "true" ]; then
     set -x
-    kubectl exec "$ovsPod" -n $KUBE_OVN_NS -- tcpdump -nn "$@"
+    kubectl exec "$ovsPod" -n $KUBE_OVN_NS -c openvswitch -- tcpdump -nn "$@"
   else
-    nicName=$(kubectl exec "$ovsPod" -n $KUBE_OVN_NS -- ovs-vsctl --data=bare --no-heading --columns=name find interface external-ids:iface-id="$podName"."$namespace" | tr -d '\r')
+    nicName=$(kubectl exec "$ovsPod" -n $KUBE_OVN_NS -c openvswitch -- ovs-vsctl --data=bare --no-heading --columns=name find interface external-ids:iface-id="$podName"."$namespace" | tr -d '\r')
     if [ -z "$nicName" ]; then
       echo "nic doesn't exist on node $nodeName"
       exit 1
     fi
     podNicType=$(kubectl get pod "$podName" -n "$namespace" -o jsonpath={.metadata.annotations.ovn\\.kubernetes\\.io/pod_nic_type})
-    podNetNs=$(kubectl exec "$ovsPod" -n $KUBE_OVN_NS -- ovs-vsctl --data=bare --no-heading get interface "$nicName" external-ids:pod_netns | tr -d '\r' | sed -e 's/^"//' -e 's/"$//')
+    podNetNs=$(kubectl exec "$ovsPod" -n $KUBE_OVN_NS -c openvswitch -- ovs-vsctl --data=bare --no-heading get interface "$nicName" external-ids:pod_netns | tr -d '\r' | sed -e 's/^"//' -e 's/"$//')
     ovnCni=$(kubectl get pod -n $KUBE_OVN_NS -l app=kube-ovn-cni -o 'jsonpath={.items[?(@.spec.nodeName=="'$nodeName'")].metadata.name}')
     if [ -z "$ovnCni" ]; then
       echo "kube-ovn-cni not exist on node $nodeName"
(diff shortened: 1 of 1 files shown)


dist/images/kubectl-ko
```diff
@@ -146,6 +147,53 @@ ipIsInCidr(){
...
+    ovnCni=$(kubectl get pod -n $KUBE_OVN_NS -l app=kube-ovn-cni -o 'jsonpath={.items[?(@.spec.nodeName=="'$nodeName'")].metadata.name}')
+    if [ -z "$ovnCni" ]; then
+      echo "kube-ovn-cni not exist on node $nodeName"
+      exit 1
+    fi
+    set -x
+    if [ "$podNicType" = "internal-port" ]; then
+      kubectl exec "$ovnCni" -n $KUBE_OVN_NS -c cni-server -- nsenter --net="$podNetNs" tcpdump -nn -i "$nicName" "$@"
+    else
+      kubectl exec "$ovnCni" -n $KUBE_OVN_NS -c cni-server -- nsenter --net="$podNetNs" tcpdump -nn -i k6t-eth0 "$@"
```
zhangzujian: Why should we specify interface `k6t-eth0` instead of `eth0`? Can we use the veth peer `xxx..._h`?

dgsponer: this is the interface list of 2 vm's. eth0 is down, he takes the k6t-eth0 instead

# ip link
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: k6t-eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 8900 qdisc noqueue state UP mode DEFAULT group default qlen 1000
    link/ether 7e:e6:90:a0:75:c0 brd ff:ff:ff:ff:ff:ff
3: tap0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 8900 qdisc fq master k6t-eth0 state UP mode DEFAULT group default qlen 1000
    link/ether ce:62:20:ea:a3:98 brd ff:ff:ff:ff:ff:ff
4: eth0: <BROADCAST,NOARP> mtu 8900 qdisc noop state DOWN mode DEFAULT group default qlen 1000
    link/ether 22:af:f9:2d:d1:aa brd ff:ff:ff:ff:ff:ff
192777: eth0-nic@if192778: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 8900 qdisc noqueue master k6t-eth0 state UP mode DEFAULT group default
    link/ether 7e:e6:90:a0:75:c0 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    alias d909221a4a4b_c


# ip link
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: k6t-eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 8900 qdisc noqueue state UP mode DEFAULT group default qlen 1000
    link/ether 3e:e2:87:a0:8f:ea brd ff:ff:ff:ff:ff:ff
3: tap0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 8900 qdisc mq master k6t-eth0 state UP mode DEFAULT group default qlen 1000
    link/ether 7a:db:a5:09:80:39 brd ff:ff:ff:ff:ff:ff
4: eth0: <BROADCAST,NOARP> mtu 8900 qdisc noop state DOWN mode DEFAULT group default qlen 1000
    link/ether 42:fb:32:3d:0e:05 brd ff:ff:ff:ff:ff:ff
99: eth0-nic@if100: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 8900 qdisc noqueue master k6t-eth0 state UP mode DEFAULT group default
    link/ether 3e:e2:87:a0:8f:ea brd ff:ff:ff:ff:ff:ff link-netnsid 0
    alias 13cfb6262756_c

dgsponer: i will test the output of the eth0-nic and try to compare the content

dgsponer: it looks like, there is no diff between the k6t-eth0 and eth0-nic

zhangzujian: There are two methods we can use so that ko tcpdump works on both normal pods and kubevrit VMs:

1. If `eth0` is a dummy interface, find the actual veth interface `eth0-nic` by interface type `veth`, interface name (`eth0-nic`) or the master name (`k6t-eth0`), and alias (`13cfb6262756_c`). Then we can capture packets on `eth0-nic`.
2. Find the ovs interface name (`xxx..._h`) and capture packets on it.

dgsponer: i can add a test case, to check is it a vm or a container by "ownerRef" and this change the behaviour, how the nicname is readed out?

dgsponer: for your context: the pod, that contains the vm, starts with virt-launcher, but the network is registered with the vm name:

VMNAME and NAMESPACE is a placeholder for my data.

```iface-id=VMNAME.NAMESPACE ip=10.16.17.216 ovn-installed=true ovn-installed-ts=1788337962632 pod_name=VMNAME pod_namespace=NAMESPACE pod_netns=/var/run/netns/cni-e14aa580-a9d0-a01b-4b44-86f6b2843d47 vendor=kube-ovn```

Comment (zhangzujian) [reply dist/images/kubectl-ko]: > Find the ovs interface name (xxx..._h) and capture packets on it.

We can simply capture packets on this interface if `nic-type` is `veth-pair`.

Comment (dgsponer) [reply dist/images/kubectl-ko]: you mean then without entering the namespace like this?
```kubectl exec "$ovnCni" -n kube-system -c cni-server -- tcpdump -nn -i 9ac2655e_net1_h "$@"```

Comment (zhangzujian) [reply dist/images/kubectl-ko]: That's exactly what I mean.

Comment (dgsponer) [reply dist/images/kubectl-ko]: yes i verfied also that, the interface in the example was wrong and i was confused about the result.

do we need then this testcase "podNicTyp" because we can dump it form the ovs-ovn without entering the namespace.

if not, i will
- cleanup the kubevirt-tcpdump
- remove the testcase for podNicTyp
- add the test: is it a vm, read out the interface and go with the default routine.

Comment (dgsponer) [reply dist/images/kubectl-ko]: changes done

File: dist/images/kubectl-ko [hunk 2]
```diff
@@ -149,14 +149,26 @@ ipIsInCidr(){
...
   podName=$(echo "$namespacedPod" | cut -d "/" -f2)
   if [ "$podName" = "$namespacedPod" ]; then
-    namespace="default"
+    namespace=$(kubectl config view --minify -o jsonpath='{..namespace}')
+    namespace=${namespace:-default}
   fi
 
-  nodeName=$(kubectl get pod "$podName" -n "$namespace" -o jsonpath={.spec.nodeName})
-  hostNetwork=$(kubectl get pod "$podName" -n "$namespace" -o jsonpath={.spec.hostNetwork})
+  isVm=$([[ "$(kubectl get vmi "$podName" -o jsonpath='{.metadata.ownerReferences[?(@.kind == "VirtualMachine")].kind}' 2>/dev/null)" == "VirtualMachine" ]] && echo true || echo false)
```
zhangzujian, line 159: The KubeVirt probes here and at lines 162, 165, and 167 omit `-n "$namespace"`, so they always use the current context namespace. With `kubectl ko tcpdump ns2/vm` while the context is `ns1`, the VMI/launcher lookup runs in `ns1`, but line 171 later fetches the Pod in `ns2`; VM detection therefore fails or resolves the wrong object and the subsequent OVS interface lookup breaks. Pass the parsed namespace to every VMI/Pod lookup.

dist/images/kubectl-ko
```diff
@@ -149,14 +149,26 @@ ipIsInCidr(){
...
+  if [ "$isVm" = "true" ]; then
+    vmName=$(echo "$podName")
+    podName=$(kubectl get pod -l vm.kubevirt.io/name="$vmName" -o jsonpath='{.items[*].metadata.name}')
```
zhangzujian, line 162: `{.items[*].metadata.name}` returns all matching launcher names as one space-separated string. During KubeVirt live migration, two virt-launcher Pods can coexist, so line 171 receives e.g. `"virt-launcher-a virt-launcher-b"` as a single Pod name and kubectl returns NotFound. Select a single eligible Running launcher (with the namespace), or process the candidates individually.

zhangzujian (changes requested):
Request changes. The tcpdump consolidation has correctness issues in namespace resolution and VM pod selection. Please also add regression coverage for explicit namespaces and KubeVirt live-migration states. The new ownerReference probes currently suppress every kubectl error with 2>/dev/null, which hides RBAC/API failures; handle NotFound separately from unexpected errors. Finally, remove the trailing whitespace at line 170 and restore the final newline.

dist/images/kubectl-ko
dgsponer: thank you very much, I already work on it

dgsponer: hi zhang, I add now a "handler" for the migration, there are 3 cases: never migrated, in progress, done. during migration: sourcePod, done: targetPod, never done: grab the container from node. I saw also, when the migration is done, the completed containers could be on the same node, that caused then an issue, this is why I try to consume all infos from the vmi.

dist/images/kubectl-ko
```diff
@@ -168,28 +194,29 @@ tcpdump(){
...
       echo "kube-ovn-cni not exist on node $nodeName"
       exit 1
     fi
     set -x
-    if [ "$podNicType" = "internal-port" ]; then
-      kubectl exec "$ovnCni" -n $KUBE_OVN_NS -- nsenter --net="$podNetNs" tcpdump -nn -i "$nicName" "$@"
-    else
-      kubectl exec "$ovnCni" -n $KUBE_OVN_NS -- nsenter --net="$podNetNs" tcpdump -nn -i eth0 "$@"
-    fi
+    kubectl exec "$ovnCni" -n $KUBE_OVN_NS -c cni-server -- tcpdump -nn -i "$nicName" "$@"
```
zhangzujian, line 219: Blocking: this unconditional host-network capture regresses the legacy `internal-port` path. `nicName` is the OVS internal-port interface, which is placed in the Pod network namespace; `cni-server` is `hostNetwork: true`, so running `tcpdump -i "$nicName"` here cannot see that interface. The previous implementation read `external-ids:pod_netns` and ran `nsenter --net="$podNetNs"`, with a separate `internal-port` branch, and `trace()` in this same file still preserves that distinction. Please retain the namespace-aware branch, or explicitly remove/validate `internal-port` support together with the stale `POD_NIC_TYPE` configuration and add regression coverage.

zhangzujian (changes requested):
Request changes. Re-review of the current head 18d82d07393256cafaf37866a9611ef680fdf802:

- Lines 159 and 179 still redirect ownerReference probe errors to /dev/null. RBAC/API failures become indistinguishable from a non-VM/non-VMI; handle NotFound separately and surface unexpected errors.
- Lines 165-175 print migration warnings and the selected Pod name to stdout. This contaminates tcpdump output when stdout is redirected or consumed as a capture stream; diagnostics should go to stderr.
- The fallback at line 174 returns every matching launcher on the node. The newline-separated result is later passed as one Pod argument, so stale or concurrent launchers can still make the lookup fail; select one eligible Running launcher explicitly.
- No regression coverage was added for explicit namespaces, live-migration states, ordinary Pod/VM selection, or the internal-port path.
- The changed file still has trailing whitespace at line 184 and lacks a final newline.

The current-head internal-port regression is also recorded in the inline blocking comment at line 219.

PR #7378, opened 2026-09-04
feat: add service-scoped OVN load balancers

### LINKED ISSUE #7364: [Feature Request] Support Service Local traffic policy and per-Service session affinity timeout with OVN distributed LB
## Description

Kube-OVN currently uses VPC-level shared load balancers for Kubernetes Services. This prevents the controller from expressing two Kubernetes Service semantics:

1. `spec.sessionAffinityConfig.clientIP.timeoutSeconds` is ignored. All ClientIP Services in the same VPC share one session load balancer and therefore one `affinity_timeout` (currently the 10800-second default).
2. `spec.internalTrafficPolicy: Local` is not enforced when OVN LB handles ClusterIP traffic. The existing `prefer_local_backend` path is optional and means local preference, not strict local-only routing.

This was reproduced with Kube-OVN v1.15.23 and `--enable-lb=true` on Kubernetes v1.35.6: a Service configured with `timeoutSeconds: 10` retained the same backend after more than 10 seconds of idle time, and an `internalTrafficPolicy: Local` Service could reach a remote backend or a backend from a node without local endpoints.

OVN 26.03 introduces the native `Load_Balancer.options:distributed` option. It generates `ct_lb_mark_local` and routes a new connection only to backend instances whose logical ports are local to the chassis receiving the traffic. OVN requires `ip_port_mappings` for this mode and drops traffic when no usable local mapping exists, which provides the data-plane primitive needed for strict `internalTrafficPolicy: Local`.

### Proposed implementation

- Keep VPC-shared LBs for ordinary `Cluster` Services.
- Create a Service-scoped LB per Service/protocol for `ClientIP` and/or `internalTrafficPolicy: Local`, keyed by Service UID and recorded with `external_ids` for lifecycle and garbage collection.
- Set `options:affinity_timeout` from `clientIP.timeoutSeconds` (default 10800 seconds).
- Set `options:distributed=true` for `internalTrafficPolicy: Local` and maintain `ip_port_mappings` in the form `backendIP=backendLogicalPort` (or `backendLogicalPort:sourceIP` when health checks are enabled).
- Derive backend logical ports and chassis from EndpointSlice/Pod targets, update mappings on endpoint readiness or node migration, and remove stale mappings transactionally.
- Ensure a Service VIP has one authoritative LB owner during policy changes and upgrades; migrate existing VIPs from shared LBs without a window that routes to an unintended backend.
- Extend subnet LB attachment and garbage collection to discover Service-scoped LBs without deleting them as unknown resources.
- Add an explicit feature gate and document the OVN/OVS version prerequisite. Do not change the existing `prefer-local` flag semantics implicitly.

### Compatibility and limits

- OVN 26.03 is a prerequisite for `distributed=true`; the actual Kube-OVN image must contain the matching OVN northd/controller implementation.
- OVN `affinity_timeout` has a maximum of 65535 seconds, while Kubernetes permits up to 86400 seconds. Values above 65535 must be rejected or surfaced as unsupported; they must not be silently truncated.
- Validate behavior for logical-switch and logical-router attachment, health checks, endpoint migration, empty local backend, DGP/multiple-DGP, IPv4, IPv6, dual-stack, TCP, UDP and SCTP.

### Acceptance criteria

- A `ClientIP` Service with `timeoutSeconds: 10` expires affinity after the configured idle timeout, and the generated OVN LB reports `affinity_timeout=10`.
- Two-node `internalTrafficPolicy: Local` tests select only the local Ready backend; when the local endpoint is removed, new connections fail instead of falling back to a remote backend; traffic recovers after the local endpoint returns.
- A Service configured with both ClientIP affinity and Local traffic policy satisfies both guarantees.
- Multiple Services in one VPC can use different affinity timeouts concurrently.
- Existing `Cluster` Services and non-LB paths remain compatible.
- Controller restart, Service deletion, endpoint readiness changes, node migration, upgrade and rollback do not leak VIPs, mappings or Service-scoped LBs.

### References

- Previous broad LB refactor request (closed/stale): https://github.com/kubeovn/kube-ovn/issues/5013
- OVN 26.03 release notes: https://github.com/ovn-org/ovn/blob/v26.03.0/NEWS
- OVN 26.03 Load_Balancer schema (`distributed` option): https://github.com/ovn-org/ovn/blob/v26.03.0/ovn-nb.xml#L2750-L2762
- OVN northd `ct_lb_mark_local` implementation: https://github.com/ovn-org/ovn/blob/v26.03.0/northd/northd.c#L3388-L3450
- OVN distributed LB tests: https://github.com/ovn-org/ovn/blob/v26.03.0/tests/ovn-northd.at#L18404-L18436

## Who will benefit from this feature?

Kubernetes users that rely on Service session stickiness, topology-local traffic, node-local caches, local proxies, or workloads where cross-node east-west traffic is undesirable.

## Anything else?

SCTP dataplane behavior should be tracked separately because it involves independent OVN SCTP LB, kube-proxy IPVS NodePort, and hairpin paths and should not be conflated with this feature request.

zhangzujian:
## Summary

- Create per-Service OVN load balancers for Kubernetes Services, including ordinary ClusterIP, ClientIP session affinity, `internalTrafficPolicy: Local`, and LoadBalancer ingress.
- Honor `clientIP.timeoutSeconds` with OVN range validation (`[1, 65535]`). Do not set `selection_fields`, and do not enable OVN health checks on ClusterIP/LoadBalancer VIPs.
- Implement ClusterIP `trafficDistribution` with OVN template load balancers and `Chassis_Template_Var`, consuming EndpointSlice `forNodes` / `forZones` hints (`PreferSameNode`: node → zone → cluster; `PreferSameZone`/`PreferClose`: zone → cluster).
- Split template load balancers by protocol and address family, migrate VIPs off the legacy shared VPC LBs, and garbage-collect empty legacy LBs.
- Use `options:distributed=true` plus backend `ip_port_mappings` for strict `internalTrafficPolicy: Local`. LoadBalancer `externalTrafficPolicy: Local` uses speaker-chassis template variables so underlay MetalLB steal flows keep locality without `prefer_local_backend`.
- Attach Service-scoped LBs to the owning VPC router and enabled OVN switches, skip missing logical switches, and annotate LoadBalancer subnet ownership so kube-ovn-cni can install underlay OpenFlow.
- Cover controller/OVS unit tests and kube-ovn/MetalLB e2e. SwitchLBRule and RouterLBRule generated Services reuse the same scoped-LB owner model so they keep working after the shared VPC LBs are retired.

Closes #7364

## Validation

- `go test ./pkg/controller ./pkg/daemon ./pkg/ovs -count=1`
- `go vet ./pkg/controller ./pkg/daemon ./pkg/ovs`
- `golangci-lint run ./pkg/controller/...`
- `git diff --check`


zhangzujian:
Follow-up fix in `7f30662a4` (rebased onto current `master`):

- Rule-owned generated Services now clean only their resource-scoped LB and do not touch retired fixed VPC LBs during Service delete races.
- RouterLBRule deletion recovers VIPs from the referenced OvnEIP when the generated Service is already absent, preventing interrupted legacy VIP migration from leaking.
- Added regressions; `go test ./pkg/controller ./pkg/ovs`, `go vet ./pkg/controller`, `make lint`, and `git diff --check` pass.

zhangzujian:
Fixed the IPv4 installation race in commit e328c7480. EndpointSlice reconciliation now polls the API for the requested health-check VIP address family when a concurrent worker observes an unallocated VIP, preventing IPv4 service setup from stalling at `vip ovn-default is not ready`. Added stale-informer and delayed-status regression tests.

Validation: `go test ./pkg/controller`, `go test ./pkg/ovs`, `go vet ./pkg/controller`, and `make lint` (0 issues).

/test e2e core

zhangzujian:
The underlay IPv4 conformance failure in [run 34709224263 / job 103597197497](https://github.com/kubeovn/kube-ovn/actions/runs/34709224263/job/103597197497) belongs to this PR and is fixed by `5b2df9488`.

The service-scoped internal LB was attached to both logical switches and the VPC router. That made host-network ClusterIP traffic enter OVN through the join path, so the backend observed the join IP (`100.64.0.2`) instead of the node IP. ClusterIP and NodePort scoped LBs now retain the legacy switch-only attachment behavior, while LoadBalancer Services and RouterLBRules keep router attachment.

Validation:

- `go vet ./pkg/controller`
- `git diff --check`

zhangzujian:
Follow-up fix in commit 16409b02b: ordinary ClusterIP service-scoped LB migration now sends target VIP/backend update, Logical_Switch LB attachment mutations, and legacy VIP removal in one OVN NB transaction. Added an OVN client regression test covering the atomic migration path.

Local validation: `go test ./pkg/controller ./pkg/ovs -count=1`, `go vet ./pkg/controller`, and `git diff --check` pass. The trusted run is currently building the DPDK base image; its control-module validation still fails on the default branch's stale cache assertion before Conformance executors start.,

zhangzujian:
Rechecked after rebasing onto master (HEAD `eb3ad22d03e060e0281ab937737f641287b0514a`) and completed trusted run [34881359892](https://github.com/kubeovn/kube-ovn/actions/runs/34881359892).

Kubernetes Conformance results: IPv4 underlay, IPv6 underlay, dual underlay, and dual overlay passed; IPv4 overlay and IPv6 overlay failed only in the upstream ClientIP session-affinity HTTP/UDP specs. The failures are not introduced by this PR: baseline run [34809407681](https://github.com/kubeovn/kube-ovn/actions/runs/34809407681) on the pre-PR code reproduced the same assertion (`Unexpected endpoints return: map[netserver-... netserver-...], expect 1 endpoints`).

The new run also contains independent infrastructure/control-plane symptoms, including `ovn-northd is not running`, missing OVS DB socket, readiness probe failures, pod sandbox failures, and provider-network bridge attachment errors in the IPv6 overlay events. Therefore there is no evidence that the service-scoped LB transaction changes or the scoped ECMP patch are the cause, and I am not changing the LB implementation based on these baseline-equivalent failures.

Local gates on the rebased HEAD remain green: `make build-go`, `make lint`, targeted Go tests, Python E2E control tests, and `git diff --check`.

PR #7405, opened 2026-09-09
fix: close orphaned NAT leaks in ovn fip/snat/dnat/eip controllers

andriishestakov:
# Pull Request

- Bug fixes
- Tests

Fixes orphaned NAT/load-balancer leaks in the OVN FIP/SNAT/DNAT/EIP controllers:

- `ovn_fip.go`, `ovn_snat.go`: `Status` (which the delete path uses to decide whether OVN cleanup is needed) was only patched as the last step of the add flow, after the NAT rule already existed in OVN. Deleting the resource before that patch landed left the NAT rule orphaned
forever. `Status` is now recorded right after the NAT rule is created. Delete paths also became idempotent (`deleteNatIgnoreNotFound`) so a NAT rule missing in OVN no longer blocks finalizer removal.
- `ovn_fip.go`: `patchOvnFipStatus` never persisted `V6Eip`/`V6Ip`, so the v6 NAT rule of any IPv6/dual-stack FIP was never deleted. Fixed.
- `ovn_dnat.go`: `ExternalPort` was only recorded once `ready=true`; the delete path needs it earlier, mirroring the fip/snat fix.
- `ovn_eip.go`: `handleDelOvnEip` released the EIP's IP in IPAM unconditionally, without checking whether a FIP/DNAT/SNAT was still using it — unlike the primary `handleUpdateOvnEip` delete path, which already guards on `getOvnEipNat`. This allowed a still-referenced IP to be
released and reissued elsewhere. Added the same guard.

Added unit tests for all four controllers covering the fixed behavior.

### DIFF
--- a/pkg/controller/ovn_dnat.go
+++ b/pkg/controller/ovn_dnat.go
@@ -205,6 +205,10 @@ func (c *Controller) handleAddOvnDnatRule(key string) error {
 			return err
 		}
 	}
+	if err = c.patchOvnDnatStatus(key, vpcName, v4Eip, v6Eip, internalV4Ip, internalV6Ip, false); err != nil {
+		klog.Errorf("failed to record nat status for dnat %s, %v", key, err)
+		return err
+	}
 	if err := c.handleAddOvnDnatFinalizer(cachedDnat); err != nil {
 		klog.Errorf("failed to add finalizer for ovn dnat %s, %v", cachedDnat.Name, err)
 		return err
@@ -556,7 +560,7 @@ func (c *Controller) patchOvnDnatStatus(key, vpcName, v4Eip, v6Eip, internalV4Ip
 		dnat.Status.InternalPort = dnat.Spec.InternalPort
 		changed = true
 	}
-	if ready && dnat.Spec.ExternalPort != "" && dnat.Status.ExternalPort != dnat.Spec.ExternalPort {
+	if dnat.Spec.ExternalPort != "" && dnat.Status.ExternalPort != dnat.Spec.ExternalPort {
 		dnat.Status.ExternalPort = dnat.Spec.ExternalPort
 		changed = true
 	}
--- a/pkg/controller/ovn_dnat_test.go
+++ b/pkg/controller/ovn_dnat_test.go
@@ -1,0 +1,110 @@
+package controller
+
+import (
+	"errors"
+	"strings"
+	"testing"
+
+	"github.com/ovn-kubernetes/libovsdb/ovsdb"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	k8stesting "k8s.io/client-go/testing"
+
+	kubeovnv1 "github.com/kubeovn/kube-ovn/pkg/apis/kubeovn/v1"
+	kubeovnfake "github.com/kubeovn/kube-ovn/pkg/client/clientset/versioned/fake"
+	"github.com/kubeovn/kube-ovn/pkg/util"
+)
+
+func TestHandleAddOvnDnatRule_RecordsNatStatusBeforeReady(t *testing.T) {
+	t.Parallel()
+
+	eip := &kubeovnv1.OvnEip{
+		Name: "test-eip",
+		Spec: kubeovnv1.OvnEipSpec{ExternalSubnet: "pubnet"},
+	}
+	eip.Status.V4Ip = "10.0.0.10"
+	dnat := &kubeovnv1.OvnDnatRule{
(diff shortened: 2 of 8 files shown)


andriishestakov:
@zbb88888 Hi. could you review this fix pls?

zbb88888 (approved):

zbb88888:
@andriishestakov Hi, Thanks for the PR!

There are some issues left, could you please continue to check and  fix?


 Title: fix(nat): make FIP/DNAT/SNAT cleanup robust, idempotent, and IPv6/cross-family complete

  Problem statement

  1. Cleanup for FIP/DNAT/SNAT deleted before their finalizer is persisted does not work. The delete queue carries only the key (c.delOvnFipQueue.Add(key) in
     enqueueDelOvnFip, same for DNAT/SNAT), and the handler re-reads the object from the informer lister (handleDelOvnFip, handleDelOvnDnatRule, handleDelOvnSnatRule).
     client-go removes the object from the indexer before invoking DeleteFunc, so lister.Get(key) returns NotFound and the handler returns nil without calling DeleteNat/
     DelDnatRule. Verified with a real shared informer + fake clientset: inside DeleteFunc, lister.Get("fip1") returns ovnfip.kubeovn.io "fip1" not found, and
     handleDelOvnFip then performs no OVN cleanup.

  2. FIP rules never get the ovn.kubernetes.io/eip_v6_ip label (patchOvnFipStatus only writes EipV4IpLabel, unlike patchOvnDnatStatus/patchOvnSnatStatus), so
     getOvnEipNat("", eip.Spec.V6Ip) cannot see IPv6-only FIPs and handleDelOvnEip releases the IPAM address while the FIP still uses it.

  3. Add and delete use different pairing rules for FIP NATs. handleAddOvnFip also creates cross-family rules (AddNat(v6Eip, v4IP) for v4-internal/v6-external, and
     AddNat(v4Eip, v6IP) for the reverse), but the delete paths only try (Status.V4Eip, Status.V4Ip) and (Status.V6Eip, Status.V6Ip). Cross-family FIPs still leak their
     NAT row.

  4. deleteNatIgnoreNotFound is redundant: OVNNbClient.DeleteNat already returns success when the NAT row is absent (added in #6911). The wrapper only adds an extra
     NatExists round trip per delete.

  5. The DNAT delete path is still not idempotent: DelDnatRule → LoadBalancerDeleteVip → GetLoadBalancer(lbName, false) fails when the LB is already gone, so the DNAT
     stays terminating. handleDelOvnDnatRule also returns an error when Status.Vpc == "", which can only spin the queue.

  Acceptance criteria

  - Deleting a FIP/DNAT/SNAT immediately after its OVN object is created leaves no orphan NAT/LB and no stuck finalizer; the cleanup converges on retry.
  - Deleting twice, or after a partial failure, is a no-op instead of an error.
  - An IPv6-only EIP used by a FIP is reported as in-use by handleDelOvnEip; the IPAM address is not released.
  - A cross-family FIP deletes exactly the NAT row that was created at add time.
  - New unit tests fail on the current code and pass after the fix.

  Suggested implementation

  - Step 1 — stop depending on a re-read of a deleted object. Preferred: add the finalizer before creating any OVN state, and re-check DeletionTimestamp on a freshly
    fetched object right before each data-plane write (return early if deletion started, so the update-queue delete branch just removes the finalizer). Alternative: make
    delOvnFipQueue/delOvnDnatRuleQueue/delOvnSnatRuleQueue carry the deleted object instead of the key, like delOvnEipQueue already does with *kubeovnv1.OvnEip, and have
    the handlers use that object rather than the lister.

  - Step 2 — persist what was actually created. In patchOvnFipStatus, set both EipV4IpLabel and EipV6IpLabel (util.IPv6ToLabelValue(v6Eip)). Also fix isOvnFipDuplicated,
    which matches {eip_v4_ip: ""} for v6-only EIPs and therefore treats unrelated IPv6 FIPs as duplicates. For cross-family rules, either record the effective (externalIP,
    logicalIP) pairs in status at creation time (least error-prone) or mirror the add-time pairing in both delete paths.

  - Step 3 — remove deleteNatIgnoreNotFound (or keep it only if the client contract changes); DeleteNat is already idempotent.
  - Step 4 — make DNAT deletion idempotent. Treat a missing LB as success, avoid calling LogicalRouterUpdateLoadBalancers once per family (the LB is per-DNAT rule, not per
    IP family), and define the expected behavior for Status.Vpc == "" instead of returning an error from an object that no longer exists.

  - Step 5 — tests. Add scenario tests that drive a real informer + fake clientset: (a) delete right after NAT/LB creation → assert cleanup happened; (b) v6 FIP → assert
    getOvnEipNat("", eip.Status.V6Ip) returns util.FipUsingEip and handleDelOvnEip returns an error; (c) cross-family FIP delete; (d) DNAT delete when the LB is already
    gone; (e) all delete handlers are idempotent when invoked twice.

PR #7407, opened 2026-09-10
fix(controller): switch_lb_vip uses own mac, not gateway mac

andriishestakov:
switch_lb_vip forced its LSP to reuse the subnet gateway's mac for arp_proxy. Two logical ports with the same mac collide in OVN's L2 lookup, non-deterministically breaking subnet egress. Actual LB traffic is delivered via ct_lb_mark (DNAT by ip:port), which runs before L2 lookup and never depended on the shared mac.


andriishestakov:
@zbb88888 could you review this fix please?

pkg/controller/vip.go
```diff
@@ -118,20 +118,7 @@ func (c *Controller) handleAddVirtualIP(key string) error {
...
-		}
-		if lrp.MAC == "" {
-			err = fmt.Errorf("logical router port %s should have mac", lrpName)
-			klog.Error(err)
-			return err
-		}
-		mac = lrp.MAC
+		// create a lsp using the vip's own mac, and set it option as arp_proxy
 		ipStr := util.GetStringIP(v4ip, v6ip)
 		if err := c.OVNNbClient.CreateLogicalSwitchPort(subnet.Name, portName, ipStr, mac, vip.Name, vip.Spec.Namespace, false, "", "", false, nil, subnet.Spec.Vpc); err != nil {
```
zhangzujian, line 144: Please update the existing `[group:vip]` E2E in this change. Passing the IPAM-assigned MAC here makes the two Switch LB VIP MACs distinct, so `test/e2e/vip/e2e_test.go:568` still checks the old shared gateway-MAC behavior and fails. Change that assertion to expect distinct MACs and update the adjacent comments. See https://github.com/kubeovn/kube-ovn/actions/runs/34590982048/job/103241390370.

zhangzujian (changes requested):
Request changes. The production change correctly stops reusing the subnet gateway MAC, but the existing VIP E2E contract was not updated. `test/e2e/vip/e2e_test.go:568` still asserts that two `SwitchLBRuleVip` objects have equal MACs, and the comments at lines 560/567 describe the old shared-MAC behavior. Trusted x86 job 103241390370 reproduced this exact failure with two different MACs. Please update the assertion to expect distinct MACs, revise the stale comments, and rerun `vip-conformance-e2e`.

zbb88888:
> @zbb88888 could you review this fix please?

I will try let switch lb type vip use its own mac address

zbb88888:
@andriishestakov  Hi, thanks for the PR!


The "arp proxy" part of switch_lb_vip never actually worked after release 1.12. 

  ## 1. The e2e contract (the only blocker)

  test/e2e/vip/e2e_test.go:568 must flip to ExpectNotEqual, and the ginkgo.By/comment at lines 560 and 567 ("should have the same mac which is from gw subnet mac", "the lb
  vip use the subnet gw mac to use lb nat flow") have to be rewritten — that comment is the last place in the repo encoding the old contract. While you're there, make the
  new invariant explicit: assert the VIP MAC also differs from the subnet gateway MAC (that's literally what the fix is about), not just that the two VIPs differ from each
  other. Then re-run the vip conformance job; it's the only PR-attributable failure.

  ## 2. Remove the dead arp_proxy machinery in this PR

  With the own-MAC decision, SetLogicalSwitchPortArpProxy(portName, true) is dead code, and it was never functional anyway:

  - OVN's Logical_Switch_Port.options:arp_proxy is an address/CIDR list for a router-type port. kube-ovn writes "true" on a plain vif LSP, so northd logs Invalid arp_proxy
    option: 'true' at lsp ... and installs nothing (I checked OVN 22.03 through master; the parsing/type requirement is the same).

  - The VIP LSP is never chassis-bound, so NB up=false, and northd only emits the ARP/ND responder for known IPs when the port is up (or is router/localport —
    check_lsp_is_up defaults to true and kube-ovn doesn't set ignore_lsp_down). So the LSP can't answer ARP for the VIP either.

  So the LSP cannot answer ARP and cannot be a delivery target (packets addressed to its MAC go to an unbound port and are dropped). Drop the call in pkg/controller/
  vip.go, and since nothing else uses it, drop the interface method, generated mock and the pkg/ovs unit test too (AGENTS.md explicitly asks to clean up related dead
  logic). That also forces the semantics to be stated in the code/PR: switch_lb_vip is a routed VIP — clients reach it through the gateway, the LB DNATs by IP/port, and
  ls_in_l2_lkup forwards the DNAT'ed packet to the gateway MAC.

  If you ever want in-subnet ARP for these VIPs, the supported way is options:arp_proxy="<vip>" on the subnet's switch-side router LSP (<vpc>-<subnet>, type=router,
  options:router-port=...) — OVN then answers with the peer LRP's MAC, which keeps the routed delivery path and introduces no duplicate MAC. That's a follow-up, not this
  PR.

  ## 3. Existing objects (this is not back-compat, it's the bug surviving)

  handleAddVirtualIP early-returns on Status.Mac != "" and the LSP MAC is only written at creation, so VIPs created before the upgrade keep the shared gateway MAC and the
  L2 collision keeps happening. Since you're not preserving the old behavior, either add a repair path (rewrite LSP addresses, vip.Status.Mac, and the IPAM entry) or
  explicitly document the delete/recreate workaround in the PR description/release notes. Without one of the two, the fix only helps newly created VIPs.

  ## 4. Tighten the unit test to the new contract

  vip_test.go only asserts gomock.Not("") for the MAC. Assert the MAC used for the LSP is the IPAM-assigned one and differs from the subnet gateway MAC, so a regression to
  the shared-MAC behavior fails the unit test too, not just the e2e.

  ## Things I checked so you don't have to worry about them

  - IPAM already guards the invariant: Subnet.GetStaticMac is called with checkConflict=true from both acquireStaticIPAddress and acquireIPAddress, and it rejects a MAC
    already allocated or equal to the subnet's GatewayMAC. So a user can't reintroduce the collision via spec.macAddress, and the random path yields a per-VIP unique MAC.

  - Delete path (handleUpdateVirtualIP) releases by port name/IPAM and is MAC-agnostic — nothing to change.
  - init.go re-registers the static address using vip.Status.Mac, so controller restarts stay consistent after the change.
  - Open design question worth one sentence in the PR: with own MAC the VIP LSP has no datapath role at all (no ARP, no delivery); it now only holds the LS-side address
    and anchors GC bookkeeping. If you agree, say so, and consider whether the LSP should exist at all for switch_lb_vip.

  - Everything else red is not yours: METALLB reproduces identically on unrelated PR #7433 (timed out waiting for exactly one VIP bypass lflow, test/e2e/metallb/
    e2e_test.go:1429), Kubernetes Conformance failed on a flaky session-affinity spec, and the Kube-OVN Conformance/Submariner jobs show the runner without external egress
    (pinger 1.1.1.1 at 100% loss).

pkg/controller/vip.go
```diff
@@ -94,6 +105,15 @@ func (c *Controller) handleAddVirtualIP(key string) error {
 		return err
 	}
 	portName := ovs.PodNameToPortName(vip.Name, vip.Spec.Namespace, subnet.Spec.Provider)
+	if needsMacRepair {
+		newMac, err := c.ipam.RenewNicMac(subnet.Name, vip.Name, portName)
```
zhangzujian, line 109: The returned `newMac` is discarded here. A pre-fix VIP has `Spec.MacAddress` set to the gateway MAC because the old `createOrUpdateVipCR` wrote it there. The subsequent `acquireStaticIPAddress` call still passes that stale value, and `GetStaticMac(..., checkConflict=true)` rejects it as a gateway conflict before the LSP or CR can be updated. Please carry `newMac` into the allocation (and add a regression case with `Spec.MacAddress` equal to the gateway MAC).

pkg/controller/vip.go
```diff
@@ -76,9 +76,20 @@ func (c *Controller) handleAddVirtualIP(key string) error {
...
-		// already ok
-		return nil
+		if cachedVip.Spec.Type != util.SwitchLBRuleVip {
+			return nil
+		}
+		// Pre-fix switch_lb_rule vips had their lsp mac forced to the subnet gateway
+		// mac, which collides with the gateway lsp in ovn's l2 lookup. Repair those
+		// by reallocating a unique mac; leave already-fixed vips alone.
+		gwMac := c.ipam.GetGatewayMAC(cachedVip.Spec.Subnet)
+		if gwMac == "" || gwMac != cachedVip.Status.Mac {
```
zhangzujian, line 88: This early return makes repair dependent on startup queue timing. `InitIPAM` registers existing VIPs before subnet reconciliation records `GatewayMAC`; if this VIP is processed while `GetGatewayMAC` is empty, it is dropped. `RecordGatewayMAC` does not enqueue the VIP again, so the old VIP can remain unrepaired indefinitely after a controller restart. Please requeue affected VIPs when the gateway MAC is recorded, or retry/defer this path instead of returning nil.

zhangzujian (changes requested):
Re-review at HEAD `cb6e27a15c5343e490fdd07174c6f1a3a289f0b6`. Request changes.

The fresh VIP path now uses an independent IPAM MAC and the E2E assertion was updated, but the legacy repair path still has two production correctness issues:

- It can fail on real pre-fix VIP objects because the stale gateway MAC remains in `Spec.MacAddress`.
- It can be skipped permanently on controller startup when the VIP queue runs before subnet reconciliation records the gateway MAC.

Lower-priority notes: `handleAddVirtualIP` is now over the repository's 100-line guideline, and `test/e2e/vip/e2e_test.go:117` still contains a stale shared-gateway-MAC comment.

Issue #6815, opened 2026-06-03
[BUG] ovn-central crashes permanently after rolling upgrade removes a control-plane node (CAPI)


spanudiez:
### Kube-OVN Version

v1.16.1

### Kubernetes Version

 v1.35.3+rke2r3

### Operation-system/Kernel Version

kernel 6.8.0-110-generic (Ubuntu 24.04 Noble)

### Description

During a k8s rolling upgrade of a 3 node control plane cluster , when CAPI deletes one node , the two remaining ovn-central pods enter a permanent crash loop  and will never recover. The RAFT databases on the two surviving nodes have the first server IP hardcoded in the cluster membership config. When pods restart, the STARTUP script attempts to contact all 3 RAFT peers before starting the databases, however the first server is unreachable during upgrade so the pre-start connectivity check times out and the pod exits, even though the surviving nodes should be in a valid quorum. 

### Steps To Reproduce

  Environment

  - kube-ovn v1.16.1, chart kube-ovn-v2
  - 3-node RKE2 control-plane cluster (capm3, Ubuntu 24.04) - it should happen on any type of k8s cluster
  - Nodes: server1 192.168.130.100, server2 192.168.130.101, server3 192.168.130.102
  - kube-ovn deployed as secondary CNI alongside Calico via Multus
  - OVN_LEADER_PROBE_INTERVAL=45 (issue persists with default 5)

  1. Deploy kube-ovn on a 3 node control plane cluster
  2. Trigger a rolling upgrade of the control plane 
  3. CAPI deletes one node (server1) and provisions a replacement
  4. Observe ovn-central on the two surviving nodes crashing permanently. 

### Current Behavior

Both surviving ovn central pods crash immediately after server1 is removed.
```
  ovn-central-7f89f6c9f9-bqwc9   0/1   Running   1 (33s ago)   101s   192.168.130.101   server2
  ovn-central-54596c84f8-vwlg7   0/1   Running   1 (34s ago)   101s   192.168.130.102   server3
  ovn-central-7f89f6c9f9-6xtm8   0/1   Pending   0             101s   <none>            <none>
```
Server2 logs : 
```
  PROBE_INTERVAL is set to 180000
  OVN_LEADER_PROBE_INTERVAL is set to 45
  192.168.130.101
   * ovn-northd is not running
   * ovnnb_db is not running
   * ovnsb_db is not running
  local address: tcp:[192.168.130.101]:6643
[...]

Issue #5669, opened 2025-09-02
[Feature Request] Automatic recovery for nb and sb databases


kvaps:
### Description

Hello, we faced with the issue where cluster stopped working after one specific master node become offline.

Two alive `ovn-central` couldn't create a cluster with no visible logs in stdout.

After digging I found one replica was continiously reporting into `/var/log/ovn/ovsdb-server-sb.log`:

```
2025-09-02T12:57:07.467Z|00041|raft|INFO|ssl:10.200.1.14:47338: syntax "{"cluster":"b8e67e45-61a2-4479-8e42-973ea3b1e817","from":"eb6244cf-b881-4a00-a580-1778f3dc034e","is_prevote":true,"last_log_index":2047110,"last_log_term":1259,"term":1259,"to":"e85f09c2-0ff3-4d91-8362-87e8cf22fe5c"}": syntax error: Parsing raft vote_request RPC failed: misrouted message (addressed to e85f but we're 5b4b)
```

After we dropped ovn database on two alive replicas they become online and started working properly.

But after dead master become online I obviously found that it can't join back to the cluster, because of split-brain.

What is more interested it still had some old master nodes in the db configuration:

```
kubectl -n cozy-kubeovn exec ovn-central-74dfcbfdcc-fjgr7 -c ovn-central -- ovn-appctl -t /var/run/ovn/ovnnb_db.ctl cluster/status OVN_Northbound
b3f5
Name: OVN_Northbound
Cluster ID: 2b98 (2b983e98-0e7d-46a2-a1ca-5366572db3ca)
Server ID: b3f5 (b3f53555-051b-4602-b368-81fb3deb479a)
Address: ssl:[10.200.1.15]:6643
Status: disconnected from the cluster (election timeout)
Role: candidate
Term: 1224
Leader: unknown
Vote: 4be9

Last Election started 1792 ms ago, reason: timeout
Election timer: 5000
Log: [1363290, 1381064]
Entries not yet committed: 0
Entries not yet applied: 0
Connections: ->4be9 ->bc9a (->1a8c)
Disconnections: 0
Servers:
    4be9 (4be9 at ssl:[10.200.1.14]:6643)
    bc9a (bc9a at ssl:[10.200.1.16]:6643)
    1a8c (1a8c at ssl:[10.200.1.17]:6643)
    b3f5 (b3f5 at ssl:[10.200.1.15]:6643) (self) (voted for b3f5)
[...]

PR #7408, opened 2026-09-10
feat(ovn-central) Add ovn-central-controller

andriishestakov:
Summary

Introduces a new ovn-central-controller binary that owns the full OVN central DB lifecycle (preflight, bring-up, runtime watchdog) when DYNAMIC_PEERS=true. It replaces the dynamic-peers branch of start-db.sh and the legacy kube-ovn-leader-checker for that mode, and recovers in-process to avoid kubelet restart cycles.

What's included

    New package pkg/ovn_central_controller/ (controller + kube + ovsdb helpers) and cmd/ovn_central_controller entrypoint wired into cmd/cmdmain.go.
    Lifecycle: preflight → bringUp → runtimeLoop, with recoverCluster() choosing between wipe-rejoin / reconvert / bootstrap based on peer state. Periodic leader labeling, stale-member kicking, raft header backup, and status publishing.
    start-db.sh short-circuit: execs the new binary when DYNAMIC_PEERS=true; static NODE_IPS flow unchanged.
    New readiness probe ovn-central-controller-readiness.sh: passes only when both NB and SB report cluster member with a known leader and committed log range — prevents stub pods from being treated as authoritative during AddServer.
    Helm: new DYNAMIC_PEERS / DYNAMIC_JOIN_TIMEOUT values, RBAC additions, Dockerfile copies the probe script.

### DIFF
--- a/charts/kube-ovn-v2/templates/_helpers.tpl
+++ b/charts/kube-ovn-v2/templates/_helpers.tpl
@@ -103,7 +103,11 @@ Uses Exists operator for empty/nil-value labels and In for specific values.
 Number of master nodes
 */}}
 {{- define "kubeovn.nodeCount" -}}
-  {{- len (split "," ((join "," .Values.masterNodes) | default (include "kubeovn.nodeIPs" .))) }}
+  {{- if .Values.nodeCount -}}
+    {{- .Values.nodeCount }}
+  {{- else -}}
+    {{- min 3 (len (split "," ((join "," .Values.masterNodes) | default (include "kubeovn.nodeIPs" .)))) }}
+  {{- end -}}
 {{- end -}}
 
 {{/*
--- a/dist/images/start-db.sh
+++ b/dist/images/start-db.sh
@@ -1,6 +1,15 @@
 #!/bin/bash
 set -eo pipefail
 
+# DYNAMIC_PEERS short-circuit: hand off the entire lifecycle (preflight,
+# bring-up, runtime watchdog/kicker) to the ovn-central-controller binary,
+# which replaces both this script's dynamic-peers branch and the legacy
+# kube-ovn-leader-checker. The static (NODE_IPS-based) flow below is
+# unchanged.
+if [[ "${DYNAMIC_PEERS:-false}" == "true" ]]; then
+    exec /kube-ovn/ovn-central-controller
+fi
+
 DEBUG_WRAPPER=${DEBUG_WRAPPER:-}
 ENABLE_COMPACT=${ENABLE_COMPACT:-false}
 PROBE_INTERVAL=${PROBE_INTERVAL:-180000}
--- a/pkg/ovn_central_controller/controller.go
+++ b/pkg/ovn_central_controller/controller.go
@@ -1,0 +1,972 @@
+// controller.go: orchestration of ovn-central startup + runtime loop.
+//
+// Lifecycle:
+//
+//	Run() -> parseConfig -> [preflight -> bringUp -> runtimeLoop] (retry on errRetry)
+//	  preflight:   sweep reconvert temp files, wipe kicked/stale-stub DBs.
+//	  bringUp:     start ovsdb-server, wait for cluster_member+leader. On
+//	               failure, take the bootstrap lease and recoverCluster()
+//	               picks wipe-rejoin / reconvert / bootstrap based on peer
+//	               state.
+//	  runtimeLoop: every TickInterval, watch leader health, label leaders,
+//	               kick stale members, backup raft headers, publish status.
+//	               If no-leader past NoLeaderTimeout, escalate to
+//	               recoverCluster().
+package ovn_central_controller
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync/atomic"
+	"syscall"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/klog/v2"
+)
+
+// errRetry signals "this attempt failed transiently; re-enter preflight +
+// bringUp after backoff." Returned for cases like a peer holding the
+// bootstrap lease or peers not yet ready.
+var errRetry = errors.New("retry preflight/bringUp")
+
+// heartbeatFile name (under cfg.OVNRunDir) is touched by a dedicated
+// goroutine. The liveness probe checks file mtime to detect a hung Go
+// runtime (deadlock, frozen goroutines) without conflating with
+// orchestration-state false negatives like a long reconvert.
+const (
+	heartbeatFile   = "ovn-central-controller.alive"
+	heartbeatPeriod = 5 * time.Second
+)
+
+// lifecycleState tracks in-container facts that can't be derived from
+// disk alone:
+//   - dbInherited: did the on-disk DB exist at container start? Cleared
+//     by wipeDB and by initStub fresh-create. Differentiates statusStale
+//     (data carried over) from statusJoining (fresh stub).
+//   - wasActive / activeSince: did we observe statusActive consecutively
+//     for sustainedActiveTicks? wasActive is sticky once true so a brief
+//     leader-unknown blip leaves us at leader_lost, not stale.
+//   - hasCommittedData: latched once we observed cluster_member + known
+//     leader + (role=leader OR logHigh>logLow) in BOTH NB and SB,
+//     i.e. we have replicated state even if wasActive hasn't latched
+//     yet. Distinguishes statusJoined (just-joined stub with full data)
+//     from statusJoining (empty stub). Sticky once true.
+type lifecycleState struct {
+	dbInherited      bool
+	wasActive        bool
+	hasCommittedData bool
+	activeSince      time.Time
+
+	// lastHealthyOvsdbNanos / inRecovery are read by the heartbeat goroutine
+	// concurrently with the main goroutine that writes them. atomic.
+	//   - lastHealthyOvsdbNanos: unix-nanos of the last time
+	//     updateAndComputeStatus returned statusActive. Zero == never seen
+	//     active (heartbeat treats startup time as the baseline).
+	//   - inRecovery: true while a recoverCluster call is in flight. The
+	//     heartbeat goroutine uses a larger staleness threshold during
+	//     recovery (BootstrapLeaseDuration) to avoid killing the pod
+	//     during legitimate multi-minute reconvert.
+	lastHealthyOvsdbNanos atomic.Int64
+	inRecovery            atomic.Bool
+}
+
+// sustainedActiveTicks: brief 5-10s "active" glimpses (ovsdb-server
+// reporting cluster_member while AddServer is mid-flight against a
+// quorum-less leader) shouldn't latch wasActive=true; otherwise a
+// stub-only pod could later claim leader_lost and hijack a reconvert.
+const sustainedActiveTicks = 3
+
+// updateAndComputeStatus derives our current status from on-disk state,
+// ovsdb runtime state, and lifecycle flags. Has side effects on ls:
+// advances activeSince and latches wasActive once we've been sustained-
+// active. Called from preflight, bringUp, runtimeTick, and recoverCluster.
+func updateAndComputeStatus(ctx context.Context, cfg *Config, ls *lifecycleState) status {
+	if !readDBState(ctx, nbDB(cfg)).exists && !readDBState(ctx, sbDB(cfg)).exists {
+		return ""
+	}
+	nbCS, errNB := readClusterStatus(ctx, nbDB(cfg))
+	sbCS, errSB := readClusterStatus(ctx, sbDB(cfg))
+	// Latch hasCommittedData on the first tick where BOTH DBs report
+	// "committed-active" state. This is what isCommittedActive checks:
+	// cluster_member + known leader + (role=leader OR logHigh>logLow).
+	// Crucial because wasActive needs sustainedActiveTicks (~15s) to
+	// latch, but a fresh-joined follower has full replicated data
+	// long before that. Without this flag a peer racing to recover
+	// might classify us as "no data" and bootstrap fresh on top of us.
+	if !ls.hasCommittedData &&
+		errNB == nil && errSB == nil &&
+		isCommittedActive(nbCS) && isCommittedActive(sbCS) {
+		ls.hasCommittedData = true
+	}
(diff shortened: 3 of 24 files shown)


SkalaNetworks:
Hi, this looks interesting.

Can you briefly tell us what this is trying to solve and why you need this?

andriishestakov:
This PR adds a new ovn-central-controller binary that fully owns the OVN central DB lifecycle when DYNAMIC_PEERS=true.

This is needed for reliable restart\rejoin\recreate of cluster members in environments where control-plane nodes can change dynamically (clouds), and for scaling the number of ovn-central nodes up/down while preserving DB state.

How it works:
- Lifecycle: preflight → bringUp → runtimeLoop.
- When only the local node has committed data (leader lost / stale / joined) it reconverts the raft cluster to standalone (cluster-to-standalone + create-cluster) under a lease, so at most one node can do this at a time — avoiding split-brain and it can later rejoin as new peers come up.
- Runs periodic maintenance: leader labeling, kicking stale raft members, backing up the raft header, and publishing cluster status.
- start-db.sh now execs this binary when DYNAMIC_PEERS=true; the static NODE_IPS flow is untouched.
- New readiness probe (ovn-central-controller-readiness.sh) only passes when both NB and SB report a cluster member with a known leader and a committed log range — prevents half-initialized pods from being treated as authoritative during AddServer.
- Helm chart gets new DYNAMIC_PEERS/DYNAMIC_JOIN_TIMEOUT values, RBAC for the controller, and the Dockerfile ships the new probe script.

andriishestakov:
@zbb88888 could you review this also? Thanks

SkalaNetworks:
Does this solve https://github.com/kubeovn/kube-ovn/issues/5669

andriishestakov:
And this #6815

zbb88888:
> @zbb88888 could you review this also? Thanks

Thanks for the PR, I will!

zbb88888:
@andriishestakov Hi, Thanks for the PR!


核心问题 1 / Core issue 1（中文）

 ovsdb.go:287 的 localNorthdActive() 用了 ovs-appctl -t ovn-northd status。ovs-appctl 对相对 target 会解析成 ovs_rundir()/ovn-northd.pid，而该二进制的编译默认 rundir 是
 /var/run/openvswitch（strings /usr/bin/ovs-appctl 可验证），northd 的 socket 却在 /var/run/ovn；容器里也没有设置 OVS_RUNDIR（Dockerfile 无 ENV，start-db.sh 在
 dynamicPeers 时最顶部就 exec）。结果是该函数恒为 false：

 - ovn-northd-leader 标签恒为 "false"，ovn-northd Service 永远没有 endpoint；
 - controller.go:731 的 if sbLeader && !anyNorthdActive(...) 恒成立，SB leader 每 5 秒执行一次 ovsdb-client steal <sb> ovn_northd，把正在正常工作的 northd 锁强行抢掉，形
   成持续 steal/re-acquire 抖动。

 仓库里其它所有 OVN 调用都用 ovn-appctl 或带全路径的 socket。建议改为 ovn-appctl，或显式传 ovs-appctl -t /var/run/ovn/ovn-northd.$(pid).ctl。

 核心问题 1 / Core issue 1（English）

 localNorthdActive() in ovsdb.go:287 runs ovs-appctl -t ovn-northd status. For a relative target, ovs-appctl resolves it as ovs_rundir()/ovn-northd.pid, and this binary's
 compiled default rundir is /var/run/openvswitch (verifiable via strings /usr/bin/ovs-appctl), while northd lives in /var/run/ovn. OVS_RUNDIR is never set in the image or
 pod (no ENV in the Dockerfile; start-db.sh execs the binary at the very top in dynamicPeers mode). So the function always returns false:

 - the ovn-northd-leader label is always "false" and the ovn-northd Service never gets an endpoint;
 - if sbLeader && !anyNorthdActive(...) at controller.go:731 is always true, so the SB leader runs ovsdb-client steal <sb> ovn_northd every 5s, forcibly stealing the lock
   from a healthy northd and causing a continuous steal/re-acquire churn loop.

 Every other OVN call in this repo uses ovn-appctl or a full socket path. Fix: use ovn-appctl, or pass the full path explicitly (ovs-appctl -t
 /var/run/ovn/ovn-northd.$(pid).ctl).

 ────────────────────────────────────────────────────────────────────────────────

 核心问题 2 / Core issue 2（中文）

 controller.go:140-145 的 isCommittedActive 以及 ovn-central-controller-readiness.sh 用 logHigh > logLow 判断“有已提交数据”，这与 OVS raft 语义不符。OVS 明确说明
 （ovsdb/raft.c：At some point, the server takes a snapshot: log_start=N, log_end=N.；加载已有 DB 时 raft.c:943 也直接设 log_start = log_end = snap_index +
 1；InstallSnapshot 同理 raft.c:4461）。而且快照是自动发生的，无需 ENABLE_COMPACT：ovsdb/storage.c:501-549 设定 10–20 分钟最小间隔、最长 1 天，ovsdb-server.c:361 每轮满足
 条件就调用 ovsdb_snapshot()。因此一次自动快照后，空闲成员的 Log: 会变成 [N,N]，与“全新空 stub”无法区分：

 - readiness 对“刚快照 / 刚 InstallSnapshot、之后无新写入”的 follower 返回失败，该 Pod 一直 NotReady（空闲集群可能长期如此）；
 - hasCommittedData 无法 latch，且若该容器生命周期内从未见过 logHigh > logLow，wasActive 也不会 latch，于是状态被判为 joining（在 dbInherited=false 时）或 stale。joining
   不在 recoverUnderLease 的 dataHolders 列表中，当同时没有 active+ready 的 peer 时会落到 bootstrap fresh 分支，wipe 本地 DB 重建空集群——即本 PR 想消除的丢数据类问题。

 建议不要用 log range 判断“是否有数据”。可用 ovsdb-server/get-db-storage-status，或 _Server.Database 的 leader/eid，或改为“是否曾收到快照/commit_index 是否推进过”这类一次
 性 latch 的标记来区分“空 stub”和“已同步成员”。

 核心问题 2 / Core issue 2（English）

 isCommittedActive in controller.go:140-145 and ovn-central-controller-readiness.sh treat logHigh > logLow as "has committed data", which contradicts OVS raft semantics.
 OVS states it plainly (ovsdb/raft.c: At some point, the server takes a snapshot: log_start=N, log_end=N.; loading an existing DB sets log_start = log_end = snap_index +
 1 at raft.c:943; InstallSnapshot does the same at raft.c:4461). Snapshots also happen automatically, without ENABLE_COMPACT: ovsdb/storage.c:501-549 schedules them with
 a 10–20 min minimum and a 1-day maximum, and ovsdb-server.c:361 calls ovsdb_snapshot() whenever the condition holds. So after an automatic snapshot an idle member's Log:
  becomes [N,N], indistinguishable from a brand-new empty stub:

 - readiness fails for a follower that just snapshotted or just caught up via InstallSnapshot and has seen no new write since, leaving the Pod NotReady (potentially for a
   long time in an idle cluster);
 - hasCommittedData never latches, and if this container never saw logHigh > logLow, wasActive never latches either, so the peer is classified joining (when
   dbInherited=false) or stale. joining is not in recoverUnderLease's dataHolders list, so when no peer is active+ready the code falls through to bootstrap fresh, wiping
   a DB that actually holds committed data — the exact data-loss class this PR is meant to fix.

 Recommend not using the log range as a "has data" signal. Use ovsdb-server/get-db-storage-status, or _Server.Database leader/eid, or a latched flag such as "have we ever
 received a snapshot / has commit_index ever advanced" to distinguish an empty stub from a synced member.

andriishestakov:
Thanks for the detailed review!

**Issue 1 (localNorthdActive rundir):** Checked on a live cluster (DYNAMIC_PEERS=true, 3 pods). `ovn-northd-leader` labels correctly match real northd active/standby state on all 3 pods, and the `ovn-northd` Service endpoint tracks the real active pod — this specific failure mode doesn't reproduce (the code already uses `ovn-appctl`, linked against `libovn` with its own default rundir, not `ovs-appctl`/`OVS_RUNDIR`).

While investigating we did find a related bug: `stealLock()`'s `ovsdb-client -t 1` fails on every single invocation on this cluster with `signal: alarm clock` — the internal timeout is too tight, and this reproduces the "steal churn" you described via a different root cause. We raised it to `-t 3`; it's an improvement over the guaranteed-fail `-t 1`, though we don't have a confirmed correct value yet — it's what we're going with for now.

**Issue 2 (log range as "has data"):** Confirmed — auto-snapshot happens regardless of `ENABLE_COMPACT` and collapses `Log:` to `[N,N]` for a genuinely-synced idle member, same signature as a fresh stub. We tried several replacement signals (magnitude threshold, DB file size, `status == cluster member` + known leader) and none hold up: magnitude breaks on `reconvert` (new raft term/index, real data), file size is version-dependent and an empty cluster is a legitimate state, status-based checks alone appear insufficient too. Still looking for a concrete repro to pin down the right signal — no fix yet.

andriishestakov:
Issue 1 (localNorthdActive rundir): Checked on a live cluster (DYNAMIC_PEERS=true, 3 pods). ovn-northd-leader labels correctly match real northd active/standby state on all 3 pods, and the ovn-northd Service endpoint tracks the real active pod — this specific failure mode doesn't reproduce (the code already uses ovn-appctl, linked against libovn with its own default rundir, not ovs-appctl/OVS_RUNDIR).

Issue 2 (log range as "has data"): Confirmed — auto-snapshot happens regardless of ENABLE_COMPACT and collapses Log: to [N,N] for a genuinely-synced idle member, same signature as a fresh stub. I tried several replacement signals (magnitude threshold, DB file size, status == 
  cluster member + known leader) and none hold up: magnitude breaks on reconvert (new raft term/index, real data), file size is version-dependent and an empty cluster is a legitimate state, status-based checks alone appear insufficient too. I suggests to keep it as is until find more proper way.

pkg/ovn_central_controller/controller.go
```diff
@@ -0,0 +1,972 @@
...
+	if err := waitForLeader(ctx, cfg, cfg.StaggerTimeout); err != nil {
+		return fmt.Errorf("waitForLeader after %s: %w", op, err)
+	}
+	if err := refreshStatus(ctx, cfg, kc, updateAndComputeStatus(ctx, cfg, ls)); err != nil {
+		klog.Warningf("label refresh post-recovery: %v", err)
+	}
+	if err := startNorthd(cfg, op == "bootstrap", peers); err != nil {
+		return err
+	}
+	if err := waitForSelfReady(ctx, cfg, kc, cfg.SelfReadyTimeout); err != nil {
```
zhangzujian, line 570: A readiness timeout cannot be treated as a successful recovery. This function has already published the lifecycle status and then releases the bootstrap Lease even when this pod is still NotReady. Another pod only treats `active+ready` as authoritative, so it can acquire the lease and enter the fresh-bootstrap branch while this pod may still hold the recovered data. Return an error/retry (and keep a non-authoritative status) until Pod.Ready is true.

pkg/ovn_central_controller/controller.go
```diff
@@ -0,0 +1,972 @@
...
+// (we're the leader OR our raft log has entries beyond the initial
+// stub). The log check filters out the optimistic "cluster_member"
+// status that ovsdb-server reports while AddServer is still mid-flight
+// against a quorum-less leader -- without it, a stub-only pod could
+// briefly look healthy and trick peers into a wipe-rejoin.
+func isCommittedActive(cs clusterStatus) bool {
+	if cs.status != "cluster member" || !isKnownLeader(cs.leader) {
+		return false
+	}
+	return cs.role == "leader" || cs.logHigh > cs.logLow
```
zhangzujian, line 144: `logHigh > logLow` is not a durable indicator that a follower has committed data. OVS raft snapshots and InstallSnapshot can legitimately leave an idle, fully-synchronized follower with `Log: [N,N]`, indistinguishable from a fresh stub. The readiness probe and `hasCommittedData` latch will then misclassify that follower and recovery can wipe it via fresh bootstrap. Please use a signal that survives snapshots, or persist an explicit joined/committed state.

dist/images/start-db.sh
```diff
@@ -1,6 +1,15 @@
 #!/bin/bash
 set -eo pipefail
 
+# DYNAMIC_PEERS short-circuit: hand off the entire lifecycle (preflight,
+# bring-up, runtime watchdog/kicker) to the ovn-central-controller binary,
+# which replaces both this script's dynamic-peers branch and the legacy
+# kube-ovn-leader-checker. The static (NODE_IPS-based) flow below is
+# unchanged.
+if [[ "${DYNAMIC_PEERS:-false}" == "true" ]]; then
```
zhangzujian, line 9: This early `exec` bypasses the legacy TLS rotation process started later in this script (`kube-ovn-tls-reload.sh ovn-central`). The new controller has no equivalent certificate-change watcher/reload path, so `ENABLE_SSL=true` dynamic-peer deployments will retain stale OVN DB/northd certificates after rotation. Preserve the reload lifecycle or implement an equivalent in-process reload.

charts/kube-ovn-v2/templates/_helpers.tpl
```diff
@@ -103,7 +103,11 @@ Uses Exists operator for empty/nil-value labels and In for specific values.
 Number of master nodes
 */}}
 {{- define "kubeovn.nodeCount" -}}
-  {{- len (split "," ((join "," .Values.masterNodes) | default (include "kubeovn.nodeIPs" .))) }}
+  {{- if .Values.nodeCount -}}
+    {{- .Values.nodeCount }}
+  {{- else -}}
+    {{- min 3 (len (split "," ((join "," .Values.masterNodes) | default (include "kubeovn.nodeIPs" .)))) }}
```
zhangzujian, line 109: This changes the global replica-count helper for every deployment, not only `central.dynamicPeers`. A static configuration with four master/node IPs now renders both ovn-central and kube-ovn-controller with three replicas while `NODE_IPS` still contains all four. That is a regression against the stated unchanged static flow; scope the minimum/quorum default to dynamic peers or preserve the existing count for static mode.

zhangzujian (changes requested):
I reviewed the current HEAD with both the recovery contract and repository standards in mind. The following issues need to be addressed before merge.

Additional validation: `go test ./pkg/ovn_central_controller ./cmd/ovn_central_controller`, shell `bash -n`, Helm rendering, and `git diff --check` pass, but the PR's current DCO and required x86 gate are still not satisfied and several Conformance/Network Policy/NAT E2E checks are failing.

The most serious unresolved risk is the `Log: [low, high]` test used to distinguish a committed follower from an empty stub. OVS snapshots can legitimately collapse this range to `[N,N]`; this can leave a real follower NotReady and route recovery into fresh bootstrap, which can wipe committed data. The recovery lease timeout path is also unsafe: releasing the lease after `waitForSelfReady` fails can let another pod bootstrap while this pod has already published `active`. Please fix these before relying on this controller for dynamic recovery.

The dynamic start path also bypasses the existing TLS reload helper, and the global node-count change alters static clusters with more than three masters. Please preserve certificate rotation and scope the replica-count behavior to dynamic peers only.

PR #7415, opened 2026-09-11
fix(controller): configure VMI migration ports while pending

### LINKED ISSUE #7416: [BUG] KubeVirt migration with hotplug volumes can stall in Pending
## Kube-OVN Version

- Affected behavior exists in the current `master` implementation before the proposed fix.
- It was also observed in a deployment derived from Kube-OVN v1.15 with OVN 25.03.

## Kubernetes Version

- Kubernetes v1.35.1
- KubeVirt v1.8.2

## Operation-system/Kernel Version

- kind nodes: `kindest/node:v1.35.1`
- The exact host kernel output from the original affected environment was not retained.

## Description

A KubeVirt live migration with a hotplug volume can remain in `Pending` because Kube-OVN configures the VMI logical switch port for OVN multi-chassis migration only in `Scheduling`.

With hotplug volumes, KubeVirt waits for the target launcher network and attachment preparation before advancing from `Pending` to `Scheduling`. The target launcher's CNI ADD already needs the multi-chassis configuration, creating this cycle:

```text
target launcher network Ready
        -> hotplug attachment Ready
        -> VMIM Pending -> Scheduling
        -> Kube-OVN sets multi-chassis options
        -> target launcher network Ready
```

The target CNI can therefore fail with:

```text
ovs interface ... is not ready after 30s
```

## Steps To Reproduce

1. Enable `--enable-live-migration-optimize=true`.
2. Run KubeVirt with Kube-OVN on at least two nodes.
3. Start a VMI with an RWX root volume and hotplug an RWX data volume.
4. Create a `VirtualMachineInstanceMigration` targeting the other node.
5. Observe the VMIM remaining in `Pending` while target CNI waits for `external_ids:ovn-installed`.

The old behavior is timing-dependent, so several attempts may be required.

## Current Behavior

The target launcher is scheduled, but CNI may time out waiting for a stable `ovn-installed=true`. The launcher does not become ready, hotplug attachment preparation cannot complete, and the VMIM never reaches the `Scheduling` phase handled by Kube-OVN.

`MigrationState` can still be nil before target-pod handoff, or may temporarily refer to an earlier migration. Returning early solely because it is nil also prevents setup even though both nodes are already available from:

```text
source = VMI.status.nodeName
target = target virt-launcher.spec.nodeName
```

If the target launcher or either node assignment is not visible during the first reconciliation, returning nil can lose the setup opportunity. Pod creation and scheduling do not guarantee another VMIM phase update, so these transient conditions need a rate-limited retry.

## KubeVirt Migration State Machine

```mermaid
stateDiagram-v2
    [*] --> Unset
    Unset --> Pending: local migration accepted
    Unset --> WaitingForSync: decentralized target
    Unset --> Synchronizing: decentralized source
    WaitingForSync --> Pending: synchronization completed
    Synchronizing --> Pending: synchronization completed

    Pending --> Pending: create and prepare target launcher
    Pending --> Pending: prepare hotplug attachment
    Pending --> Scheduling: launcher and attachment prerequisites satisfied
    Scheduling --> Scheduled: target launcher and attachments Ready
    Scheduled --> PreparingTarget: target-pod handoff
    PreparingTarget --> TargetReady: target preparation completed
    TargetReady --> Running: migration started
    Running --> Succeeded: completed=true
    Running --> Failed: failed=true

    Pending --> Failed
    Scheduling --> Failed
    Scheduled --> Failed
    PreparingTarget --> Failed
    TargetReady --> Failed
    Succeeded --> [*]
    Failed --> [*]
```

All local migrations pass through `Pending`; hotplug volumes make this phase persistent because attachment preparation depends on target launcher readiness. `Pending` is therefore the required early network-configuration point. `Scheduling` remains necessary as an idempotent retry path, while `Succeeded` and `Failed` retain their existing cleanup responsibilities.

## Complete Flow Before the Change

```mermaid
sequenceDiagram
    autonumber
    participant KV as KubeVirt migration controller
    participant M as VMIM
    participant TL as Target virt-launcher
    participant K as Target kubelet
    participant CNI as Kube-OVN CNI
    participant OVC as Target ovn-controller and OVS
    participant KO as Kube-OVN controller
    participant NB as OVN NB LSP
    participant HP as Hotplug attachment preparation

    KV->>M: Set phase to Pending
    KV->>TL: Create target virt-launcher
    K->>CNI: Start CNI ADD
    CNI->>OVC: Create local OVS interface
    KO-->>M: Reconcile Pending VMIM
    KO-->>KO: Skip because old handler only configures Scheduling
    Note over NB: requested-chassis=source,target is not configured

    loop Poll every 500 ms for up to 30 seconds
        CNI->>OVC: Read external_ids:ovn-installed
        OVC-->>CNI: Usually absent and may be true for only a few milliseconds
    end

    alt CNI poll hits a transient true window
        CNI-->>K: CNI ADD succeeds
        K-->>TL: Target launcher becomes Ready
        KV->>HP: Prepare hotplug attachment
        HP-->>KV: Attachment becomes Ready
        KV->>M: Pending -> Scheduling
        KO->>NB: Set requested-chassis=source,target and activation-strategy=rarp
        KV->>M: Scheduling -> Scheduled -> PreparingTarget -> TargetReady -> Running
        alt Migration succeeds
            KV->>M: Running -> Succeeded
            KO->>NB: Converge requested-chassis to target
        else Migration fails
            KV->>M: Running -> Failed
            KO->>NB: Roll requested-chassis back to source
        end
    else CNI misses every transient true window
        CNI-->>K: ovs interface is not ready after 30s
        K-->>TL: FailedCreatePodSandBox and launcher is not Ready
        KV-->>HP: Attachment preparation cannot complete
        KV-->>M: Remain Pending until retry, timeout, or cancellation
    end
```

## Complete Flow After the Change

```mermaid
sequenceDiagram
    autonumber
    participant KV as KubeVirt migration controller
    participant M as VMIM
    participant TL as Target virt-launcher
    participant S as Scheduler
    participant KO as Kube-OVN controller
    participant NB as OVN NB LSP
    participant K as Target kubelet
    participant CNI as Kube-OVN CNI
    participant OVC as Target ovn-controller and OVS
    participant HP as Hotplug attachment preparation

    KV->>M: Set phase to Pending
    KV->>TL: Create target virt-launcher
    S->>TL: Assign target spec.nodeName
    K->>CNI: Start CNI ADD and wait for ovn-installed
    KO-->>M: Reconcile Pending VMIM

    alt Target launcher or node assignment is not visible yet
        KO-->>KO: Return an error and AddRateLimited
        loop Rate-limited reconciliation
            KO->>TL: Retry until source and target nodes are available
        end
    end

    KO->>KO: Derive source from VMI.status.nodeName
    KO->>KO: Derive target from launcher.spec.nodeName
    Note over KO: Current MigrationState may still be nil
    KO->>NB: Set requested-chassis=source,target
    KO->>NB: Set activation-strategy=rarp
    NB-->>OVC: Allow target chassis to prepare the shared LSP
    OVC-->>CNI: external_ids:ovn-installed becomes stable
    CNI-->>K: CNI ADD succeeds
    K-->>TL: Target launcher becomes Ready
    KV->>HP: Prepare hotplug attachment
    HP-->>KV: Attachment becomes Ready
    KV->>M: Pending -> Scheduling
    KO->>NB: Idempotently retain migration options
    KV->>M: Scheduling -> Scheduled
    KV->>M: Scheduled -> PreparingTarget -> TargetReady -> Running

    alt Migration succeeds
        KV->>M: Running -> Succeeded
        KO->>NB: Remove activation-strategy and keep target chassis
    else Migration fails
        KV->>M: Running -> Failed
        KO->>NB: Remove activation-strategy and restore source chassis
    end
```

The controller and target CNI may run concurrently. The required ordering is that multi-chassis options are installed during `Pending` before the CNI readiness deadline, not that controller reconciliation must finish before CNI ADD starts.

## Why the Failure Is Intermittent

Affected logs captured two short intervals in which the target interface received `ovn-installed=true` and then lost it:

```text
18:49:20.285 Changing chassis ... source to target
18:49:20.294 Setting lport ... ovn-installed in OVS
18:49:20.303 Removing iface ... ovn-installed in OVS

18:49:22.288 Changing chassis ... source to target
18:49:22.347 Setting lport ... ovn-installed in OVS
18:49:22.353 Removing iface ... ovn-installed in OVS
```

The ready windows lasted 9 ms and 6 ms, while Kube-OVN CNI polls every 500 ms for up to 30 seconds. The old flow succeeds only if a poll happens to hit one of these short windows. A successful attempt then reaches `Scheduling`, where Kube-OVN finally installs the intended multi-chassis options.

A successful historical attempt confirms that multi-chassis setup occurred only after CNI had already reported the interface ready:

```text
18:44:22.506 target CNI ADD started
18:44:40.631 target OVS interface became Ready
18:44:56.052 requested-chassis=source,target was configured
```

`Setting ... ovn-installed` proves that a flow generation reached installation acknowledgement. The generic `Removing iface ... ovn-installed` message proves that the flag was unstable, but does not uniquely identify whether removal came from a binding eligibility change or the `OIF_REM_OLD_OVN_INST` path for replacing an old ready flag.

A stable source-only `requested-chassis` does not allow the target chassis to claim the port. The retained logs do not contain synchronized NB options, SB binding fields, and OVS interface state, so they cannot identify which field produced each short eligibility window.

Lost-port GC may amplify later failures, but the first 30-second CNI timeout occurred before GC deletion in the failure sequence. It is not the origin of this dependency cycle.

## Expected Behavior

Kube-OVN should configure the logical switch port while the VMIM is `Pending`, once the target launcher has a node assignment:

```text
requested-chassis=<source>,<target>
activation-strategy=rarp
```

`Scheduling` should remain as an idempotent retry path because migrations without hotplug volumes can pass through `Pending` quickly, and the first `Pending` reconciliation may occur before the target launcher is visible or assigned.

The migration handler should also:

- allow `Pending` and `Scheduling` to proceed before the current `MigrationState` is available;
- use `VMI.status.nodeName` as the source before migration-state handoff;
- requeue while the target launcher or node assignments are not yet visible;
- preserve the existing `Succeeded` and `Failed` cleanup behavior.

## Scope and Compatibility

- This affects only installations with live migration optimization enabled. The command-line, Helm, Helm v2, and install-script defaults are currently `true`.
- Current `master` builds against `kubevirt.io/api v1.9.0`; the observed runtime used KubeVirt v1.8.2. The relevant phases and migration-state fields are compatible.
- No CRD, public API, configuration field, or default value change is required.

jimyag:
# Pull Request

## What type of this PR

- [ ] Features
- [x] Bug fixes
- [ ] Docs
- [x] Tests

## Summary

KubeVirt creates the target `virt-launcher` while a VMI migration is still `Pending`. With hotplug volumes, the attachment pod is created only after that launcher becomes ready, and the migration cannot advance to `Scheduling` until the attachment pod is ready.

Kube-OVN previously configured OVN multi-chassis options only in `Scheduling`. The target launcher's CNI ADD needs those options to observe a stable `external_ids:ovn-installed=true`, creating a dependency cycle that can leave the migration in `Pending` with `ovs interface ... is not ready after 30s`.

This change:

- configures migration LSPs in both `Pending` and `Scheduling`;
- derives the source node from `VMI.status.nodeName` before the current `MigrationState` is available;
- selects the target launcher with both the migration job UID and `kubevirt.io=virt-launcher`, excluding hotplug attachment pods;
- requeues while the target launcher or its node assignment is not yet visible.

`Scheduling` remains an idempotent retry path, and terminal migration cleanup is unchanged.

## Why the failure was intermittent

In failed runs, the target interface exposed `ovn-installed=true` for only 6-9 ms before the flag was removed. Kube-OVN CNI polls the flag every 500 ms, so the old flow succeeded only when a poll happened to hit one of those short windows. Configuring `requested-chassis=source,target` in `Pending` removes that timing dependency.

## Verification

- `go test ./pkg/controller -count=1`
- `CI=true make lint`
- The equivalent downstream patch was validated in a fresh two-node kind cluster with KubeVirt v1.8.2: 12/12 bidirectional migrations with RWX root and hotplug data volumes succeeded, including after retaining 12 completed launcher pods.
- No `ovs interface ... is not ready after 30s` errors were observed after the fix.

Fixes #7416.

### DIFF
--- a/pkg/controller/kubevirt.go
+++ b/pkg/controller/kubevirt.go
@@ -144,6 +144,17 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
 		// needed to roll back those options.
 		switch vmiMigration.Status.Phase {
 		case kubevirtv1.MigrationFailed:
+			// A stale Failed event can be processed after a newer migration has already
+			// pinned the same ports. Do not let the old event roll back the newer options.
+			hasUnfinishedMigration, err := c.hasUnfinishedVMIMigration(vmiMigration)
+			if err != nil {
+				return err
+			}
+			if hasUnfinishedMigration {
+				klog.Infof("VirtualMachineInstanceMigration %s is stale because another migration for VMI %s is unfinished, skipping cleanup",
+					key, vmiMigration.Spec.VMIName)
+				return nil
+			}
 			srcNodeName = vmi.Status.NodeName
 		case kubevirtv1.MigrationSucceeded:
 			klog.V(3).Infof("VirtualMachineInstanceMigration %s migration state is Succeeded but VMI migration state is stale or nil, skipping", key)
@@ -251,6 +262,32 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
 		}
 	}
 	return nil
+}
+
+// hasUnfinishedVMIMigration reports whether another migration for the same VMI is still active.
+// Terminal cleanup must yield to any such migration because its OVN options may already own the ports.
+func (c *Controller) hasUnfinishedVMIMigration(vmiMigration *kubevirtv1.VirtualMachineInstanceMigration) (bool, error) {
+	if c.vmiMigrationIndexer == nil {
+		return false, nil
+	}
+
+	key := fmt.Sprintf("%s/%s", vmiMigration.Namespace, vmiMigration.Spec.VMIName)
+	migrations, err := c.vmiMigrationIndexer.ByIndex(informer.ByVMINameIndex, key)
+	if err != nil {
+		return false, fmt.Errorf("failed to find migrations for VMI %s: %w", key, err)
+	}
+
+	for _, obj := range migrations {
+		migration, ok := obj.(*kubevirtv1.VirtualMachineInstanceMigration)
+		if !ok {
+			return false, fmt.Errorf("unexpected object type %T in VMI migration index", obj)
+		}
+		if migration.UID != vmiMigration.UID && !migration.IsFinal() {
+			return true, nil
+		}
+	}
+
+	return false, nil
 }
 
 func (c *Controller) isKubevirtCRDInstalled() (bool, error) {
--- a/pkg/controller/kubevirt.go
+++ b/pkg/controller/kubevirt.go
@@ -105,7 +105,11 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
 		utilruntime.HandleError(fmt.Errorf("failed to get VMI migration by key %s: %w", key, err))
 		return err
 	}
-	if vmiMigration.Status.MigrationState == nil {
+	// MigrationState may still be nil before target virt-launcher pod handoff. Pending and
+	// Scheduling can derive both nodes without it, so allow network setup to proceed.
+	if vmiMigration.Status.MigrationState == nil &&
+		vmiMigration.Status.Phase != kubevirtv1.MigrationPending &&
+		vmiMigration.Status.Phase != kubevirtv1.MigrationScheduling {
 		klog.V(3).Infof("VirtualMachineInstanceMigration %s migration state is nil, skipping", key)
 		return nil
 	}
@@ -155,10 +159,17 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
 	klog.Infof("collected port names of vmi %s, port names are %v", vmi.Name, strings.Join(portNames, ", "))
 
 	switch vmiMigration.Status.Phase {
-	case kubevirtv1.MigrationScheduling:
+	case kubevirtv1.MigrationPending, kubevirtv1.MigrationScheduling:
+		// KubeVirt creates the target virt-launcher in Pending and, for hotplug volumes,
+		// creates its attachment pod only after the launcher is ready. An attachment pod
+		// then gates Pending -> Scheduling, and both pods gate Scheduling -> Scheduled.
+		// Configure Pending to unblock launcher CNI and keep Scheduling for retries.
+		// Hotplug attachment pods share MigrationJobLabel, so AppLabel must distinguish
+		// the target virt-launcher whose NodeName identifies the migration target.
 		selector, err := metav1.LabelSelectorAsSelector(&metav1.LabelSelector{
 			MatchLabels: map[string]string{
 				kubevirtv1.MigrationJobLabel: string(vmiMigration.UID),
+				kubevirtv1.AppLabel:          "virt-launcher",
 			},
 		})
 		if err != nil {
@@ -167,52 +178,52 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
 			return err
 		}
 
-		pods, err := c.podsLister.Pods(vmiMigration.Namespace).List(selector)
+		launcherPods, err := c.podsLister.Pods(vmiMigration.Namespace).List(selector)
 		if err != nil {
-			err = fmt.Errorf("failed to list pods with migration job UID %s: %w", vmiMigration.UID, err)
+			err = fmt.Errorf("failed to list target launcher pods with migration job UID %s: %w", vmiMigration.UID, err)
 			klog.Error(err)
 			return err
 		}
 
-		if len(pods) > 0 {
-			targetPod := pods[0]
-			// During MigrationScheduling phase, use vmi.Status.NodeName if SourceNode is empty
-			// because vmi.Status.MigrationState may not be fully synchronized yet
+		if len(launcherPods) > 0 {
+			targetLauncherPod := launcherPods[0]
+			// Before MigrationState is synchronized, use the VMI's current node as the source.
 			sourceNode := srcNodeName
 			if sourceNode == "" {
 				sourceNode = vmi.Status.NodeName
 			}
 
-			if sourceNode == "" || targetPod.Spec.NodeName == "" || sourceNode == targetPod.Spec.NodeName {
-				klog.Warningf("VM pod %s/%s migration setup skipped, source node: %s, target node: %s (migration job UID: %s)",
-					targetPod.Namespace, targetPod.Name, sourceNode, targetPod.Spec.NodeName, vmiMigration.UID)
-				// The source or target node may not be known yet while the migration is still
-				// in the Scheduling phase; this produces no further phase-change event, so
-				// return an error to requeue with rate limiting and retry once scheduling
-				// completes, instead of dropping the event.
-				// https://github.com/kubeovn/kube-ovn/issues/6823
-				if sourceNode == "" || targetPod.Spec.NodeName == "" {
-					return fmt.Errorf("VM pod %s/%s migration setup deferred, source node %q, target node %q not ready yet (migration job UID %s)",
-						targetPod.Namespace, targetPod.Name, sourceNode, targetPod.Spec.NodeName, vmiMigration.UID)
-				}
+			if sourceNode == "" || targetLauncherPod.Spec.NodeName == "" {
+				klog.Warningf("target launcher pod %s/%s migration setup skipped, source node: %s, target node: %s (migration job UID: %s)",
+					targetLauncherPod.Namespace, targetLauncherPod.Name, sourceNode, targetLauncherPod.Spec.NodeName, vmiMigration.UID)
+				// Pod scheduling may complete without another VMIM phase update. Requeue until
+				// both node assignments are visible instead of dropping the pending setup.
+				return fmt.Errorf("target launcher pod %s/%s migration setup deferred, source node %q, target node %q not ready yet (migration job UID %s)",
+					targetLauncherPod.Namespace, targetLauncherPod.Name, sourceNode, targetLauncherPod.Spec.NodeName, vmiMigration.UID)
+			}
+			if sourceNode == targetLauncherPod.Spec.NodeName {
+				klog.Warningf("target launcher pod %s/%s migration setup skipped, source and target node are both %s (migration job UID: %s)",
+					targetLauncherPod.Namespace, targetLauncherPod.Name, sourceNode, vmiMigration.UID)
 				return nil
 			}
 
-			klog.Infof("VM pod %s/%s is migrating from %s to %s (migration job UID: %s)",
-				targetPod.Namespace, targetPod.Name, sourceNode, targetPod.Spec.NodeName, vmiMigration.UID)
+			klog.Infof("target launcher pod %s/%s is migrating from %s to %s (migration job UID: %s)",
+				targetLauncherPod.Namespace, targetLauncherPod.Name, sourceNode, targetLauncherPod.Spec.NodeName, vmiMigration.UID)
 
 			for _, portName := range portNames {
-				if err := c.OVNNbClient.SetLogicalSwitchPortMigrateOptions(portName, sourceNode, targetPod.Spec.NodeName); err != nil {
+				if err := c.OVNNbClient.SetLogicalSwitchPortMigrateOptions(portName, sourceNode, targetLauncherPod.Spec.NodeName); err != nil {
 					err = fmt.Errorf("failed to set migrate options for VM pod lsp %s: %w", portName, err)
 					klog.Error(err)
 					return err
 				}
-				klog.Infof("successfully set migrate options for lsp %s from %s to %s", portName, sourceNode, targetPod.Spec.NodeName)
+				klog.Infof("successfully set migrate options for lsp %s from %s to %s", portName, sourceNode, targetLauncherPod.Spec.NodeName)
 			}
 		} else {
-			klog.Warningf("target pod not yet created for migration job UID %s in phase %s, waiting for pod creation",
+			klog.Warningf("target launcher pod not yet created for migration job UID %s in phase %s, waiting for pod creation",
 				vmiMigration.UID, vmiMigration.Status.Phase)
-			return nil
+			// Target launcher pod creation does not necessarily change the VMIM phase.
+			// Requeue so the same Pending migration is retried after the pod appears.
+			return fmt.Errorf("target launcher pod not yet created for migration job UID %s in phase %s", vmiMigration.UID, vmiMigration.Status.Phase)
 		}
 	case kubevirtv1.MigrationSucceeded:
 		for _, portName := range portNames {
--- a/pkg/controller/kubevirt.go
+++ b/pkg/controller/kubevirt.go
@@ -106,10 +106,12 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
 		return err
 	}
 	// MigrationState may still be nil before target virt-launcher pod handoff. Pending and
-	// Scheduling can derive both nodes without it, so allow network setup to proceed.
+	// Scheduling can derive both nodes without it, so allow network setup to proceed. Failed
+	// migrations also need to proceed so options configured while Pending can be cleaned up.
 	if vmiMigration.Status.MigrationState == nil &&
 		vmiMigration.Status.Phase != kubevirtv1.MigrationPending &&
-		vmiMigration.Status.Phase != kubevirtv1.MigrationScheduling {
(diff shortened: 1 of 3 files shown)


pkg/controller/kubevirt.go
```diff
@@ -105,7 +105,11 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
 		utilruntime.HandleError(fmt.Errorf("failed to get VMI migration by key %s: %w", key, err))
 		return err
 	}
-	if vmiMigration.Status.MigrationState == nil {
+	// MigrationState may still be nil before target virt-launcher pod handoff. Pending and
+	// Scheduling can derive both nodes without it, so allow network setup to proceed.
+	if vmiMigration.Status.MigrationState == nil &&
```
zhangzujian, line 111: Blocking: this early return now conflicts with the new Pending write path. When a Pending VMIM with nil status.migrationState has a target launcher, lines 213-220 set OVN migration options. If KubeVirt then aborts or fails before handoff (for example, target pod disappearance or cancellation), KubeVirt v1.8.2 can set the VMIM phase to Failed without populating status.migrationState. The next reconcile returns here, so the Failed cleanup below never runs and requested-chassis/activation-strategy remain on the LSP. Please make terminal cleanup work for this path (for example, derive the source from VMI.status.nodeName or otherwise track that options were installed) and add a regression test for Pending -> Failed with a nil migration state.

zhangzujian (changes requested):
Request changes: I found one P1 lifecycle bug in the new Pending reconciliation path. The inline comment explains the failure mode. The controller package tests and formatting checks pass, but this terminal-state cleanup case needs to be fixed and covered before merge.

pkg/controller/kubevirt.go
jimyag: Thanks for pointing this out.

I reproduced the lifecycle in my local kind cluster with KubeVirt:

1. The migration reached Pending and the target virt-launcher pod was created.
2. The OVN LSP contained migration options for source -> target.
3. I intentionally updated the VMIM status to Failed with migrationState set to nil, reproducing the edge case from this review.
4. The controller reconciled the Failed migration and reset the LSP back to the source node, removing the migration activation strategy.

I also added a regression test covering the Pending -> Failed transition with a nil MigrationState.

pkg/controller/kubevirt.go
```diff
@@ -133,10 +139,14 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
...
 		}
-		// If we're at an end state and the vmi migration state is stale or nil, we're probably looking at an old migration
-		// either way, we can't proceed since we don't have the source and target nodes for resetting the migrate options
-		if vmiMigration.Status.Phase == kubevirtv1.MigrationSucceeded || vmiMigration.Status.Phase == kubevirtv1.MigrationFailed {
-			klog.V(3).Infof("VirtualMachineInstanceMigration %s migration state is Succeeded/Failed but VMI migration state is stale or nil, skipping", key)
+		// A failed migration may have configured options while Pending, before the VMI
+		// migration state was populated. The VMI's current node is still the source
+		// needed to roll back those options.
+		switch vmiMigration.Status.Phase {
+		case kubevirtv1.MigrationFailed:
```
zhangzujian, line 146: Blocking: this Failed/nil-state branch does not establish that the LSP options still belong to this failed migration. For example, migration A can configure source,target-A, fail before MigrationState is populated, and have its Failed event delayed while migration B configures source,target-B. When A's handler runs, it uses the current VMI node and the Reset call below rewrites every LSP back to source, clobbering B's active options. The old stale/nil terminal-state guard avoided this. Please add an ownership/current-migration check (or equivalent serialization/provenance) before resetting and cover an older Failed migration racing with a newer Pending migration.

zhangzujian (changes requested):
Request changes: the previous nil-MigrationState cleanup issue is addressed, but the new Failed cleanup path introduces a migration ownership race. A delayed cleanup for an older failed migration can overwrite OVN options installed by a newer migration for the same VMI. Please guard cleanup against a newer/current migration and add an interleaving regression test.

pkg/controller/kubevirt.go
```diff
@@ -105,7 +105,13 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
 		utilruntime.HandleError(fmt.Errorf("failed to get VMI migration by key %s: %w", key, err))
 		return err
 	}
-	if vmiMigration.Status.MigrationState == nil {
+	// MigrationState may still be nil before target virt-launcher pod handoff. Pending and
+	// Scheduling can derive both nodes without it, so allow network setup to proceed. Failed
+	// migrations also need to proceed so options configured while Pending can be cleaned up.
+	if vmiMigration.Status.MigrationState == nil &&
```
zhangzujian, line 111: Blocking: Failed + nil-MigrationState cleanup now depends on the VMI GET immediately below succeeding. If Pending reconciliation has already written requested-chassis/activation-strategy and the VMI is deleted (or temporarily NotFound) before the Failed VMIM event is processed, this returns before the cleanup at lines 254-262. If the VMIM is then removed, the LSP options can be left behind permanently. Please make the Failed cleanup path handle a missing VMI (using persisted/OVN state or another source of the original node) and add a VMI-disappeared regression test.

pkg/controller/kubevirt.go
```diff
@@ -133,10 +139,25 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
...
-		if vmiMigration.Status.Phase == kubevirtv1.MigrationSucceeded || vmiMigration.Status.Phase == kubevirtv1.MigrationFailed {
-			klog.V(3).Infof("VirtualMachineInstanceMigration %s migration state is Succeeded/Failed but VMI migration state is stale or nil, skipping", key)
+		// A failed migration may have configured options while Pending, before the VMI
+		// migration state was populated. The VMI's current node is still the source
+		// needed to roll back those options.
+		switch vmiMigration.Status.Phase {
+		case kubevirtv1.MigrationFailed:
+			// A stale Failed event can be processed after a newer migration has already
+			// pinned the same ports. Do not let the old event roll back the newer options.
+			hasUnfinishedMigration, err := c.hasUnfinishedVMIMigration(vmiMigration)
```
zhangzujian: Blocking: this ownership check only sees the informer indexer. A newer migration can already exist in the API and have (or be about to have) valid LSP options while its informer Add event has not updated this indexer yet. The delayed Failed worker for the older migration then sees no unfinished peer and resets the shared LSPs, clobbering the newer migration. Please use an API/resource-version or persisted ownership check (or equivalent serialization) and add a test covering informer lag between the newer migration's creation and indexer update.

zhangzujian (changes requested):
Request changes: the previous stale Failed-migration race is partially addressed, but two P1 cleanup/ownership hazards remain in the latest implementation. Please fix both and add the corresponding regression tests before merge.

zhangzujian:
unittest failed:

```txt
--- FAIL: TestOvnClientTestSuite (4.28s)
    --- FAIL: TestOvnClientTestSuite/Test_GetKubeOvnChassises (5.00s)
        ovn-sb-chassis_test.go:473: 
            	Error Trace:	/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/ovn-sb-chassis_test.go:473
            	            				/home/runner/work/kube-ovn/kube-ovn/pkg/ovs/ovn-nb-suite_test.go:1198
            	            				/opt/hostedtoolcache/go/1.27.1/x64/src/runtime/asm_amd64.s:1264
            	Error:      	Condition never satisfied
            	Test:       	TestOvnClientTestSuite/Test_GetKubeOvnChassises
```

pkg/controller/kubevirt.go
```diff
@@ -105,13 +105,33 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
...
+	// migrations also need to proceed so options configured while Pending can be cleaned up.
+	if vmiMigration.Status.MigrationState == nil &&
+		vmiMigration.Status.Phase != kubevirtv1.MigrationPending &&
+		vmiMigration.Status.Phase != kubevirtv1.MigrationScheduling &&
+		vmiMigration.Status.Phase != kubevirtv1.MigrationFailed {
 		klog.V(3).Infof("VirtualMachineInstanceMigration %s migration state is nil, skipping", key)
 		return nil
 	}
+	if vmiMigration.Status.Phase == kubevirtv1.MigrationFailed {
+		hasUnfinishedMigration, err := c.hasUnfinishedVMIMigration(vmiMigration)
```
zhangzujian, line 119: P1: this ownership check is not atomic with the reset below. A newer migration can be created after this API List returns and before the Failed cleanup runs, so the older migration can still overwrite the newer migration’s OVN options. Please add an ownership/provenance check or equivalent per-VMI serialization immediately before reset.

pkg/controller/kubevirt.go
```diff
@@ -236,6 +267,49 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
...
+
+	for _, lsp := range lsps {
+		if lsp.Options["activation-strategy"] != "rarp" {
+			continue
+		}
+		nodes := strings.Split(lsp.Options["requested-chassis"], ",")
+		if len(nodes) != 2 || nodes[0] == "" || nodes[1] == "" || nodes[0] == nodes[1] {
+			continue
+		}
+		if err := c.OVNNbClient.ResetLogicalSwitchPortMigrateOptions(lsp.Name, nodes[0], "", true); err != nil {
```
zhangzujian, line 287: P1: when the VMI is gone, this path blindly resets any LSP carrying migration options; it has no migration UID/provenance check. A newer migration can install these options after the earlier check and be clobbered here. Please make this cleanup ownership-safe and cover the interleaving case.

pkg/controller/kubevirt.go
```diff
@@ -105,13 +105,33 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
...
+		if hasUnfinishedMigration {
+			klog.Infof("VirtualMachineInstanceMigration %s is stale because another migration for VMI %s is unfinished, skipping cleanup",
+				key, vmiMigration.Spec.VMIName)
+			return nil
+		}
+	}
 
 	vmi, err := c.config.KubevirtClient.VirtualMachineInstance(namespace).Get(context.TODO(), vmiMigration.Spec.VMIName, metav1.GetOptions{})
 	if err != nil {
+		if k8serrors.IsNotFound(err) && vmiMigration.Status.Phase == kubevirtv1.MigrationFailed && vmiMigration.Status.MigrationState == nil {
```
zhangzujian, line 132: P1: cleanup is only attempted when `MigrationState` is nil. If the VMI is deleted while this Failed VMIM still has a non-nil state, this returns NotFound and leaves any Pending-written OVN migration options behind. Please handle both state variants or explicitly prove the LSP was already removed.

zhangzujian (changes requested):
## Request changes

The Pending setup path and its focused regression tests are directionally correct, but the current head still has three P1 lifecycle hazards that can leave or overwrite OVN migration state:

1. The `hasUnfinishedVMIMigration` API list at `pkg/controller/kubevirt.go:118-128` is not atomic with the reset at `pkg/controller/kubevirt.go:257-264`. A newer migration can be created after the list and before the reset, allowing an older Failed migration to overwrite the newer migration's `requested-chassis`.
2. `cleanupFailedVMIMigrationWithoutVMI` at `pkg/controller/kubevirt.go:272-289` scans any matching `rarp/requested-chassis` LSP and resets it without migration ownership/provenance. The same race can therefore clobber a newer migration after the VMI has disappeared.
3. If the VMI is NotFound while the Failed VMIM still has a non-nil `MigrationState`, `pkg/controller/kubevirt.go:132-136` returns the error and skips cleanup, so options written during Pending can remain indefinitely.

Please make Failed cleanup ownership-safe (for example, serialize/check ownership immediately before reset or persist migration provenance), cover both VMI-deleted variants, and add an interleaving regression test for a newer Pending migration.

Non-blocking: `handleAddOrUpdateVMIMigration` is over 100 lines, contrary to `CODE_STYLE.md`; splitting Pending setup and terminal cleanup would also make the ownership boundary clearer.

PR #7418, opened 2026-09-11
fix(pinger,monitor): allow monitoring/metrics components to find correct ctrl sockets

### LINKED ISSUE #7417: [BUG] various monitoring components are broken after removal of hostPid
### Kube-OVN Version

v1.16.3

### Kubernetes Version

v1.35.4

Red Hat Enterprise Linux 10.1 (Coughlan)
6.12.0-124.40.1.el10_1.x86_64

### Description

Since #6230, kube-ovn-pinger and kube-ovn-monitor cannot reach the control sockets of the OVS/OVN daemons, so their health metrics are permanently stuck at "down" even on a fully healthy cluster.

#6230 removed the Go-side pid file resolution added by #5973 and let ovs-appctl/ovn-appctl resolve a bare daemon name themselves. That resolution cannot work from these two pods.

ovs-appctl resolves a non-absolute `--target` via `read_pidfile()` (utilities/ovs-appctl.c):

```
  pidfile_name = xasprintf("%s/%s.pid", ovs_rundir(), target);
  pid = read_pidfile(pidfile_name);
  socket_name = xasprintf("%s/%s.%ld.ctl", ovs_rundir(), target, (long int) pid);
```

and `read_pidfile__()` (lib/daemon-unix.c) does not use the file's contents:

```
  file = fopen(pidfile_, "r+");
  ...
  if (lck.l_pid != strtoul(line, NULL, 10)) {
      error = ESRCH;
      VLOG_WARN("%s: stale pidfile for pid %s being deleted by pid %ld", ...);
```

This fails for two reasons:
1. `fcntl(F_GETLK).l_pid` is translated through locks_translate_pid(), which returns 0 when the lock owner is in a namespace that is not an ancestor of the caller's. Both pods run with hostPID: false, while the daemons live in ovs-ovn (hostPID: true) and ovn-central. So l_pid is 0, never matches the pid in the file, and the lookup fails with ESRCH.

2. The pid file is 0644 root:root, but `read_pidfile__` opens it "r+". Both pods run as uid 65534 (kubeovn.runAsUser), so the open fails with EACCES. This blocks the lookup even where the PID namespace is shared.

The control sockets themselves are reachable so the issue is just in the pid resolution.

### Steps To Reproduce

On a healthy cluster:

```
kubectl -n kube-system exec ds/kube-ovn-pinger -- curl -sS http://localhost:8080/metrics | grep -E '^pinger_(ovs|ovn_controller)_(up|down)'
```
```
kubectl -n kube-system exec deploy/kube-ovn-monitor -- curl -sS http://localhost:10661/metrics | grep 'component="ovn-northd"'
```
The logs will contain errors with respect to these components trying to talk to other components.

### Current Behavior

The metrics pinger_ovs_up and pinger_ovn_controller_up are 0 and their _down counterparts 1 on every node for a healthy cluster. Also kube_ovn_ovn_status{component="ovn-northd"} is 0.

### Expected Behavior

pinger_ovs_up, pinger_ovn_controller_up, and kube_ovn_ovn_status should reflect cluster health

jskrill:
# Pull Request

## What type of this PR

- Bug fixes
- Tests

Since #6230, kube-ovn-pinger and kube-ovn-monitor cannot reach the control sockets of the OVS/OVN daemons, so their health metrics are permanently stuck at "down" even on a fully healthy cluster.

#6230 removed the Go-side pid file resolution added by #5973 and let ovs-appctl/ovn-appctl resolve a bare daemon name themselves. That resolution cannot work from these two pods.

ovs-appctl resolves a non-absolute `--target` via `read_pidfile()` (utilities/ovs-appctl.c):

```
  pidfile_name = xasprintf("%s/%s.pid", ovs_rundir(), target);
  pid = read_pidfile(pidfile_name);
  socket_name = xasprintf("%s/%s.%ld.ctl", ovs_rundir(), target, (long int) pid);
```

and `read_pidfile__()` (lib/daemon-unix.c) does not use the file's contents:

```
  file = fopen(pidfile_, "r+");
  ...
  if (lck.l_pid != strtoul(line, NULL, 10)) {
      error = ESRCH;
      VLOG_WARN("%s: stale pidfile for pid %s being deleted by pid %ld", ...);
```

This fails for two reasons:
1. `fcntl(F_GETLK).l_pid` is translated through locks_translate_pid(), which returns 0 when the lock owner is in a namespace that is not an ancestor of the caller's. Both pods run with hostPID: false, while the daemons live in ovs-ovn (hostPID: true) and ovn-central. So l_pid is 0, never matches the pid in the file, and the lookup fails with ESRCH.

2. The pid file is 0644 root:root, but `read_pidfile__` opens it "r+". Both pods run as uid 65534 (kubeovn.runAsUser), so the open fails with EACCES. This blocks the lookup even where the PID namespace is shared.

The control sockets themselves are reachable so the issue is just in the pid resolution.

This is fixed by reading the pidfile in Go instead of delegating to ovs-appctl/ovn-appctl.

Fixes #7417

### DIFF
--- a/test/e2e/kube-ovn/metrics/metrics.go
+++ b/test/e2e/kube-ovn/metrics/metrics.go
@@ -1,0 +1,187 @@
+package metrics
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"slices"
+	"strconv"
+	"strings"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/onsi/ginkgo/v2"
+
+	"github.com/kubeovn/kube-ovn/test/e2e/framework"
+)
+
+const (
+	// the exporters populate their gauges on a polling loop, so a scrape right after
+	// the pod has started may not expose the samples we are looking for yet
+	scrapeInterval = 2 * time.Second
+	scrapeTimeout  = time.Minute
+)
+
+// sampleValue returns the value of the first sample of the given metric whose line contains
+// all the given labels, e.g. `hostname="node1"`. Matching the labels individually keeps the
+// lookup working whatever their order is and whenever a new label is added to the metric.
+func sampleValue(metrics, name string, labels ...string) (float64, bool) {
+	for line := range strings.Lines(metrics) {
+		line = strings.TrimSpace(line)
+		if !strings.HasPrefix(line, name+"{") && !strings.HasPrefix(line, name+" ") {
+			continue
+		}
+		if slices.ContainsFunc(labels, func(label string) bool { return !strings.Contains(line, label) }) {
+			continue
+		}
+		// the value is the last field: a label value may contain spaces
+		index := strings.LastIndexByte(line, ' ')
+		if index < 0 {
+			continue
+		}
+		if value, err := strconv.ParseFloat(line[index+1:], 64); err == nil {
+			return value, true
+		}
+	}
+	return 0, false
+}
+
+// predicate is an assertion on the value of a sample.
+type predicate struct {
+	desc  string
+	match func(float64) bool
+}
+
+func equalTo(expected float64) predicate {
+	return predicate{
+		desc:  fmt.Sprintf("to be %v", expected),
+		match: func(value float64) bool { return value == expected },
+	}
+}
+
+func oneOf(expected ...float64) predicate {
+	return predicate{
+		desc:  fmt.Sprintf("to be one of %v", expected),
+		match: func(value float64) bool { return slices.Contains(expected, value) },
+	}
+}
+
+func nonZero() predicate {
+	return predicate{
+		desc:  "to be non-zero",
+		match: func(value float64) bool { return value != 0 },
+	}
+}
+
+// expectation is an assertion on the sample of a metric identified by its labels.
+type expectation struct {
+	name   string
+	labels []string
+	pred   predicate
+}
+
+func (e expectation) String() string {
+	return fmt.Sprintf("%s{%s}", e.name, strings.Join(e.labels, ","))
+}
+
+// waitForMetrics scrapes the metrics exposed by the given container from within the pod
+// itself until all the expectations are met.
+func waitForMetrics(f *framework.Framework, pod *corev1.Pod, container, port string, secure bool, expectations []expectation) {
+	ginkgo.GinkgoHelper()
+
+	scheme, curlArgs := "http", "-fsS"
+	if secure {
+		scheme, curlArgs = "https", "-fsSk"
+	}
+	url := fmt.Sprintf("%s://%s/metrics", scheme, net.JoinHostPort(pod.Status.PodIP, port))
+
+	framework.WaitUntil(scrapeInterval, scrapeTimeout, func(_ context.Context) (bool, error) {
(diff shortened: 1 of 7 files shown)


test/e2e/kube-ovn/metrics/metrics.go
```diff
@@ -0,0 +1,187 @@
...
+				{name: "pinger_ovn_controller_down", labels: []string{"nodeName=" + node}, pred: equalTo(0)},
+				// collected from ovs-vswitchd via dpctl/dump-dps
+				{name: "kube_ovn_dp_total", labels: []string{"hostname=" + node}, pred: nonZero()},
+			}
+			for _, component := range []string{"ovsdb-server", "ovs-vswitchd"} {
+				expectations = append(expectations, expectation{
+					name:   "kube_ovn_ovs_status",
+					labels: []string{fmt.Sprintf("component=%q", component), "hostname=" + node},
+					pred:   equalTo(1),
+				})
```
zhangzujian, line 148: These extra samples are not the #7417 health metrics, and they are the Conformance failure.

`kube_ovn_dp_total` / `kube_ovn_ovs_status` use the OVS exporter hostname (`external_ids:hostname`), not `pod.Spec.NodeName`. Trusted x86 ipv4 overlay (`670bdf0`, job 103353694294) already had `pinger_ovs_up` / `pinger_ovn_controller_up` matching, then timed out on `kube_ovn_dp_total{hostname="kube-ovn-control-plane"} has not been exposed yet`.

Please delete these two extra expectations (and the `ovsdb-server`/`ovs-vswitchd` loop), or scrape without assuming hostname == NodeName. The `nodeName=` pinger gauges above are the right coverage for this bug.

zhangzujian (changes requested):
## Review

The pidfile fix for #7417 is the right approach. After #6230, `ovs-appctl`/`ovn-appctl` resolve a bare target via `fopen("r+")` + `F_GETLK`, which cannot work from pinger/monitor (`hostPID: false`, uid 65534). Reading the pidfile in Go and targeting `{component}.{pid}.ctl` is correct, and trusted x86 CI already shows the issue metrics recovering:

- `kube-ovn-monitor should report ovn-northd as healthy` passed
- `pinger_ovs_up` / `pinger_ovn_controller_up` (and the `_down` counterparts) already matched before the extra assertions

DCO is now signed on `670bdf0`.

## Requested change

Please drop or rewrite the extra pinger Conformance assertions on `kube_ovn_dp_total` and `kube_ovn_ovs_status`. They are outside #7417, and they are what fail the suite.

On HEAD `670bdf0`, Kube-OVN Conformance ipv4 overlay still fails only that pinger case after waiting on `kube_ovn_dp_total{hostname="kube-ovn-control-plane"}` for a minute:

https://github.com/kubeovn/kube-ovn/actions/runs/34624617455/job/103353694294

Those gauges are labeled with `e.Client.System.Hostname` from OVS `external_ids:hostname`, not `pod.Spec.NodeName`. `waitForMetrics` also checks expectations in order, so a missing extra sample hides that the health metrics from the issue already succeeded.

Keeping the `nodeName=` pinger gauges plus monitor `kube_ovn_ovn_status{component="ovn-northd"}` is enough to lock the fix.

HA / MetalLB failures on this run look unrelated to the pid resolution change.

PR #7420, opened 2026-09-11
fix(controller): Protect VMs from incorrect chassis settings

### LINKED ISSUE #7419: [BUG] KubeVirt live migration can resolve to incorrect lsp options in various scenarios
### Kube-OVN Version

v1.16.3

### Kubernetes Version

v1.35.4

Red Hat Enterprise Linux 10.1 (Coughlan)
6.12.0-124.40.1.el10_1.x86_64

### Description

This issue is a bit of grab bag of places where KubeVirt VMs can get incorrect lsp options which were found while stress testing failure scenarios and with some static analysis from Claude. These issues are generally encountered when `EnableKeepVMIP` and `EnableLiveMigrationOptimize` are enabled.

The general theme is that the `virt-launcher` never gets its port bound:

```
Init:0/1
ovs interface <id>_h is not ready after 30s
```

Which boils down to `requested-chassis` / `activation-strategy` written by one code path and destroyed or forgotten by another. Here's a summary of the issues found courtesy of Claude:

---
## 1. A controller restart strips the options of a migration that is mid-setup

`handleAddOrUpdateVMIMigration` replays as an informer Add for every `VirtualMachineInstanceMigration`
object that still exists when kube-ovn starts, terminal ones included.

The UID filter reduces those to at most one per VM — the migration whose UID still matches
`vmi.Status.MigrationState.MigrationUID`. That survivor is exactly the dangerous one: it is the
immediate predecessor of whatever is running now, because KubeVirt only moves the VMI's
`MigrationState` to the new migration at handoff (`handleTargetPodHandoff`, once the target pod is
`Ready`). While the successor is still being set up, the VMI still points at the predecessor.

So on restart:

| | event | `vmi.Status.MigrationState` | port |
|---|---|---|---|
| t0 | migration A (X→Y) succeeds | UID=A, X→Y | `Y` |
| t1 | migration B (Y→Z) starts, kube-ovn writes its options | still UID=A | `Y,Z` + rarp |
| t2 | controller restarts, replays A and B | still UID=A | |
| t3 | A handled: UID matches, resets the port to its target `Y` | | `Y` |

A's reset writes `Y` — B's *source*. The port is pinned to the source with no activation strategy,
and B's target on Z never binds.

Replay order comes from `indexer.List()` (map iteration), so this reproduces intermittently.
In steady state A is never re-processed: its phase never changes again, and a resync delivers
`old == new`, which the update handler drops. **A controller restart is the trigger.**

## 2. Deleting an unrelated VM pod wipes the options of a running migration

`handleDeletePod` cleaned the migrate options for *every* deleted VM pod. Since `getNameByPod`
resolves all virt-launcher pods of a VM to the same key, deleting any of them — including a
long-defunct pod from an earlier migration — reached the shared LSPs and deleted both
`requested-chassis` and `activation-strategy`.

Nothing re-establishes them: the migration is past the phase that writes them.

Two ways this fires mid-migration:

- **A user or cleanup job removing completed pods** (`--field-selector status.phase==Succeeded`,
  cron hygiene). No timing constraint — it can land at any moment.
- **KubeVirt's own GC.** `garbageCollectFinalizedMigrations` keeps the 5 most recent finalized
  migrations per VM (`defaultFinalizedMigrationGarbageCollectionBuffer = 5`) and deletes the
  defunct pod of each older one. It runs on every reconcile of a final migration, and finalized
  migrations are re-enqueued by *any* VMI update — including the updates the running migration
  itself produces.

The GC path is narrow in practice because the trim is eager: each completion trims back to
exactly 5, so a subsequent migration usually sees `count - 5 == 0`. It bites when a backlog
builds up — virt-controller restarting across completions, or its GC erroring out. The manual
deletion path has no such protection.

## 3. A migration that fails before handoff never resets its options

The options are written before `vmi.Status.MigrationState` exists for the migration. During
`MigrationScheduling` the source node comes from the `vmi.Status.NodeName` fallback, because
handoff — which writes the migration's UID into `vmi.Status.MigrationState` — only happens at
`MigrationScheduled`, once the target pod is `Ready`.

A migration that dies in between has therefore written its options but never reaches the VMI
migration state, so the terminal handler finds no source/target pair and returns without resetting
anything. The window is: target pod scheduled to a node but not yet `Ready`, with the migration
already at `Scheduling` — reached by `defaultCatchAllPendingTimeoutSeconds = 900` or a cancellation.

The port keeps `requested-chassis="X,Y"` + `activation-strategy=rarp` until another live migration or until
the VM pod is deleted.

## 4. Migrate options are written once, with no retry

Setup ran only on `MigrationScheduling`. If the target pod was not yet in the informer cache at
that moment, the handler returned without setting anything; the following phase changes re-enqueued
the migration but matched no case, so the options were never written for that migration at all.

---

Apologies for bundling multiple small issues into one but the fixes tended to overlap with each other as they were all targeting the same general symptoms.

### Steps To Reproduce

Many of these are hard to reproduce due to the tight timing required to trigger the issues. To reproduce:

1. Restart virt-controller mid migration while VM target pod is still in Pending/Scheduling - live migration target pod will fail to start
2. Delete pod from old live migration pod while current live migration is Pending/Scheduling/Running - watch lsp options get cleared early
3. Cancel a migration while it's still in the Scheduling phase - `requested-chassis="X,Y"` + `activation-strategy=rarp` will stay on lsp

### Current Behavior

In various scenarios, VM pods fail to start due to incorrect lsp options with:

```
Init:0/1
ovs interface <id>_h is not ready after 30s
```

### Expected Behavior

VM pods should start with correct lsp options

jskrill:
# Pull Request

## What type of this PR

- Bug fixes
- Tests

Incorrect chassis settings can occur when kube-ovn-controller restarts during live migration and when live migrations fail before the target pod is Ready and the migration UID written to the VMI. Additionally, deletion of pods that are unrelated to an active live-migration would cause chassis options to be cleaned which is fixed.

More details at #7419

Right before creating this PR @jimyag opened #7415 which should be looked at first. It's a real issue and also revealed a bad assumption I had - that the target pod of a live migration would not exist until the Scheduling phase. I think it makes sense to merge that PR first and rebase this on top of it. This PR is standalone so it can be reviewed independently.

This PR also adds some much needed test coverage to kubevirt.go.

Fixes #7419

### DIFF
--- a/pkg/controller/kubevirt.go
+++ b/pkg/controller/kubevirt.go
@@ -6,8 +6,10 @@ import (
 	"strings"
 	"time"
 
+	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
@@ -16,6 +18,9 @@ import (
 	"github.com/kubeovn/kube-ovn/pkg/informer"
 	"github.com/kubeovn/kube-ovn/pkg/util"
 )
+
+// value kubevirt sets on kubevirtv1.AppLabel for the pods a VM runs in
+const virtLauncherAppLabel = "virt-launcher"
 
 func (c *Controller) enqueueAddVMIMigration(obj any) {
 	key := cache.MetaObjectToName(obj.(*kubevirtv1.VirtualMachineInstanceMigration)).String()
@@ -133,10 +138,17 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
 		} else {
 			klog.Infof("current vmiMigration %s status %s, vmi MigrationState is nil", key, vmiMigration.Status.Phase)
 		}
-		// If we're at an end state and the vmi migration state is stale or nil, we're probably looking at an old migration
-		// either way, we can't proceed since we don't have the source and target nodes for resetting the migrate options
-		if vmiMigration.Status.Phase == kubevirtv1.MigrationSucceeded || vmiMigration.Status.Phase == kubevirtv1.MigrationFailed {
-			klog.V(3).Infof("VirtualMachineInstanceMigration %s migration state is Succeeded/Failed but VMI migration state is stale or nil, skipping", key)
+		// A migration that fails before its target pod is ready never reaches the VMI
+		// migration state, so the nodes it used have to be recovered from the target pod.
+		if vmiMigration.Status.Phase == kubevirtv1.MigrationFailed {
+			if srcNodeName, targetNodeName, err = c.recoverVMIMigrationNodes(vmiMigration, vmi); err != nil {
+				return err
+			}
+		}
+		// A succeeded migration with a stale state was taken over by a newer one, which pinned
+		// the ports itself; a failed one still has options of its own to release below
+		if (srcNodeName == "" || targetNodeName == "") && vmiMigration.Status.Phase == kubevirtv1.MigrationSucceeded {
+			klog.V(3).Infof("VirtualMachineInstanceMigration %s succeeded but VMI migration state is stale or nil, skipping", key)
 			return nil
 		}
 	}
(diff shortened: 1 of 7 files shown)


pkg/controller/kubevirt.go
```diff
@@ -133,10 +138,17 @@ func (c *Controller) handleAddOrUpdateVMIMigration(key string) error {
 		} else {
 			klog.Infof("current vmiMigration %s status %s, vmi MigrationState is nil", key, vmiMigration.Status.Phase)
 		}
-		// If we're at an end state and the vmi migration state is stale or nil, we're probably looking at an old migration
-		// either way, we can't proceed since we don't have the source and target nodes for resetting the migrate options
-		if vmiMigration.Status.Phase == kubevirtv1.MigrationSucceeded || vmiMigration.Status.Phase == kubevirtv1.MigrationFailed {
-			klog.V(3).Infof("VirtualMachineInstanceMigration %s migration state is Succeeded/Failed but VMI migration state is stale or nil, skipping", key)
+		// A migration that fails before its target pod is ready never reaches the VMI
+		// migration state, so the nodes it used have to be recovered from the target pod.
+		if vmiMigration.Status.Phase == kubevirtv1.MigrationFailed {
```
zhangzujian, line 144: MigrationState can be nil before handoff, including the failed-before-handoff path. The guard at the start of this handler returns before reaching this new recovery branch, so a Failed migration can leave requested-chassis and activation-strategy pinned. Please handle terminal Failed cleanup before the nil-state guard (or explicitly allow Failed through it) and add a regression test with a nil VMIMigration status state.

zhangzujian (changes requested):
Request changes.

Standards: handleAddOrUpdateVMIMigration is about 149 lines, exceeding the repository CODE_STYLE.md limit of 100 lines for one function. Please split the phase setup, migration lookup, and terminal cleanup paths.

Validation performed locally: go test ./pkg/controller ./pkg/ovs, gofmt -d, and git diff --check passed.

pkg/controller/kubevirt.go
jskrill: `vmiMigration.Status.MigrationState` is set at two places. The first is https://github.com/kubevirt/kubevirt/blob/v1.9.0/pkg/controller/controller.go#L423-L438 (only fires during the Pending phase) and the second is https://github.com/kubevirt/kubevirt/blob/v1.9.0/pkg/virt-controller/watch/migration/migration.go#L553-L557 (fires during Final phases).

In both these cases, `vmiMigration.Status.MigrationState` only ever goes nil -> non-nil. It's instantiated in the first link and the second link sets it to `vmi.Status.MigrationState` only when `vmi.Status.MigrationState` is non nil (it's guarded by `IsMigrationSynchronized` which asserts non-nil).

Since we only set chassis-options in kube-ovn once `vmiMigration.Status.MigrationState` is non-nil, it follows that if `vmiMigration.Status.MigrationState` was never non-nil, we don't need to do any cleanup.

This new recovery path is specifically for when `vmiMigration.Status.MigrationState` was set and target pod was created but the target pod never became Ready causing the migration to fail (target pod fails to start, image pull error, node pressure or eviction, migration timeout, etc). This is reachable since `vmi.Status.MigrationState` (separate from `vmiMigration.Status.MigrationState`) is only set after handoff at https://github.com/kubevirt/kubevirt/blob/v1.9.0/pkg/virt-controller/watch/migration/migration.go#L1894 which requires the target pod to be Ready. At this point we may have already written chassis option at Scheduling (and after #7415, Pending). If we can't recover the source and target here, this falls through to the CleanLogicalSwitchPortMigrateOptions branch (https://github.com/kubeovn/kube-ovn/blob/c9333c1ef/pkg/controller/kubevirt.go#L229-L239), which clears the port options

jskrill:
@zhangzujian Thanks for the review! Split up handleAddOrUpdateVMIMigration into a few function (on the switch branches) and responded to your code comment.

zhangzujian (approved):
Re-reviewed at cb6607007ec604e755106e21877a5950ac3b850f.

The production handler has been split into focused helpers and now satisfies the repository's 100-line function limit. I also rechecked my earlier inline concern against the KubeVirt v1.9.0 controller lifecycle: updateStatus calls SetSourcePod while the VMIM is Pending, before target-pod creation can advance it to Scheduling, and SetSourcePod initializes VMIM status.migrationState. failMigration does not clear that state. Therefore the outer nil-state guard does not make the pre-handoff failure cleanup unreachable in the supported flow, and my previous concern is resolved.

Validation: go test ./pkg/controller ./pkg/ovs, gofmt -d on the changed Go files, and git diff --check all passed. The currently failing generic E2E jobs do not exercise KubeVirt paths; Build kube-ovn (including unit tests), CodeQL, and DCO passed.

Approved.

PR #7427, opened 2026-09-12
fix(dhcp): keep commas inside braced DHCP option values

gurinderu:
# Pull Request

## What type of this PR

- Bug fixes
- Tests

`parseDHCPOptions` splits the raw option string on every comma. Only `dns_server` can carry a list, via a `;` escape that is translated back to `,` for that key alone. Any other value containing a comma is truncated at the first comma and the remainder silently dropped.

This makes option 121 (`classless_static_route`) impossible to deliver: its OVN format always contains a comma, even for a single route, so

```
classless_static_route={10.0.0.0/24,10.0.0.1}
```

is stored as `classless_static_route={10.0.0.0/24`, the `10.0.0.1}` tail is discarded, `router` is then re-injected from the subnet gateway, and OVN can reject the malformed entry, leaving the port without DHCP at all. The same applies to every other list-valued option (`ntp_server`, `log_server`, `domain_search_list`, `ms_classless_static_route`, DHCPv6 `domain_search`).

Motivation: multi-NIC KubeVirt VMs with managedTap need a default route on exactly one interface. `router` is a mandatory option and is always re-injected, so it cannot be suppressed; option 121 is the standard way to do this (RFC 3442: a client that understands 121 must ignore option 3).

### Change

The split is now brace-aware: commas inside `{...}` are part of the value, so list options can be written in their native OVN form. Nothing else changes:

- the `dns_server` `;` escape is kept as is for compatibility;
- scalar values are passed through untouched, so `wpad=http://example.test/a;b` round-trips unchanged;
- unbalanced braces fall back to the plain comma split, so a malformed value cannot swallow the options following it.

Unit tests cover braced values in `parseDHCPOptions` and `formatDHCPOptions`, semicolons in scalar values, and the unbalanced-brace fallback.

N/A

### DIFF
--- a/pkg/ovs/util.go
+++ b/pkg/ovs/util.go
@@ -169,6 +169,35 @@ func formatDHCPOptions(options map[string]string) string {
 	return sb.String()
 }
 
+// splitDHCPOptions splits the raw option string on commas,
+// commas inside braces are part of the value and do not split,
+// e.g. router=10.0.0.1,classless_static_route={10.0.0.0/24,10.0.0.1}
+func splitDHCPOptions(raw string) []string {
+	var options []string
+	depth, start := 0, 0
+	for i, c := range raw {
+		switch c {
+		case '{':
+			depth++
+		case '}':
+			if depth > 0 {
+				depth--
+			}
+		case ',':
+			if depth == 0 {
+				options = append(options, raw[start:i])
+				start = i + 1
+			}
+		}
+	}
+	if depth != 0 {
+		// unbalanced braces, fall back to a plain split so that a malformed
+		// value does not swallow the options following it
+		return strings.Split(raw, ",")
+	}
+	return append(options, raw[start:])
+}
+
 // parseDHCPOptions parses dhcp options,
 // the raw option's format is: server_id=192.168.123.50,server_mac=00:00:00:08:0a:11
 func parseDHCPOptions(raw string) map[string]string {
@@ -181,8 +210,7 @@ func parseDHCPOptions(raw string) map[string]string {
 
 	// trim blank
 	raw = strings.ReplaceAll(raw, " ", "")
-	options := strings.SplitSeq(raw, ",")
-	for option := range options {
+	for _, option := range splitDHCPOptions(raw) {
 		kv := strings.Split(option, "=")
 		// TODO: ignore invalidate option, maybe need further validation
 		if len(kv) != 2 || len(kv[0]) == 0 || len(kv[1]) == 0 {
--- a/pkg/ovs/util_test.go
+++ b/pkg/ovs/util_test.go
@@ -70,6 +70,38 @@ func Test_parseDHCPOptions(t *testing.T) {
 		}
 		require.Equal(t, expected, result)
 	})
+
+	t.Run("braced values with commas", func(t *testing.T) {
+		t.Parallel()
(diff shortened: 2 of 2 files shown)


pkg/ovs/util.go
```diff
@@ -188,10 +185,7 @@ func parseDHCPOptions(raw string) map[string]string {
 		if len(kv) != 2 || len(kv[0]) == 0 || len(kv[1]) == 0 {
 			continue
 		}
-		if kv[0] == "dns_server" {
-			kv[1] = strings.ReplaceAll(kv[1], ";", ",")
-		}
-		dhcpOpt[kv[0]] = kv[1]
+		dhcpOpt[kv[0]] = strings.ReplaceAll(kv[1], ";", ",")
```
zhangzujian: `parseDHCPOptions` still calls `strings.SplitSeq(raw, ",")` above. Consequently the comma-form value shown in the PR description (`classless_static_route={10.0.0.0/24,10.0.0.1}`) is split before this replacement runs, and only `{10.0.0.0/24` survives. Please add support for bracketed/quoted values, or explicitly enforce the semicolon form and update the claimed fix.

pkg/ovs/util.go
```diff
@@ -161,10 +161,7 @@ func formatDHCPOptions(options map[string]string) string {
 		if sb.Len() > 0 {
 			sb.WriteString(",")
 		}
-		if k == "dns_server" {
-			v = strings.ReplaceAll(v, ",", ";")
-		}
-		fmt.Fprintf(&sb, "%s=%s", k, v)
+		fmt.Fprintf(&sb, "%s=%s", k, strings.ReplaceAll(v, ",", ";"))
```
zhangzujian: Formatting now replaces commas for every key, paired with the global semicolon decode at line 188. Scalar `str` options such as `wpad`, `domain_name`, or `path_prefix` can legally contain semicolons; for example, `wpad=https://example.test/a;b` now round-trips as a value containing a comma. Restrict escaping to list-valued keys or use an unambiguous encoding.

zhangzujian (changes requested):
Request changes

I found two behavioral issues:

1. The parser still splits raw input on every comma, so the comma-form `classless_static_route` shown in the PR description remains truncated. The new tests only cover semicolon encoding.
2. Decoding semicolons for every key changes scalar string DHCP options that can legally contain semicolons.

Local validation: `go test ./pkg/ovs` and `git diff --check` pass.

gurinderu:
Thanks for the review, both points were valid. Reworked: the split is now brace-aware instead of decoding `;` globally, so `classless_static_route={10.0.0.0/24,10.0.0.1}` works in its native form, the `dns_server` `;` escape stays as is, and scalar values such as `wpad` are untouched. Added tests for all three cases plus the unbalanced-brace fallback, and signed off the commit for DCO.

zhangzujian (approved):
Re-reviewed HEAD `325c0d0304b54034ed70f221fa57eb7256a2a36d`. The two requested changes are resolved: parsing is brace-aware for native OVN list values, `dns_server` semicolon compatibility is preserved, scalar semicolons are left unchanged, and regression tests cover braced values and unbalanced braces. `go test ./pkg/ovs` and `git diff --check` pass.

Approval is for the code change. The trusted x86 E2E check set still shows stale failures from run `34701956777`; the later dispatcher run `34722570566` did not execute replacement E2E jobs, so those gates still need separate follow-up.

PR #7432, opened 2026-09-12
fix(controller): keep ClusterIP load balancers off routers

zhangzujian:
## Summary

- Keep service-scoped ClusterIP and NodePort load balancers attached to enabled logical switches only.
- Preserve logical-router attachments for `LoadBalancer` Services and router load-balancer rules.
- Add regression coverage for ClusterIP attachment behavior.

## Root cause

GitHub Actions run [34709224263](https://github.com/kubeovn/kube-ovn/actions/runs/34709224263), job [103597197497](https://github.com/kubeovn/kube-ovn/actions/runs/34709224263/job/103597197497) failed the conformance test `[CNI:Kube-OVN] [group:node] should access overlay services using node ip`. The backend observed the OVN join IP (`100.64.0.2`) instead of the host-network node IP.

The service-scoped internal load balancer was attached to both logical switches and the VPC router. Router attachment sends host-network ClusterIP traffic through the OVN join path, changing the source address observed by the backend. This restores the legacy VPC load-balancer attachment behavior for internal Service traffic while retaining router attachment where it is required for external ingress.

## Validation

- `go vet ./pkg/controller`
- `git diff --check`


zhangzujian:
Closing this duplicate PR because commit 5b2df9488 fixes a regression introduced and detected by PR #7378. The commit has been pushed to #7378 directly.

PR #7442, opened 2026-09-13
Add an option to set a max MTU for default/custom subnets

SkalaNetworks:
# Pull Request

## What type of this PR

- Features

## Description

**Current situation:**

Custom/default subnets can be configured with a MTU. This is useful if we're dealing with MTUs above 1500.

Example with the physical network backing the Kube-OVN network set to 9000 (jumbo):

If the traffic from our subnet has the internet as its destination, the auto-detection of Kube-OVN will set the MTU of the pods to 8900. This can end up a problem on the internet, with various switches silently dropping traffic. Even if they don't, we still end up with fragmentation all over the place, which  is not desirable.

So to deal with that, we can simply set .mtu to 1500 on the subnets to force them to get back to a "normal" MTU that is accepted on the internet.

**With this patch:**

We can set a max MTU for the entire CNI. If the auto-detection of the MTU goes above it, the MTU is floored back to the max MTU.

On a MTU of 9000 for the physical network, the pod network is auto-detected at 8900 (-100 for the GENEVE headers).
The max MTU cap kicks in and we end up back at 1500.

I just added this patch to lessen the needs for manual intervention on the MTUs of subnets when running jumbos.
I have all my networks at 9000, to remove any issue with GENEVE and to allow my pod MTU to be 1500. But that forces me to create all my subnets with a fixed 1500 MTU or I start fragmenting on the internet, and sometimes get PMTU blackhole.


SkalaNetworks:
@zhangzujian small quality of life PR to deploy in environments with exotic MTUs without risking problems sending fragmentable traffic on the internet

PR #7445, opened 2026-09-14
ci: keep the published kube-ovn images fresh

zbb88888:
# Pull Request

## What type of this PR

- CI

N/A

**English.** This PR keeps the images that the charts and users pull in step with master. `Build x86 Image` published `kubeovn/kube-ovn:<VERSION>-x86` only after the whole x86 E2E matrix succeeded, and `publish.yaml` assembled the multi-arch tags only on a daily cron, so a failing or cancelled group left the amd64 tag unpublished for two weeks (`v1.17.0-x86` on 2026-08-31 vs `v1.17.0-arm` on 2026-09-14) and the multi-arch tag could still ship an amd64 image built from an old commit - which is what made `ct install` fail with `unknown flag: --enable-nftable-lb-svc`. The amd64 tags are now published as soon as the build succeeds, and a dedicated `Assemble multi-arch tags` job in each architecture workflow dispatches `publish.yaml` once its own tags are pushed (default branch only, whichever architecture finishes last triggers the assembly). The auxiliary `vpc-nat-gateway` amd64 tags moved to their own job, so a failure of that build no longer blocks the kube-ovn publication (the trivy scan is split into `scan-kube-ovn` / `scan-vpc-nat-gateway` accordingly). Build arm64 Image records its kube-ovn and vpc-nat-gateway pushes with separate markers, so a failing vpc image cannot hide an already published kube-ovn pair from the assembly either. The dispatch carries the tag it was built for - so a version bump cannot redirect the queued run - and is gated on a marker that the tags consumed by the publish were really pushed, so a failed build cannot trigger an assembly of leftover tags while an auxiliary push failure no longer skips it. The publish pulls each pair itself (a counterpart that is not published yet is skipped with a warning, other pull/inspect failures keep the run red), compares both legs' `org.opencontainers.image.revision` (an absent label counts as unknown) and only assembles a tag when both legs come from the same revision, pinning the verified digests in the manifest list so a later push cannot change what was verified; a mismatching pair is skipped with a warning, and every assembly runs in a concurrency group keyed by the assembled tag - a version bump cannot replace the pending assembly of the previous release, and the daily schedule re-dispatches the workflow with the tag from VERSION instead of publishing directly, so it is serialized with the other runs for that tag. The merge is the publish gate: the post-merge E2E matrix remains a visible quality signal but no longer delays or blocks the push. `hack/test_publish_workflow.py` extracts the publish shell block and exercises it with a stub `docker` under `bash -e` (same revision / stale counterpart / missing and unrendered revision labels / counterpart not published / registry error / blob-not-found pull error / inspect error) derives the exercised tag from `VERSION` (a release bump cannot silently stop covering a scenario) and runs with `make ut` and the x86 control validation next to `hack/test_e2e_selector.py` and `hack/test_e2e_control.py` (114 tests), which also cover the new push jobs.

**中文。** 这个 PR 让 chart 与用户拉到的镜像与 master 保持同步。此前 `Build x86 Image` 只有整条 x86 E2E 全过才推 `kubeovn/kube-ovn:<VERSION>-x86`，而 `publish.yaml` 只在每日 cron 组装多架构 tag；只要有 group 失败或被取消，amd64 tag 就长期不更新（`v1.17.0-x86` 停在 2026-08-31，而 `-arm` 已到 2026-09-14），多架构 tag 因此可能夹带旧 commit 的 amd64 镜像，`ct install` 报 `unknown flag: --enable-nftable-lb-svc`。现在 amd64 系列 tag 随构建成功立即发布，两个架构 workflow 各自有一个独立的 `Assemble multi-arch tags` job：在本架构 tag 推完后 dispatch `publish.yaml`（仅默认分支，谁最后完成谁触发）。辅助的 `vpc-nat-gateway` amd64 tag 已拆到独立 job，它的构建失败不再阻塞 kube-ovn 的发布（trivy 扫描相应拆成 `scan-kube-ovn` / `scan-vpc-nat-gateway`）。 arm64 侧则用两个独立标记分别记录 kube-ovn 与 vpc-nat-gateway 的推送，vpc 推送失败也不会让已完成的 kube-ovn 架构对被漏掉组装。dispatch 会带上本次构建的 tag（版本 bump 不会重定向排队中的运行），并以"被消费的 tag 确实推送成功"的标记门控：构建失败不会去组装残留 tag，辅助 tag 推送失败也不跳过组装。publish 自己按组拉取（对侧未发布则告警跳过，其它 pull/inspect 失败保持红灯），组装前比较两腿的 `org.opencontainers.image.revision`（缺失视为未知），只有一致时才发布对应 tag，并用被校验镜像的**不可变 digest** 组装 manifest list；不一致则告警跳过。发布按目标 tag 排队：版本切换不会顶掉上一版本的待执行组装；每日 schedule 不再直接发布，而是带上 VERSION 派发一次本 workflow，因此与同一 tag 的其它运行串行，不会回滚已发布的 manifest。merge 即发布门槛——merge 后的 E2E 仍是可见质量信号，但不再延迟或阻断推送。`hack/test_publish_workflow.py` 会抽取 publish 的 shell 块，用假 `docker` 在 `bash -e` 下覆盖 8 种场景，并随 `make ut` 与 x86 控制校验一起运行（连同 `hack/test_e2e_selector.py`、`hack/test_e2e_control.py` 共 114 项测试，后者也覆盖新增的 push jobs）。


zbb88888:
@zhangzujian Could you please help review? Thanks!

PR #7446, opened 2026-09-14
ci: publish multi-arch tags when an architecture image is pushed

zbb88888:
# Pull Request

## What type of this PR

- CI

N/A

## Changes

`publish.yaml` assembles the multi-arch tags (`kube-ovn`, `kube-ovn-debug`, `vpc-nat-gateway`) from the
architecture tags (`-x86`, `-arm`), but it only ran on a daily cron. A multi-arch tag is a snapshot of the two
digests, so re-pushing an architecture tag does not update it: `v1.17.0-arm` was pushed at 2026-09-14 03:26
while the arm64 member of `v1.17.0` still pointed at the digest assembled the day before.

- `on.workflow_run` for `Build x86 Image` / `Build arm64 Image` on `master` / `release-*`, so a new
  architecture image is assembled into the multi-arch tag right away; the daily schedule stays as a safety net.
  The publish job only runs for completed push runs (`workflow_run.event == push` and `conclusion == success`),
  so arm64 pull request builds and the x86 pull request E2E dispatches (which run on the master ref) never publish.
- Before assembling a tag, both legs' `org.opencontainers.image.revision` labels are compared; a mismatch is
  reported as a warning and that tag is skipped, because the other architecture's trigger publishes once both
  legs come from the same commit.
- `Build x86 Image` dispatches this workflow from its push job right after the image tags are pushed (default
  branch pushes only, `actions: write`). `workflow_run` only reports completion after the whole run, and the x86
  run continues with the E2E matrix for hours, so a publish triggered by the arm64 run can find the amd64 leg
  missing and skip; the dispatch fires as soon as the amd64 leg exists, and the arm64 run still covers the
  opposite order.

## Notes

- `workflow_run` executes this file from the default branch, so the change is effective after it is merged.
- `concurrency.cancel-in-progress: true` means the second trigger of a merge cancels the first; the last run
  re-publishes all three tags, so the final state converges (a tag set may briefly be partially updated).
- Release branches keep assembling their multi-arch tags through `hack/release.sh`: a `release-*` push triggers
  this workflow on the default branch, which reads the default branch `VERSION`.

## Verification

- workflow YAML parses; the `workflow_run` trigger, the job condition, the push-job permissions and the dispatch
  step were asserted from the parsed files;
- the publish shell block was extracted from the YAML and exercised with a stub `docker`:
  - both legs at the same revision -> all three multi-arch tags are created and pushed;
  - amd64 leg one revision behind -> that tag is skipped with a `::warning::` while the other two are published;
- `kubeovn/kube-ovn`, `kubeovn/kube-ovn-debug` and `kubeovn/vpc-nat-gateway` published images carry the
  `org.opencontainers.image.revision` label (checked against the registry), so no group is skipped permanently;
- `python3 -m unittest hack/test_e2e_selector.py hack/test_e2e_control.py` passes once the contract test update
  from #7445 is applied (this branch keeps the master version of that test).


zbb88888:
Consolidated into #7445: this change touches the same push job in `.github/workflows/build-x86-image.yaml` as the amd64 publish alignment there, so both are reviewed and merged together.

PR #7451, opened 2026-09-15
feat(vpc-nat-gw): serve Services through nft share DNAT

zbb88888:
### 目标

为 vpc-nat-gw 增加基于 nftables share DNAT 的 Service 负载均衡能力，使同一个网关可以承载：

- LoadBalancer Service：同时服务 EIP ingress 地址和 ClusterIP。
- ClusterIP Service：不依赖 EIP，直接通过指定的 vpc-nat-gw 服务内部 VIP。

三种 Service LB 实现是全局互斥模式：

- `--enable-lb`：OVN Load Balancer。
- `--enable-lb-svc`：每个 LoadBalancer Service 一个 iptables 转发 Pod。
- `--enable-gw-nftable-lb-svc`：vpc-nat-gw 内的 nft share DNAT。

最多只能启用一个模式；全部关闭表示禁用 Service LB。只有被选中的模式会启动 worker 并处理 Service 事件。由于 `--enable-lb` 默认开启，使用 Pod 或 gateway 模式时必须显式关闭它。

### API 与控制面

- 为 `IptablesDnatRuleSpec` 增加可选字段：
  - `clusterIP`：规则服务的内部 VIP，仅支持 IPv4 和 `type=share`。
  - `vpcNatGwDp`：无 EIP 时用于指定网关；同时存在 EIP 时必须与 EIP 所属网关一致。
- `clusterIP` 规则不可原地修改，变更通过删除并重建收敛。
- Service 使用 `ovn.kubernetes.io/vpc-nat-gw` 指定网关；LoadBalancer Service 还需通过 `eip` 注解指定 ingress 地址。
- Controller 根据 Ready EndpointSlice 后端生成一个或多个 share DNAT 规则，并在 Service、EndpointSlice、Pod NIC、EIP、VpcNatGateway 和 DNAT Ready/Delete 事件上重新收敛。
- 生成规则使用 Service owner labels。Controller 重启后会扫描这些规则，恢复仍存在的 Service，并清理已删除 Service 的遗留规则。
- 同一 nft identity 只允许一个确定 owner 写数据面；冲突由 webhook 和 controller 双层约束，避免不同 Service 或手工规则混合 backend。

### 数据面与生命周期

- 每个 share DNAT identity 会编程到所有运行中的 gateway 实例。
- ClusterIP 以 `/32` 持有在 gateway 的 `lo` 上，并用专用标签限定同步范围，不影响其他 loopback 地址。
- VPC 内访问 EIP 或 ClusterIP 时，通过优先级 `29210` 的 OVN policy route 转发到所有已初始化 gateway 实例。
- 每个 VIP/端口 identity 配置 hairpin SNAT，源地址使用当前 gateway 实例的 VPC 地址，保证 HA 场景下回包返回持有 conntrack 的实例。
- backend 增删会基于当前全部 sibling DNAT 规则重建 nft map；VIP 地址与 policy route 按 gateway 的完整期望集合收敛。
- Gateway Pod 的 StatefulSet/Deployment 实例替换或容器重启会触发初始化和 DNAT redo，恢复 nft identity、VIP 地址、hairpin 和 route。
- 删除顺序为：清理数据面和共享 VIP 状态，更新引用状态，最后移除 DNAT finalizer。

### 功能边界

- 当前仅支持 IPv4 VIP/backend 和 TCP/UDP。
- gateway 模式使用 Cluster traffic policy，不实现 Local traffic policy、拓扑感知路由或 terminating endpoint fallback。
- ClientIP session affinity 使用 nftables timeout sets，默认超时 10800 秒。
- `--enable-gw-nftable-lb-svc` 默认关闭；kind 的 vpc-nat-gw 测试入口会显式关闭 OVN LB 并开启 gateway 模式。

### 测试

- Controller UT：模式互斥、事件队列隔离、规则生成、backend 聚合、identity owner、Ready/redo、删除/finalizer、VIP 集合、HA next hops、并发串行化和 informer 延迟收敛。
- Webhook UT：新字段校验、IPv4/share 限制、网关一致性、不可变性、owner 与 session affinity 冲突。
- Shell UT：`vip-addr-sync`、`vip-hairpin-add/del` 的幂等、集合替换、输入校验和地址作用域。
- E2E：LoadBalancer、ClusterIP、HA gateway、backend 增删、gateway 实例替换、ClientIP affinity、owner 冲突、注解移除和数据面清理。

验证命令：

```text
go test ./pkg/controller -count=1
go test ./pkg/webhook -run 'IptablesDnat' -count=1
bash dist/images/vpcnatgateway/nat_gateway_vip_test.sh
go test ./test/e2e/iptables-vpc-nat-gw -run '^$'
make lint
```

### Goal

Add nftables share-DNAT Service load balancing to vpc-nat-gw so one gateway can serve:

- LoadBalancer Services through both their EIP ingress address and ClusterIP.
- ClusterIP Services through an explicitly selected vpc-nat-gw without requiring an EIP.

The three Service LB implementations are globally mutually exclusive modes:

- `--enable-lb`: OVN Load Balancer.
- `--enable-lb-svc`: one iptables forwarding Pod per LoadBalancer Service.
- `--enable-gw-nftable-lb-svc`: nft share DNAT inside vpc-nat-gw.

At most one mode may be enabled; disabling all three disables Service load balancing. Only the selected mode starts workers and handles Service events. Since `--enable-lb` defaults to true, Pod or gateway mode must explicitly disable it.

### API and control plane

- Add optional fields to `IptablesDnatRuleSpec`:
  - `clusterIP`: the internal VIP served by the rule; IPv4 and `type=share` only.
  - `vpcNatGwDp`: selects the gateway when no EIP is present and must match the EIP gateway when both are set.
- Rules carrying a ClusterIP are immutable; changes converge through delete and recreate.
- A Service selects its gateway with `ovn.kubernetes.io/vpc-nat-gw`; a LoadBalancer Service also references its ingress address through the `eip` annotation.
- The controller creates share DNAT rules from Ready EndpointSlice backends and reconciles on Service, EndpointSlice, Pod NIC, EIP, VpcNatGateway, and DNAT Ready/Delete events.
- Generated rules carry Service owner labels. On restart, the controller scans these rules to restore existing Services and clean up rules whose Service no longer exists.
- Each nft identity has one deterministic writer. Admission and controller-side checks prevent backend mixing across Services and hand-managed rules.

### Data plane and lifecycle

- Every share DNAT identity is programmed on every running gateway instance.
- ClusterIPs are held as `/32` addresses on the gateway loopback interface under a dedicated label, so reconciliation does not touch unrelated loopback addresses.
- OVN policy routes at priority `29210` send VPC traffic for EIP and ClusterIP VIPs to all initialized gateway instances.
- Per-VIP/port hairpin SNAT uses the current gateway instance's VPC address, ensuring replies return to the replica that owns the conntrack entry in HA deployments.
- Backend changes rebuild the nft map from all current sibling DNAT rules. VIP addresses and policy routes reconcile from the complete desired set for each gateway.
- StatefulSet/Deployment Pod replacement and container restart trigger gateway initialization and DNAT redo, restoring nft identities, VIP addresses, hairpin rules, and routes.
- Deletion cleans the data plane and shared VIP state, updates references, and removes the DNAT finalizer last.

### Scope

- IPv4 VIPs/backends and TCP/UDP only.
- Gateway mode follows Cluster traffic policy; Local traffic policy, topology-aware routing, and terminating-endpoint fallback are not implemented.
- ClientIP session affinity uses nftables timeout sets with a default timeout of 10800 seconds.
- `--enable-gw-nftable-lb-svc` defaults to false. The kind vpc-nat-gw test target explicitly disables OVN LB and enables gateway mode.

### Tests

- Controller unit tests cover mode exclusivity, queue isolation, rule generation, backend aggregation, identity ownership, Ready/redo handling, deletion/finalizers, VIP set reconciliation, HA next hops, serialization, and informer-lag convergence.
- Webhook unit tests cover the new fields, IPv4/share constraints, gateway consistency, immutability, ownership conflicts, and session-affinity conflicts.
- Shell tests cover idempotency, set replacement, input validation, and address scoping for `vip-addr-sync` and `vip-hairpin-add/del`.
- E2E coverage includes LoadBalancer and ClusterIP Services, HA gateways, backend changes, gateway instance replacement, ClientIP affinity, ownership conflicts, annotation removal, and data-plane cleanup.

Validation:

```text
go test ./pkg/controller -count=1
go test ./pkg/webhook -run 'IptablesDnat' -count=1
bash dist/images/vpcnatgateway/nat_gateway_vip_test.sh
go test ./test/e2e/iptables-vpc-nat-gw -run '^$'
make lint
```


zbb88888:
@SkalaNetworks Hi, Could you please help review?

pkg/controller/controller.go
```diff
@@ -1541,18 +1554,23 @@ func (c *Controller) startWorkers(ctx context.Context) {
...
+	if c.config.EnableGwNftableLbSvc {
+		// Gateway mode consumes Service and EndpointSlice informer events through its own queue.
+		if err := c.enqueueNftableLbSvcOwnersFromRules(); err != nil {
+			util.LogFatalAndExit(err, "failed to enqueue nftable lb service owners from generated dnat rules")
+		}
+		go wait.Until(runWorker("add/update nftable lb service", c.addOrUpdateNftableLbSvcQueue, c.handleAddOrUpdateNftableLbService), time.Second, ctx.Done())
+	}
+
+	if c.config.EnableOvnLB || c.config.EnablePodLbSvc {
+		// OVN and Pod modes share Service queues, but the global mode check makes their handlers
```
zbb88888: EnableGwNftableLbSvc 也会消费 Kubernetes Service 事件，但不使用 addServiceQueue、updateServiceQueue、deleteServiceQueue 这组三个 OVN/Pod 共用队列。

 当前分流是：

 ```go
   switch {
   case c.config.EnableOvnLB:
       c.enqueueEndpointSliceService(...)
   case c.config.EnablePodLbSvc:
       c.addServiceQueue.Add(key)
   case c.config.EnableGwNftableLbSvc:
       c.enqueueNftableLbService(key)
   }
 ```

 gateway 模式使用自己的队列：

 ```go
   addOrUpdateNftableLbSvcQueue
 ```

 原因是三种模式处理的资源和收敛逻辑不同：

  ```
┌────────────────┬───────────────────────────────────────────────────────────────────────┬────────────────────────────────────────────────────────┐
 │ 模式           │ Service 事件入口                                                      │ 处理内容                                               │
 ├────────────────┼───────────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────┤
 │ OVN LB         │ addOrUpdateEndpointSliceQueue、updateServiceQueue、deleteServiceQueue │ OVN LoadBalancer、EndpointSlice、VPC LB                │
 ├────────────────┼───────────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────┤
 │ Pod LB         │ addServiceQueue、updateServiceQueue、deleteServiceQueue               │ lb-svc-* Deployment、Pod 内 iptables                   │
 ├────────────────┼───────────────────────────────────────────────────────────────────────┼────────────────────────────────────────────────────────┤
 │ gateway nft LB │ addOrUpdateNftableLbSvcQueue                                          │ IptablesDnatRule、nft map、VIP route、lo 地址、hairpin │
 └────────────────┴───────────────────────────────────────────────────────────────────────┴────────────────────────────────────────────────────────┘

 ```

 gateway 模式不需要 OVN/Pod 的 Service handler，因为：

 - 不创建或更新 OVN LoadBalancer；
 - 不创建 lb-svc-* Deployment；
 - 不需要 updateServiceQueue 中的 OVN VIP 更新；
 - 需要根据 EndpointSlice 生成 IptablesDnatRule，所以 EndpointSlice 事件也会进入 gateway 专用队列；
 - Service 删除、注解变化、EIP/Gateway 变化会进入 gateway 专用队列，统一调用 handleAddOrUpdateNftableLbService 做创建、更新或清理。

 因此这段注释：

 ```go
   // OVN and Pod modes share Service queues, but the global mode check makes their handlers
   // mutually exclusive at runtime.
 ```

 表达的是 OVN 和 Pod 两个 handler 复用 Service queue，但全局互斥保证不会同时处理。它没有说 gateway 不消费 Service 事件；不过容易让人误解，建议改成：

 ```go
   // OVN and Pod modes share the generic Service queues. Gateway mode uses its dedicated
   // nftable LB Service queue, so only the selected mode processes Service events.
 ```

 更准确。

SkalaNetworks:
> @SkalaNetworks Hi, Could you please help review?

Hi, no problem, I'll start on that today

pkg/webhook/vpc_nat_gateway.go
```diff
@@ -307,6 +307,15 @@ func (v *ValidatingHook) iptablesDnatUpdateHook(ctx context.Context, req admissi
 	}
 
 	if dnatNew.Spec != dnatOld.Spec {
```
zhangzujian, line 309: Owner labels participate in the authorization decision below, but this guard skips all validation when only metadata changes. A caller can therefore patch the reserved namespace/name labels without re-running the ownership conflict check. Validate ownership metadata changes as well, and reject reserved owner-label writes unless they come from the controller identity.

pkg/webhook/vpc_nat_gateway.go
```diff
@@ -809,22 +874,22 @@ func (v *ValidatingHook) ValidateIptablesDnat(ctx context.Context, dnat *ovnv1.I
 			if existingType != ovnv1.DnatRuleTypeShare {
 				continue
 			}
-			if existing.Spec.EIP != dnat.Spec.EIP ||
+			if !identitiesOverlap(&existing.Spec, &dnat.Spec) ||
 				existing.Spec.ExternalPort != dnat.Spec.ExternalPort ||
 				!strings.EqualFold(existing.Spec.Protocol, dnat.Spec.Protocol) {
 				continue
 			}
 			if nftableLbDnatOwnerKey(existing) != newOwner {
```
zhangzujian, line 882: These owner keys come entirely from mutable metadata labels. Matching keys therefore do not prove common ownership: a caller with DNAT rule write permission can copy another Service namespace/name labels and pass this check, after which its backend is eligible for the same nft map. Authenticate the owner tuple, include the Service UID, and reject reserved owner-label writes from non-controller identities.

zhangzujian (changes requested):
The Service ownership metadata introduced here crosses an authorization boundary but is not authenticated. The two owner labels are mutable metadata on a cluster-scoped IptablesDnatRule. A principal with permission to create, update, or patch these rules can copy another Service namespace/name pair; admission then treats the forged rule as same-owner, and shared-backend aggregation can program its InternalIP:InternalPort into the same nft map until asynchronous reconciliation removes it. Owner-label-only updates also bypass this validation because the update hook runs it only when Spec changes.

The restart and cleanup paths trust the same labels. Startup enqueues every decoded owner key, while cleanup treats any matching rule as evidence that this controller owns the current Service status. A forged or stale rule can therefore clear status.loadBalancer.ingress for a Service that this controller never managed. Namespace/name alone also conflates different Service incarnations after deletion and recreation.

Please treat these labels as indexes, not proof of ownership. Persist the Service UID in server-validated controller-owned state, verify the live UID and desired rule shape before backend aggregation, deletion, or status mutation, and reject reserved owner-label writes from non-controller identities. Please add regression coverage for forged owner labels, owner-label-only updates, and same-name/different-UID restart cleanup.

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
```diff
@@ -680,7 +699,8 @@ spec:
                 description: Internal port number to forward traffic to
                 type: string
               protocol:
-                description: Protocol type (TCP or UDP)
+                description: Protocol type (TCP or UDP). See ValidateProtocol for
```
SkalaNetworks, line 702: Unsure it is necessary to include implementation details/limitations in the comments of the CRD.

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
```diff
@@ -721,6 +741,12 @@ spec:
                 - exclusive
                 - share
                 type: string
+              vpcNatGwDp:
+                description: |-
```
SkalaNetworks, line 745: "when there is no EIP to derive the gateway from, and must name that gateway when both are"

I'm not sure I get what "both" is refering to? I'm under the impression it only needs to be set when clusterIP is set.

What does "Dp" stand for in vpcNatGwDp?

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
```diff
@@ -667,8 +667,27 @@ spec:
             type: object
           spec:
             properties:
+              clusterIP:
+                description: |-
```
SkalaNetworks, line 671: This description seems very programmer oriented and not user oriented. 

I have a hard time grasping what the feature does from this description which is more of a design document than an explanation of what clusterIP does.

charts/kube-ovn/templates/kube-ovn-crd.yaml
SkalaNetworks, line 1: Same comments as the one I put on  kube-ovn-v2

dist/images/install.sh
SkalaNetworks, line 1: Same comment as the one I put on kube-ovn-v2

test/e2e/iptables-vpc-nat-gw/e2e_test.go
```diff
@@ -376,6 +375,149 @@ func nftDnatAffinityRuleExists(natGwPodName, eip, externalPort, protocol string,
...
+	stdout, _, err := framework.KubectlExec(framework.KubeOvnNamespace, natGwPodName, "ip -4 addr show dev lo label lo:ko-vip")
+	if err != nil {
+		return false
+	}
+	return strings.Contains(string(stdout), vip+"/32")
+}
+
+// vipHairpinRuleExists checks the per-identity hairpin SNAT rule of a share DNAT VIP: traffic that
+// entered the gateway from the VPC and was DNAT'd back into the VPC is SNAT'd to the gateway's own
+// VPC address, so the backend's reply returns to the instance holding the conntrack (and not to a
```
SkalaNetworks: In HA mode, any of the instances should have the exact same network configuration (including addresses/nftables rules/iptables rules), so I'm unsure this comment actually represents the reality of things.

We want the traffic to return to the NAT gateway to handle hairpin scenario, but it doesn't have anything to do with multi instance. The destination might choose to forward the traffic back to another instance based. There's nothing preventing that but the VPC configuration (mostly the BFD/ECMP config)

test/e2e/iptables-vpc-nat-gw/e2e_test.go
```diff
@@ -602,7 +744,28 @@ var _ = framework.OrderedDescribe("[group:iptables-vpc-nat-gw]", func() {
...
+				return len(gws.Items)
+			}, 2*time.Minute, time.Second).Should(gomega.Equal(0), "all vpc nat gateways should be deleted before waiting for the EIPs they held")
+
 			ginkgo.By("Waiting for all EIPs using subnet " + networkAttachDefName + " to be deleted")
+			// An EIP's finalizer is only released once the data plane cleanup ran, and deleteEipInPod
+			// keeps retrying while the gateway CRD still exists but its Pod is gone (natGwDeleted
+			// returns false, getNatGwPods fails on "no active pod now"). That is why the wait above
+			// comes first and the two minutes below are enough: this used to fail a run whose own
+			// assertions all passed, observed in a dual-stack environment and reproduced with an
+			// upstream/master controller, when a spec had just deleted its gateway.
```
SkalaNetworks, line 768: I was under the impression we now had fixed this problem.

The cleanup logic shouldn't hang for EIPs/SNAT/DNATs if the gateway doesn't exist anymore. There's no reason to even try to cleanup the rules if the gateway is offline.

pkg/util/const.go
```diff
@@ -36,7 +36,12 @@ const (
 	VMAnnotation                 = "ovn.kubernetes.io/virtualmachine"
 	ActivationStrategyAnnotation = "ovn.kubernetes.io/activation_strategy"
 
-	VpcNatGatewayAnnotation     = "ovn.kubernetes.io/vpc_nat_gw"
+	VpcNatGatewayAnnotation = "ovn.kubernetes.io/vpc_nat_gw"
+	// VpcNatGatewaySvcAnnotation names the vpc-nat-gw that serves a Service handled by the
+	// nftable LB service feature. A Service selects its gateway with it, for a LoadBalancer
+	// Service (which also carries the eip annotation its ingress IP comes from) and for a
+	// ClusterIP Service alike; without it the Service is not handled.
+	VpcNatGatewaySvcAnnotation  = "ovn.kubernetes.io/vpc-nat-gw"
```
SkalaNetworks, line 44: This is the same annotation as VpcNatGatewayAnnotation but with the underscores replaced by dashes, why can't we use the pre-existing VpcNatGatewayAnnotation?

It make sense to re-use that annotation in my opinion. It just has a different meaning depending on where its placed:
- on Services: the gateway backing that service
- on  Pods/STS/Deployment: the gateway represented by that Pod/STS/Deployment

test/e2e/iptables-vpc-nat-gw/e2e_test.go
```diff
@@ -1703,6 +1868,69 @@ var _ = framework.OrderedDescribe("[group:iptables-vpc-nat-gw]", func() {
 		}, 60*time.Second, 2*time.Second).Should(gomega.BeTrue(),
 			"nft DNAT map rule should list both service backends")
 
+		ginkgo.By("Verifying the ClusterIP is programmed with the same backends as the EIP")
```
SkalaNetworks, line 1871: I understand the test but not the name of the test

"with the same backends as the Service" makes more sense to me?

SkalaNetworks: ClusterIP should also probably be called VIP in that context, considering we're checking for something INSIDE the NAT gateway.

I'd use "ClusterIP" for services of type ClusterIP (which we're not dealing with here, those are LoadBalancers)

SkalaNetworks (commented):
A few remarks I have on that feature:
- What does the "share" in share DNAT mean? I understand we're trying to get LBs/Cluster IPs to work on VPC NAT gateways, but why is it called "share"?
- It seems like this system is mutually exclusive with normal OVN LBs and SwitchLBRules. I think we could benefit from having both work together? We need annotations anyway on  the Services right now for them to be picked by the "new" system.

I see a use case for that feature:
- You're in a custom VPC/Subnets
- You're creating LBs targetting backends within that VPC/Subnet
- You want to automatically expose the backend to the Internet using an external IP (LBs work great for that)

In that scenario, SwitchLBRules stay relevant for internal load balancers, especially since they support real IPs instead of labels.

But those new LBs are very practical for anything else.

Something I could find missing (but it seems harder to do and out of the scope of this PR):
- auto generating an EIP for loadbalancers, instead of having to specify one
- automatically attaching it to the NAT gateway specified in the annotation

Basically, make the EIP annotation on Services something optional, and auto-create one.

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
zbb88888: should remove

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
zbb88888: Dp means deployment.
`I'm under the impression it only needs to be set when clusterIP is set.` you are right.

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
zbb88888: will fix

charts/kube-ovn-v2/crds/kube-ovn-crd.yaml
SkalaNetworks: Dp seems redundant, considering we're targeting a VPC NAT GW object and not an actual deployment

PR #7480, opened 2026-09-19
fix(controller): delete node router policies by owner on node deletion

chomatdam:
# Pull Request

## What type of this PR

- Bug fixes

## What we observed

On EKS clusters running kube-ovn v1.16.2 with Karpenter-managed nodes, we found priority 30000 reroute policies on `ovn-cluster` (`ip4.dst == <nodeIP>` reroute `<joinIP>`) whose node no longer existed.

These are harmless until the VPC reuses that IP. When it was handed to an EKS control plane ENI, the address became one of the `kubernetes` service endpoints, and every overlay pod's traffic to it was rerouted to a join IP that no longer existed.

The policies only disappeared after a kube-ovn-controller restart, when the startup gc runs.

## Root cause

The node IP CR is owned by the Node, so deleting the node also deletes the IP, and the two handlers race. Controller log for one node:

| Offset | Event |
|---|---|
| +0 ms | node deletion starts |
| +32 ms | IP handler releases the node's join address |
| +55 ms | node cleanup deletes policies with nexthop `"<nil>"` |
| +72 ms | node cleanup finishes successfully, policy still there |

1. `deletePolicyRouteForNode` deletes the policy by nexthop only, using the join address from IPAM.
2. The IP handler releases the node address under the pod style key `<namespace>/<name>`, but node addresses are allocated under the port name `node-<name>`. The address is freed and the node's entry stays behind with a nil address.
3. Node cleanup reads that entry, gets the nexthop `"<nil>"`, deletes nothing and returns success.

We confirmed this signature in the logs for 4 of 17 orphaned policies; the logs of the others had rotated.

## What this PR does

- `deletePolicyRouteForNode` deletes the policies owned by the node, by their `vendor` and `node` external IDs, and no longer deletes by nexthop. This is the fix: it does not read IPAM at all. Every priority 30000 node policy carries these IDs and the startup migration relabels older rows, so the same rows are covered. Deleting by nexthop had no ownership check: a node reusing the join address has policies with the same nexthop and could lose them. `DeleteLogicalRouterPolicyByNexthop` had no other caller and is removed.
- Each listed row is deleted with the new `DeleteLogicalRouterPolicyIfUnchanged`: by UUID, behind an ovsdb `wait` on its external IDs in the same transaction. A node reusing the address can take the row over between the list and the delete, and must keep it.
- `AddLogicalRouterPolicy` relabels an existing row in place when only the external IDs differ. If the row was deleted after it was read, the update touched no row and the call still reported success. It now creates the policy.
- The IP handler releases a node address under its IPAM key, so the stale entry is no longer left behind. On its own this did not fix the leak (the lookup was then empty and delete-by-nexthop still a no-op), it removes the `"<nil>"` state.
- The centralized gateway cleanup moves out of `deletePolicyRouteForNode` into `deletePolicyRouteForCentralizedSubnetOnNode`, to stay under the 100 line limit. No behaviour change.

An earlier revision of this PR also ran the node gc periodically. I removed it: with the cause fixed it is not needed, existing orphans are collected by the startup gc when the controller restarts on upgrade, and running it next to node reconciliation added races of its own.

## Tests

- `TestHandleDeleteNodeRemovesPolicyWhenIPIsReleasedFirst` runs the real node and IP handlers with the real IPAM against an in-memory northbound database, in both orders. It fails on master in the production order.
- `TestDeletePolicyRouteForNodeKeepsPolicyOfNodeReusingJoinAddress` keeps the deleted node's address in IPAM while another node already has a policy with that join IP as nexthop. The other node's policy must survive.
- `TestHandleUpdateIPReleasesNodeAddressWithNodeKey` covers the IPAM key.
- `Test_DeleteLogicalRouterPolicyIfUnchanged` covers a row relabelled to another node after it was listed, and a row already gone.
- `Test_AddLogicalRouterPolicyRecreatesVanishedPolicy` deletes the row between the read and the owner update.

## How we mitigated it meanwhile

A script run against each cluster. It lists the priority 30000 `ip4.dst` policies first, then the nodes, so a node created mid-run can never be treated as an orphan, and deletes the policies whose IP belongs to no node.

```bash
nbctl() { kubectl -n kube-system exec "$OVN_POD" -c ovn-central -- ovn-nbctl "$@"; }

POLICY_IPS=$(nbctl lr-policy-list ovn-cluster | awk '$1=="30000" && $2=="ip4.dst" {print $4}' | sort -u)
NODE_IPS=$(kubectl get nodes -o jsonpath='{range .items[*]}{.status.addresses[?(@.type=="InternalIP")].address}{"\n"}{end}' | sort -u)

for ip in $(comm -23 <(echo "$POLICY_IPS") <(echo "$NODE_IPS")); do
  nbctl lr-policy-del ovn-cluster 30000 "ip4.dst == $ip"
done
```

## Not in this PR

The startup `gcNode` reads `ExternalIDs["node"]` as the owner for gateway priority policies too. Centralized ECMP policies use node names as external ID keys, so a gateway node named `node` yields `"node": "<ip>"` and its valid policy is collected at startup. That is on master today and deserves its own fix.

None filed.


pkg/controller/gc.go
```diff
@@ -259,7 +290,11 @@ func (c *Controller) gcNode() error {
 		if _, ok := policy.ExternalIDs["node"]; !ok {
 			continue
 		}
-		if nodeNames.Has(policy.ExternalIDs["node"]) {
+		exists, err := c.nodeExists(policy.ExternalIDs["node"])
+		if err != nil {
+			return err
+		}
+		if exists {
```
zhangzujian: The stale-policy decision and the deletion are not bound to the same row. After `nodeExists` returns false, a new node can reuse the same destination IP and `addPolicyRouteToVpc` can replace or relabel the policy with the same priority and match. `DeleteLogicalRouterPolicy` then re-lists by `(priority, match)` and removes the new live policy, preserving the blackhole race this PR is intended to close. Please delete the observed candidate with an ownership/version guard rather than a fresh broad lookup, and add a test that replaces the policy after the lister check but before deletion.

pkg/controller/gc.go
```diff
@@ -259,7 +290,11 @@ func (c *Controller) gcNode() error {
 		if _, ok := policy.ExternalIDs["node"]; !ok {
```
zhangzujian: `ExternalIDs["node"]` is not unambiguously a node-owner field for every policy in this combined list. Centralized ECMP policies store each node name as an external-ID key, so a valid node named `node` produces `"node": "<IP>"`. This code then looks up that IP as a Kubernetes node and periodically deletes the valid gateway policy. Please distinguish policy kinds/ownership with a dedicated metadata key (or avoid treating gateway-priority rows as node-owned without one), and cover a centralized ECMP node named `node`.

zhangzujian (changes requested):
I found two correctness issues in the periodic GC path: a time-of-check/time-of-use race can delete a replacement live node policy, and an external-ID key collision can cause a valid centralized ECMP policy to be collected. Details are inline.

pkg/controller/gc.go
chomatdam: Correct. Fixed in 5c2aa8508. 

UUID alone is not enough: if the new node gets the same join IP, AddLogicalRouterPolicy keeps the row and only rewrites external_ids. DeleteLogicalRouterPolicyIfUnchanged deletes by UUID with an ovsdb wait on external_ids in the same transaction. If the row changed or is gone, nothing is deleted.

Test_DeleteLogicalRouterPolicyIfUnchanged relabels the row between list and delete. It fails with a plain UUID delete.

pkg/controller/gc.go
chomatdam: Correct. Fixed in 5c2aa8508. Startup gcNode on master has the same bug.

At the gateway priority, node now counts as an owner only if the row's subnet is distributed. No new key, so no migration. I can add one if you prefer.

Tests: TestGcNodeRouterPoliciesKeepsCentralizedECMPPolicy (gateway node named node), TestGcNodeRouterPoliciesDeletesStaleDistributedGatewayPolicy.

chomatdam:
marking as draft this PR until the root cause is fixed instead of adding an extra gc sweep

chomatdam:
this now addresses the root cause:
- node-policy deletion works even when the IP finalizer has already released the address
- the finalizer releases the address using the correct IPAM node key
- ownership checks protect policies taken over by replacement nodes, and takeover recreates a policy if concurrent deletion removed it
- initial PR with periodic GC has been removed

and new regression tests exercise the production deletion race and the ownership handoff races

pkg/controller/node.go
```diff
@@ -1129,6 +1129,20 @@ func (c *Controller) deletePolicyRouteForNode(nodeName, portName string) error {
 			return err
 		}
 	}
+	// IPAM may no longer hold the node address, so also delete by node name. Each listed row is deleted
```
zhangzujian: **P1:** The legacy nexthop deletion above still runs before this owner-guarded cleanup. If `addresses` was snapshotted while non-empty, then the IP handler releases the address and a replacement node reuses it and creates or relabels the priority-30000 policy, the old cleanup can delete the replacement live policy by nexthop before this guarded path runs. Please remove or narrow the legacy nexthop deletion path, or make it owner/UUID-checked in the same transaction. The current test only covers the order where IPAM is already empty.

pkg/controller/node.go
```diff
@@ -1129,6 +1129,20 @@ func (c *Controller) deletePolicyRouteForNode(nodeName, portName string) error {
...
+	// only if it still belongs to this node: a node reusing the address may take the row over meanwhile.
+	externalIDs := map[string]string{"vendor": util.CniTypeName, "node": nodeName}
+	policies, err := c.OVNNbClient.ListLogicalRouterPolicies(c.config.ClusterRouter, util.NodeRouterPolicyPriority, externalIDs, false)
+	if err != nil {
+		klog.Errorf("failed to list logical router policies of node %s on %s: %v", nodeName, c.config.ClusterRouter, err)
+		return err
+	}
+	for _, policy := range policies {
+		if _, err = c.OVNNbClient.DeleteLogicalRouterPolicyIfUnchanged(c.config.ClusterRouter, policy); err != nil {
```
zhangzujian, line 1131: **P2 (CODE_STYLE):** `deletePolicyRouteForNode` now spans lines 1114-1217 (104 lines), exceeding the repository's 100-line function limit in `CODE_STYLE.md`. Please extract the owner-policy cleanup or centralized-gateway cleanup into a helper.

zhangzujian (changes requested):
## Re-review

The previous periodic-GC TOCTOU finding is resolved because that periodic cleanup path is no longer present in this PR. The centralized ECMP external-ID key collision remains an existing master issue explicitly marked out of scope by the PR description.

I found one remaining correctness issue and one standards issue below.

Local validation passed for the three new controller regression tests, the OVS client suite, and `git diff --check`.

pkg/controller/node.go
chomatdam: Correct. Fixed in 88986b869.

I removed the nexthop deletion instead of guarding it. Every priority 30000 node policy is created with the `vendor` and `node` external IDs (the node `ipX.dst` policy and the local DNS ones), and the startup migration relabels older rows, so the owner guarded cleanup already covers the same rows. Node policy deletion no longer reads IPAM at all, so there is no snapshot left to go stale. `DeleteLogicalRouterPolicyByNexthop` had no other caller and is gone.

`TestDeletePolicyRouteForNodeKeepsPolicyOfNodeReusingJoinAddress` covers the order you describe: the deleted node's address is still in IPAM while another node already has a policy with that join IP as nexthop. It fails on 85a87b2e6 (the other node's policy is deleted) and passes now.

pkg/controller/node.go
chomatdam: Fixed in 88986b869. The centralized gateway cleanup is now `deletePolicyRouteForCentralizedSubnetOnNode`, with no behaviour change. `deletePolicyRouteForNode` is 50 lines and the new function 42.

zhangzujian (changes requested):
## Re-review

I found one remaining correctness issue and one lower-priority maintainability issue on head fb296d46d375d52d9a7a60f39dc071f6d68dd065.

### P1: startup GC can still delete a live replacement policy

controller.go:1268-1271 starts the workers before running the startup GC. In gc.go:219-227, gcNode snapshots the node names once; it then lists router policies at gc.go:246-256 and deletes an apparently orphaned row at gc.go:266 with DeleteLogicalRouterPolicy(priority, match), without a UUID/owner guard.

If a replacement node takes over the same policy after the policy list, or a new node creates its policy after the node-name snapshot, the current live policy can still be classified as orphaned and deleted. This can blackhole traffic for the replacement node. The path is pre-existing on master, but this PR explicitly relies on startup GC to remove historical orphans and requires takeover policies to survive.

Please make startup GC use the same guarded deletion (DeleteLogicalRouterPolicyIfUnchanged) and re-check ownership, or explicitly document this as an out-of-scope follow-up rather than relying on the current unsafe deletion.

### P2: duplicated policy creation branch

pkg/ovs/ovn-nb-logical_router_policy.go:58-63 and :81-88 duplicate the create/log/error-wrapping sequence. A small helper would keep the normal and race-recovery paths consistent.

The previous node deletion nexthop race and the function-length issue from the prior review are fixed. Local controller and OVS tests, gofmt, and git diff --check passed.

chomatdam:
Both addressed in fe8297a.

- P1: `gcNode` now re-reads the owning node from the lister per policy and deletes the listed row with `DeleteLogicalRouterPolicyIfUnchanged`. `TestGcNodeKeepsPolicyOfNodeCreatedDuringGc` adds a node while the policy list is returned and fails on the previous revision, which tried to delete that node's policy.
- P2: both create paths in `AddLogicalRouterPolicy` go through `createLogicalRouterPolicy`.

PR description updated to cover the guarded startup gc.

zhangzujian (changes requested):
## Re-review

The latest commit `fe8297a7f` correctly fixes the previous startup policy-GC race: `gcNode` re-reads the owner for each policy and uses `DeleteLogicalRouterPolicyIfUnchanged`. The duplicated policy-creation path is also fixed. I still found the following correctness gaps:

### P1: startup GC can still delete a newly created node's resources

`pkg/controller/gc.go:219-243` snapshots the Node names, then lists IP objects. If a Node is created after the Node snapshot and its `node-<name>` IP appears before the IP list, the stale-node branch can still call `deleteNode`. That can remove the new node's logical switch port, chassis, policies, and IP. Re-check the Node identity (including UID where applicable) before deleting, or add an equivalent guard around this cleanup path. Please add a regression test for this Node/IP snapshot window.

### P1: local-DNS policy cleanup is still unguarded

`pkg/controller/node.go:1335-1356` lists local-DNS policies and then deletes each row unconditionally with `DeleteLogicalRouterPolicyByUUID`. A replacement Node can take over the listed row between those operations, allowing deletion of the replacement owner's policy. This path should use the same unchanged-row/owner guard as the priority-30000 cleanup.

### P1: Node owner checks ignore UID

`pkg/controller/gc.go:358-381` treats a Node owner reference as valid when a Node with the same name exists, without comparing the owner reference UID. A same-name replacement can therefore keep an old IP object alive indefinitely and prevent address reclamation. Validate the owner UID before treating the IP as live.

### P2: missing takeover-window coverage

`TestGcNodeKeepsPolicyOfNodeCreatedDuringGc` covers a Node arriving before the policy list, but not a replacement takeover between policy list and delete. Add a controller-level test for that window so the startup-GC guarantee is exercised end to end.

Local validation on this head passed: `go test ./pkg/controller -count=1`, `go test ./pkg/ovs -run ^TestOvnClientTestSuite -count=1`, `gofmt`, and `git diff --check`.

chomatdam:
Fixed in 77104c6: the IP snapshot window, and the stale-node cleanup now runs under the node key lock so a node joining during gc keeps what its add handler creates. Local-DNS policies are deleted under the unchanged-row guard, with a direct takeover test. Tests cover the IP window, the handler interleaving, and the takeover between policy list and delete.

On the owner UID: ordinary node IPs return from the name-only branch at gc.go:325 and never reach the owner-reference code you cited. While checking this I found that `createOrUpdateIPCR` did not persist an owner-UID-only change, so a same-name replacement kept the old UID on the reused IP object. That is fixed here with a regression test. I have not added a UID-mismatch delete to gcIP: deleting the IP object releases its join address in IPAM, and a replacement node may already hold that address, which is the race this PR removes. Safe reuse of a node IP across a same-name replacement needs coordinated adoption and cleanup rather than a GC check, and I would treat that as a separate change.

PR description updated accordingly.

chomatdam:
One more fix from our own review of 77104c6, pushed as 63e3db4.

The policy loop in `gcNode` still ran outside the node key lock. A same-name replacement adopts its policy row unchanged, same UUID and same external IDs, so `DeleteLogicalRouterPolicyIfUnchanged` passed its guard and gc could delete the live node's policy. The owner read and the guarded delete now run under the owner's node key lock, like the stale node cleanup. `TestGcNodeKeepsPolicyAdoptedBySameNameReplacement` reproduces it against the in-memory northbound database: the policy is gone on 77104c6 and survives with the lock.

PR description updated.

zhangzujian (approved):

PR #7490, opened 2026-09-20
fix: hash policy route ECMP on L4 ports

zhangzujian:

## What type of this PR

- Bug fixes

Policy ECMP (`Logical_Router_Policy` with multiple nexthops) currently hashes only on `ip_src`. That pins every connection from a source IP onto one nexthop. The previous `dp_hash` default also splits TIME-WAIT ACK/RST from the same TCP flow, because those packets are built with `ip_send_unicast_reply` and do not inherit `sk_txhash`.

This follows the static-route `selection_fields={tp_src,tp_dst}` model:

- `tcp` / `udp` / `sctp` flows use `select(values=(...); hash_fields="ip_src,ip_dst,ipv6_src,ipv6_dst,ip_proto,<proto>_src,<proto>_dst")`
- other IP traffic hashes on addresses plus `ip_proto`
- matches are disjoint, so the extra flows keep the original policy priority

Verified `git apply --unidiff-zero` of the full OVN patch sequence on `ovn-org/ovn` `origin/branch-26.03`.


zhangzujian:
Review finding fixed in b7603d54d: the non-TCP/UDP/SCTP ECMP flow used `ip`, which matches IPv4 only, so IPv6 ICMP/ESP and other non-L4 traffic skipped policy-route ECMP entirely. The match now uses `(ip || ip6)` while retaining the original policy match. The updated patch was applied successfully to OVN branch-26.03@adfc375955 with `git apply --check --unidiff-zero`; DCO is passing. Build arm64, CodeQL, and trusted x86 E2E are running for the new head.

zhangzujian:
Follow-up fix pushed in `2ff7bc49b`: regenerated the embedded OVN patch with zero-context hunks so the Kube-OVN outer diff passes `git diff --check` without changing the generated northd code.

Validation: applied with `git apply --check --unidiff-zero` and then applied to OVN `branch-26.03` at `2a23559f2a`; resulting `northd.c` diff remains 40 insertions / 6 deletions. The previous trusted x86 failures were infrastructure/test-environment failures (Kind/kubeadm/container startup and OVN cache inconsistencies), not patch-application or image-build failures.

zhangzujian:
CI job 106196321585 exposed a real patching bug: the zero-context OVN hunk matched shifted line numbers inside `bfd_table_sync` and `build_bfd_map`, causing undeclared `actions`, `lflows`, `od`, and `rule` build errors.

Commit `e971a7f4e` restores unique `build_ecmp_routing_policy_flows` context while retaining the IPv6 non-L4 fallback (`(ip || ip6)`). I validated the embedded patch with repository `git diff --check`, then ran `git apply --check` and applied it against OVN `branch-26.03` at `2a23559f2a`. The resulting source diff is exactly 40 insertions/6 deletions in `build_ecmp_routing_policy_flows`; no changes land in the BFD functions. The branch was rebased onto the latest `origin/pr-7490-head` before pushing.

zhangzujian:
Trusted x86 run `35558058335` has now passed the relevant build gates for commit `e971a7f4e`: `Build kube-ovn-base`, `Build kube-ovn-dpdk-base`, and `Build kube-ovn` all succeeded. The previous `northd.c` undeclared-variable failure is not reproduced. The trusted E2E matrix has started and remains in progress.

zhangzujian:
The remaining trusted x86 failures are unrelated to the policy ECMP patch: several jobs fail while creating a Kind or Hosted OVN Central environment, the IPv6 NetworkPolicy job reports a known hairpin self-connection case (`Expected blocked pod connection` while the reachability matrix ignores hairpin traffic), and the dual-stack VPC Egress Gateway job times out waiting for an EVPN gateway to become Ready with EvpnConf/init failures. The relevant build gates remain green, and no policy-route/ECMP test has failed.

PR #7491, opened 2026-09-20
fix: reconcile distributed-subnet AAP VIP ports

### LINKED ISSUE #7489: [BUG] VIP (port_vips / AAP) cannot egress the cluster on a subnet with `gatewayType: distributed`
### Kube-OVN Version

v1.16.2

### Kubernetes Version

v1.37.0

Talos (v1.14.1)
6.18.51-talos

### Description

A VIP allowed on a pod port — `port_vips`, `subnet.spec.vips`, or a `Vip` CR
with `namespace`/`selector` (AAP) — works from inside the cluster but not from
outside it, on an overlay subnet in the default VPC with
`gatewayType: distributed`. Ingress reaches the pod and the pod replies; the
reply is dropped inside OVN.

**Cause.** Distributed-gateway egress is a policy that matches the port group's
implicit address set, and OVN builds that set from each member port's
`addresses`. Kube-OVN writes the VIP to `port_security` only, so the VIP never
matches and falls through to the router's default route, which points at the
router's own join address:

```
# ovn-nbctl lr-policy-list ovn-cluster
  29000   ip4.src == $public.subnet.01.talos.node.01_ip4   reroute   100.64.0.2
# ovn-nbctl lr-route-list ovn-cluster
  0.0.0.0/0    100.64.0.1   dst-ip
# ovn-nbctl list logical_switch_port keepalived-0.keepalived
addresses     : ["56:ef:1a:ec:09:13 192.168.255.10 2001:db8:1203:9::10"]
port_security : ["56:ef:1a:ec:09:13 192.168.255.10 2001:db8:1203:9::10 192.168.255.100"]
```

This makes keepalived/VRRP VIPs unusable for external clients, which is what
they are for. VMs (KubeVirt) are hit hardest: a VM knows only the address it
was given.

Not a duplicate of the two issues with the same symptom: #3956 was fixed by
removing `vrrp_strict` (not set here), #5785 was Cilium dropping VRRP (no
Cilium here).

[20-lsp.txt](https://github.com/user-attachments/files/32433455/20-lsp.txt)
[30-lr-policy-list.txt](https://github.com/user-attachments/files/32433453/30-lr-policy-list.txt)
[31-lr-route-list.txt](https://github.com/user-attachments/files/32433452/31-lr-route-list.txt)
[32-port-group.txt](https://github.com/user-attachments/files/32433454/32-port-group.txt)

### Steps To Reproduce

Full manifest attached as `repro.yaml`. Single interface, no Multus, no NAD.

1. Overlay subnet in the default VPC, `gatewayType: distributed`,
   `natOutgoing: false`, routed to the outside (here via `kube-ovn-speaker`).
2. A `Vip` reserving `192.168.255.100` from that subnet.
3. Two pods on the subnet, annotated with `logical_switch`,
   `port_security: "true"` and `port_vips: "192.168.255.100"`.
4. keepalived in both, VRRP on `eth0`, no `use_vmac`, no `vrrp_strict`.
5. From outside the cluster, ping the VIP and a pod's own address.

[repro.txt](https://github.com/user-attachments/files/32433555/repro.txt)

### Current Behavior

| Client | pod address `.10` | VIP `.100` |
|---|---|---|
| Pod in another subnet | replies | replies |
| The node (hostNetwork) | replies | replies |
| Host outside the cluster | replies | **100% loss** |

Inside the pod, every request arrives and every reply leaves on `eth0`. The
reply enters OVS from the pod's veth and never reaches `ovn0`. `ovn-trace`
shows where it dies: sourced from the VIP, the router ARPs for its own join
address; sourced from the pod address, it is forwarded to the node.

<details><summary>tcpdump inside the pod, ovn-trace VIP vs pod address</summary>

```
# tcpdump -ni any icmp   (in keepalived-0, while an outside host pings the VIP)
eth0  In  IP 10.255.255.2 > 192.168.255.100: ICMP echo request, id 112, seq 1, length 64
eth0  Out IP 192.168.255.100 > 10.255.255.2: ICMP echo reply, id 112, seq 1, length 64
# outside host: 3 packets transmitted, 0 received, 100% packet loss

# ovn-trace, src = 192.168.255.100  (VIP)
    eth.dst = 00:00:00:00:00:00;
    arp { arp.spa = 0x64400001; arp.tpa = 0x64400001; arp.op = 1; ... };   # ARP for 100.64.0.1, dropped

# ovn-trace, src = 192.168.255.10   (pod address)
    eth.dst = 3a:c4:61:41:f5:c6;
    output("node-talos-node-01");                                          # forwarded
```

</details>

#### Proof of the cause

Adding the VIP to the port's `addresses` by hand, nothing else changed, fixes
it; removing it breaks it again:

```
addresses : ["56:ef:.. 192.168.255.10 2001:db8:1203:9::10"]                   -> 100% loss
addresses : ["56:ef:.. 192.168.255.10 2001:db8:1203:9::10 192.168.255.100"]   ->   0% loss
addresses : ["56:ef:.. 192.168.255.10 2001:db8:1203:9::10"]                   -> 100% loss
```

But that is **not** the fix. Two pods share a VIP, and OVN refuses the same IP
on two ports of one switch:

```
ovn-nbctl: Error on switch public-subnet-01: duplicate IPv4 address '192.168.255.100' found on logical switch port 'keepalived-0.keepalived'
```

Worse, every IP in a normal port's `addresses` gets an ARP responder
(`build_lswitch_arp_nd_responder_known_ips`, `northd/northd.c`). OVN then
answers ARP for the VIP with one pod's MAC and swallows the other pod's
gratuitous ARP. On failover to the second pod the VIP is lost from outside
**and inside** the cluster; reverting `addresses` restores it at once.

<details><summary>ARP responder flow, traces and failover result after adding the VIP to addresses</summary>

```
# ovn-sbctl lflow-list public-subnet-01 | grep 'arp.tpa == 192.168.255.100'
table=22(ls_in_arp_rsp), priority=50,
  match=(arp.tpa == 192.168.255.100 && arp.op == 1 && eth.dst == ff:ff:ff:ff:ff:ff),
  action=(eth.dst = eth.src; eth.src = 56:ef:1a:ec:09:13; arp.op = 2; ...)

# gratuitous ARP from keepalived-1 (backup): answered back to it, never flooded
# ovn-trace public-subnet-01 'inport=="keepalived-1.keepalived" && arp.op==1 && arp.spa==192.168.255.100 && arp.tpa==192.168.255.100 ...'
eth.src = 56:ef:1a:ec:09:13; arp.op = 2; arp.sha = 56:ef:1a:ec:09:13;
output("keepalived-1.keepalived");

# router's ARP for the VIP: answered with keepalived-0's MAC
# ovn-trace public-subnet-01 'inport=="public-subnet-01-ovn-cluster" && arp.op==1 && arp.tpa==192.168.255.100 ...'
eth.src = 56:ef:1a:ec:09:13; arp.op = 2; arp.sha = 56:ef:1a:ec:09:13;

# failover: keepalived-0 frozen, VIP moved to keepalived-1
ping VIP from outside : 3 packets transmitted, 0 received, 100% packet loss
ping VIP from a pod   : 3 packets transmitted, 0 received, 100% packet loss
# revert addresses, keepalived-1 still master
ping VIP from a pod   : 3 packets transmitted, 3 received, 0% packet loss
MAC_Binding 192.168.255.100 -> e2:a4:38:dd:90:ac  (keepalived-1)
```

</details>

#### What does not work

Same cluster, no manual OVN edits:

| Attempt | Result |
|---|---|
| `port_vips` + `port_security: "true"` | VIP in `port_security` only; unreachable |
| `subnet.spec.vips` | same code path; unreachable |
| `Vip` with `namespace` + `selector`, pod annotated `aaps` | virtual port created with both parents; unreachable |
| the above, virtual port added to the port group by hand | still unreachable: its `addresses` is `[]` |
| `<provider>.kubernetes.io/layer2_forward: "true"` | adds `unknown`, which is not an IP; unreachable |

#### Workarounds

- `gatewayType: centralized`: the policy becomes a CIDR match and covers the
  VIP. Not acceptable where distributed egress is required.
- A `policySrc` static route on the VPC per VIP (`cidr: <vip>/32`,
  `nextHopIP: <node join address>`). Does not scale, mutates the default VPC
  per VIP, pins the VIP's egress to one node.

#### Related: dual-stack `virtual-ip` never binds

On a dual-stack subnet Kube-OVN writes
`virtual-ip="192.168.255.100,2001:db8:1203:9::f"`. northd parses that option
with `ip46_parse`, which fails on the pair, so it installs zero `bind_vport`
flows and the `Port_Binding` never gets a chassis. OVN's `virtual-ip` is a
single address.

[40-ovn-trace.txt](https://github.com/user-attachments/files/32433572/40-ovn-trace.txt)
[41-pod-tcpdump.txt](https://github.com/user-attachments/files/32433573/41-pod-tcpdump.txt)
[50-proof-addresses-field.txt](https://github.com/user-attachments/files/32433574/50-proof-addresses-field.txt)
[51-addresses-on-pod-port-breaks-failover.txt](https://github.com/user-attachments/files/32433576/51-addresses-on-pod-port-breaks-failover.txt)
[60-reachability-matrix.txt](https://github.com/user-attachments/files/32433575/60-reachability-matrix.txt)

### Expected Behavior

A VIP allowed on a port egresses the same way the port's own address does: it is
covered by the distributed-gateway reroute policy.

#### Suggested fix

Use the AAP virtual port Kube-OVN already creates for a `Vip` with
`namespace` + `selector`, the way neutron's ML2/OVN driver handles an
allowed-address-pair: give it `addresses` and put it in the port group. For a
`type: virtual` port northd installs only `bind_vport` flows and **no** ARP
responder, so the VIP can sit in `addresses` safely, and the port group's
address set then contains it.

Verified by hand with two NB edits on the unmodified AAP setup.
`2e:a5:9b:20:42:d2` is the MAC Kube-OVN already allocated to the `Vip`:

```
ovn-nbctl lsp-set-addresses keepalived-vip "2e:a5:9b:20:42:d2 192.168.255.100 2001:db8:1203:9::f"
ovn-nbctl add port_group public.subnet.01.talos.node.01 ports <uuid of keepalived-vip>
```

| Check | Result |
|---|---|
| port group address set | contains `192.168.255.100` |
| `ls_in_arp_rsp` flows for the VIP | none; router's ARP is flooded, VRRP master answers |
| ping from outside, keepalived-0 master | 0% loss |
| failover to keepalived-1, then failback | 0% loss from outside and inside; router `MAC_Binding` follows the master |

Controller change, in `handleUpdateVirtualParents` (`pkg/controller/vip.go`):

1. set the virtual port's `addresses` to `<vip mac> <v4ip> <v6ip>`;
2. add the virtual port to the subnet's per-node port group — following the
   southbound `Port_Binding.chassis`, or simply every node that hosts a
   parent (overlapping `29000` policies still egress, just not locally).

This covers only the `Vip` + `aaps` path. `port_vips` and `subnet.spec.vips`
have no virtual port and no MAC to give OVN; they would need a separate
per-node address set, or docs steering users to the `Vip` CR. The dual-stack
`virtual-ip` binding above does not block this fix, but a complete AAP
implementation needs it too.

[52-virtual-port-fix-with-failover.txt](https://github.com/user-attachments/files/32433590/52-virtual-port-fix-with-failover.txt)

zhangzujian:
## What does this PR do?

Fixes the allowed-address-pair (`Vip` with namespace/selector) path for VIPs on distributed gateway subnets.

The controller now:

- creates one virtual logical switch port per VIP address family, using `vip:<VIP-NAME>:ipv4` and `vip:<VIP-NAME>:ipv6`;
- records each VIP MAC and IP on the corresponding virtual logical switch port;
- splits dual-stack AAP VIPs into two virtual ports because OVN `virtual-ip` accepts one address;
- atomically migrates an existing pre-dual-stack single-stack `<VIP-NAME>` virtual port to the new address-family name, preserving its UUID, port-group membership, and virtual parents;
- adds virtual ports to current per-node distributed-subnet port groups before removing stale memberships, avoiding transient port-group blackholes;
- removes stale port-group memberships when AAP parents move or disappear;
- removes generated virtual ports during VIP cleanup, including ports from the earlier development naming scheme.

Virtual ports do not install normal logical-switch ARP responder flows, so this avoids adding a shared VIP to the parent ports' `addresses` field.

Fixes #7489 (AAP path).

`port_vips` and `subnet.spec.vips` use separate code paths and are not changed by this PR.

## How Has This Been Tested?

- `go vet ./pkg/controller ./pkg/ovs`
- `golangci-lint run ./pkg/controller/... ./pkg/ovs/...`
- `make lint`
- `git diff --check`

`go test ./...` still reaches the pre-existing `cmd/controller/TestCommandHelpDescriptions` assertions for `kube-ovn-daemon --help`, which are unrelated to this diff.


nfarhadian:
Thanks for this PR, I'm glad team maintaining Kubeovn responds to issues in less than a day!

Here are 2 points Claude added about this:

**1. `syncVirtualVipPortGroups` blackholes the VIP on every reconcile.**
The remove-from-all-groups and add-back are two transactions with no diff against
current state, so an unchanged reconcile still drops the VIP from the address set.

Measured on v1.16.2, replaying that exact remove -> add sequence while pinging
the VIP from outside: 0% loss at steady state, 5% loss across 5 reconciles (~100ms each).

It fires often. `enqueueUpdatePod` enqueues on any pod update carrying the
`aaps` annotation, not on the annotation changing. ordinary status updates trigger it.

**2. `-ipv6` suffix can collide.** A Vip literally named `foo-ipv6` clashes with the
v6 port of a dual-stack Vip named `foo`. Also the delete fallback orphans the v6 port
when status and spec IPs are both empty.

zbb88888 (approved):

zbb88888:
@zhangzujian hi，PR 看起来没问题，但是太多 E2E 失败了，后续应该需要修复下。

zhangzujian:
Validated both points from comment 5758592329 and pushed d5e0d6c6b.

1. Confirmed: syncVirtualVipPortGroups previously removed every VIP membership in one transaction and added desired memberships in later transactions, so every reconcile could briefly blackhole the VIP. The fix adds desired node port-group memberships first and removes only stale node groups. Reconciles with unchanged ownership no longer remove the VIP from its current group.

2. Confirmed: the -ipv6 secondary-port name could collide with a valid VIP name, and deletion with missing IP status/spec could only recover the primary port. The fix uses a stable IPv6-derived hash suffix, and deletion discovers matching virtual LSPs from the subnet (including the secondary port when IP state is absent). Known-IP discovery also checks the virtual-ip option before deleting a hashed secondary.
 
Validation: go test ./pkg/controller ./pkg/ovs -count=1; go vet ./pkg/controller ./pkg/ovs; make lint; git diff --check.

zhangzujian:
Implemented the requested VIP LSP naming and upgrade migration.

- New names are `vip:<VIP-NAME>:ipv4` and `vip:<VIP-NAME>:ipv6`, matching the master LB naming style.
- The legacy pre-dual-stack `<VIP-NAME>` LSP is migrated only when it is a virtual port on the same subnet with the requested `virtual-ip`.
- Migration updates the existing LSP `name` in a single OVN transaction, preserving its UUID, port-group membership, and virtual-parents.
- IPv4-only and IPv6-only migration tests verify UUID preservation and removal of the legacy name.
- Local validation passed: `go test ./pkg/controller ./pkg/ovs -count=1`, `go vet ./pkg/controller ./pkg/ovs`, targeted `golangci-lint`, and `git diff --check`.

HEAD: `b8b9e52d9c77106720c8438d244582757524487d`

zhangzujian:
@nfarhadian Could you please review again?

nfarhadian:
I checked it, it works fine, thanks

Only thing I want to mention is upgrading a cluster that already has a dual-stack VIP leaves a dead port behind
After upgrading on my cluster, one VIP had three ports:

keepalived-vip            <- the old one, still here
vip:keepalived-vip:ipv4   <- new
vip:keepalived-vip:ipv6   <- new

It's harmless, but it's there
Your two migration tests use an address each so this case is untested rather, than rejected on purpose. To cover it: also accept a comma‑joined virtual‑ip which contains the address or delete the port once both new ports exist.

Also vip_test.go is not working on golang 1.26 , backporting to 1.16 is not possible with this test (but code itself is working)

PR #7505, opened 2026-09-22
fix: use valid protocol matches for policy ECMP

zhangzujian:
## Summary

- Replace invalid `!tcp`, `!udp`, and `!sctp` nominal-predicate negations in the generic policy-ECMP flow.
- Use explicit `ip.proto !=` comparisons so OVN accepts and installs the non-L4 flow.
- Fix the malformed unified-diff hunk metadata after splitting the generated C string.

## Root cause

OVN rejected the generated match with `Nominal predicate tcp may only be tested positively`. The dropped flow caused ICMP traffic in the IPv4 EVPN VPC Egress Gateway E2E to time out.

## Validation

- `git apply --check --unidiff-zero dist/images/patches/scope-ecmp-source-hashing-to-policy-routes.patch` against OVN `branch-26.03`
- `git diff --check`
- Original failure: [run 35708154165, job 106688499120](https://github.com/kubeovn/kube-ovn/actions/runs/35708154165/job/106688499120)


zhangzujian:
The follow-up x86 E2E run (35719365644) still used the old OVN base image.

Evidence:
- Build kube-ovn-base checked out `65ec88ef544d6f024aa48b7cb77478f4545ed50b` and applied the updated OVN patch.
- Build kube-ovn then resolved `kubeovn/kube-ovn-base:v1.17.0[-debug]` from Docker Hub instead of the freshly loaded artifact.
- The failing OVN logs still contained `!tcp && !udp && !sctp`.

Commit `5b6c10e22` makes the x86 workflow pass `DOCKER_BUILD_ARGS=--pull=false` when rebuilding the base image, so the final image uses the local patched base. A new dispatcher run has been triggered for this commit.

zhangzujian:
The latest executor run `35732226231` still used the old base image: its OVN logs again contain `!tcp && !udp && !sctp`. The trusted executor uses the base workflow definition, so the workflow-level `DOCKER_BUILD_ARGS=--pull=false` change was not available to that run.

Commit `28cfa3a7a` now makes `DOCKER_BUILD_ARGS` default to `--pull=false` in `makefiles/build.mk`; the PR checkout therefore applies the setting even when the executor workflow itself comes from the base ref. The override remains available for callers that need a different pull policy.

zhangzujian:
The previous executor confirmed that `--pull=false` alone was insufficient.

Evidence from run `35744525681`, Build kube-ovn job `106815012315`: the rebuilt base artifact was loaded successfully, but `docker build` still resolved `kubeovn/kube-ovn-base:v1.17.0[-debug]` from Docker Hub through the `docker-container` Buildx builder (old digests `sha256:c7f9a970...` and `sha256:973187d9...`). The resulting EVPN job still logged `!tcp && !udp && !sctp`.

Commit `b11cbec84` sets `BUILDX_BUILDER=default` on the final `Build` and `Build dpdk` steps, so Docker builds use the daemon image store populated by the base artifacts. A new trusted x86 executor is required to verify the patched base is now used.

zhangzujian:
Follow-up: the previous workflow-only builder override would still be skipped by trusted executor runs, which execute the base workflow definition. Commit `7b0f3b61c` moves the override into the PR checkout's `makefiles/build.mk`: all `docker build` targets now default to `BUILDX_BUILDER=default` via `DOCKER_BUILD_ENV`. The base workflow therefore inherits the fix when it runs `make build-kube-ovn`, while regular buildx base-image targets remain unchanged.

zhangzujian:
The latest trusted executor reproduced a different failure than the original parser error.

All OVN nodes logged:
```
error parsing match "(ip4.src == $VEG....) && (ip || ip6) && ip.proto != 6 && ip.proto != 17 && ip.proto != 132": Nominal field ip.proto may only be tested for equality
```

Commit `a31515b12` removes that invalid fallback match. The generic non-L4 flow now keeps the original policy priority, while TCP/UDP/SCTP flows use `priority + 1`, following OVN static-route ECMP. IPv4/dual EVPN ICMP therefore use the valid generic flow; L4 traffic still hashes on protocol-specific ports. New validation is running in trusted executor run `35774717257`.

zhangzujian:
Trusted executor run 35775541331 reached the VPC Egress Gateway E2E jobs on commit ede100558b16debd9bff8fdf49ff26140108ce25.

Results:
- IPv6: passed (job 106925632990)
- IPv4: failed (job 106925633052) at the existing EVPN test's final ping to 10.99.0.1
- Dual: failed (job 106925633092) at the same EVPN ping

The IPv4 log shows BGP sessions established, EVPN routes learned, and VRF routes present before the ping. It contains no Nominal field ip.proto error. The previous trusted run on the parent SHA accee11ff4cfe561ff853044e613f70f69ed240b failed the same IPv4 and dual EVPN ping at the same test location (e2e_evpn_test.go:345). This indicates the remaining EVPN ping failure is baseline/flaky and is not introduced by ede100558.

Relevant links:
- Current IPv4: https://github.com/kubeovn/kube-ovn/actions/runs/35775541331/job/106925633052
- Current dual: https://github.com/kubeovn/kube-ovn/actions/runs/35775541331/job/106925633092
- Previous trusted run: https://github.com/kubeovn/kube-ovn/actions/runs/35762822001

zhangzujian:
Follow-up fix in commit `74d5361ae1ad3a494c1c06ac2ba34184c9facd93` (rebased remote PR HEAD):

- Create the EVPN VXLAN interface with `nolearning` so FRR remains the source of EVPN FDB entries.
- Place the test backend address on `br-vpn` as the EVPN SVI instead of the VRF device.
- Add a shell regression check for the VXLAN initialization mode.

Local validation passed: `bash -n`, the EVPN init shell regression check, targeted Go package compilation, `make lint`, and `git diff --check`. Trusted executor run: https://github.com/kubeovn/kube-ovn/actions/runs/35797872759 (base image build still in progress).

